PR #1, opened 2016-01-28
[DO NOT MERGE] Ori changes 20160127

orischwartz:
refactor to support file encryption, separate rails ActiveJob from application logic


shanear:
Couple notes: 
1) There was a lot of test coverage removed (`download_file_job_spec.rb`, ` spec/services/download_documents_spec.rb `) without any added to replace it.

2) There's a lot going on here. Adding functionality and refactoring at the same time. We're forced to take this all or nothing. I'd like to be able to discuss the refactoring and the encryption apart from each other. It's possible we don't want/need the encryption and want the refactor (and vis-versa).

3) The message for the pull request should reflect exactly what the pull request is trying to do (ie: "Encrypt files when saving them to the filesystem" or "Refactor DownloadDocuments service object into EFolderExpress object")

4) Totally agree with your method name changes in `vbms_service.rb`

5) Your use of dependency injection is awesome. I want to use this pattern more.

6) Not sure about the EFolderExpress object. Seems like a kitchen sink class.

More to come maybe... glad to chat in person about any of this.

orischwartz:
Should have added, DO NOT MERGE, @shanear and I to discuss.

orischwartz:
Closing in favor of smaller changes as discussed

PR #11, opened 2016-02-02
Add Rubocop linter and fix all the issues

shanear:


paultag:
Is there a reason to use double quotes? I remember someone saying something about if you're not doing string interpolation, you should use `'foo'` -- urban myth or a thing?

shanear:
@paultag there's no performance issue with using double quoted strings. Tons of people have tested this (example: https://viget.com/extend/just-use-double-quoted-ruby-strings).

It's easier to just say "Always use double quotes", than "Use single quotes, except for when using string interpolation"

paultag:
@shanear Interesting. Most of these posts are about MRI, and seem a bit old. I just re-tried with jruby, and modified a script to try to throw more cycles to show any differences I can find. Based on https://gist.github.com/bloudermilk/3681645

Running 1,000,000 cycles, I got:

```
single    0.840000   0.090000   0.930000 (  0.443566)
double    0.460000   0.070000   0.530000 (  0.426730)
```

And, with 100,000,000

```
              user     system      total        real
single   15.280000   0.210000  15.490000 ( 14.419640)
double   16.770000   0.180000  16.950000 ( 16.180061)
```

Nearly identical in both cases. Neat.

paultag:
This looks good to me. You added it to travis, which would have been my last nit, so I'll wait until we're green and land it.

shanear:
@paultag nice! glad you tried it out. I didn't think about the diff between MRI & jruby possibly being an issue.

paultag:
:us:

paultag:
I merged master back into this branch, it might break the fixes, but we can't merge without it :)

paultag:
Nevermind, went green!

PR #18, opened 2016-05-17
Specify platform, add execjs env

paultag:
PR #15


orischwartz:
This has run its course ... :cry:

PR #31, opened 2016-05-19
Use an absolute path for downloads_dir

orischwartz:
Worker job will not be started in rails root dir


shanear:
lol on this branch name 💯

Issue #34, opened 2016-05-20
initialize VBMS client the same way as caseflow-certification


orischwartz:
https://github.com/department-of-veterans-affairs/caseflow-certification/blob/master/config/secrets.yml#L37

https://github.com/department-of-veterans-affairs/caseflow-certification/blob/master/app/services/appeal_repository.rb#L121

orischwartz:
fixed in #38

PR #38, opened 2016-05-20
initilize VBMS client for production, add logger

orischwartz:


shanear:
Rubocop... gets me every time.

orischwartz:
Closes #34

PR #59, opened 2016-05-25
Don't hardcode the saml entity id

### LINKED ISSUE #45: Make omniauth entity ID configurable via environment variable, not hardcoded
https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/config/initializers/omniauth.rb#L21

orischwartz:
closes #45


shanear:
😃 🙇 👍 💯

PR #80, opened 2016-06-02
Use PostgreSQL

### LINKED ISSUE #79: Use PostgreSQL instead of SQL lite

shanear:
fixes #79

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -25,6 +25,4 @@ test:
 
 production:
   <<: *default
-  username: <%= ENV["POSTGRES_USERNAME"] %>
-  password: <%= ENV["POSTGRES_PASSWORD"] %>
-  database: caseflow_efolder
+  url: <%= ENV["POSTGRES_URL"] %>
--- a/config/database.yml
+++ b/config/database.yml
@@ -5,21 +5,26 @@
 #   gem 'sqlite3'
 #
 default: &default
-  adapter: sqlite3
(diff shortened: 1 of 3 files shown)


config/database.yml
```diff
@@ -5,21 +5,26 @@
...
-  database: db/test.sqlite3
+  username: <%= `whoami` %>
+  database: caseflow-efolder-test
 
 production:
   <<: *default
-  database: db/production.sqlite3
+  username: <%= ENV["POSTGRES_USERNAME"] %>
+  password: <%= ENV["POSTGRES_PASSWORD"] %>
+  database: caseflow-efolder
```
orischwartz: RDS doesn't allow database names with hyphens. Can we use `efolder_express` (the way our other AWS resources are named)?

config/database.yml
```diff
@@ -5,21 +5,26 @@
...
   <<: *default
-  database: db/test.sqlite3
+  username: <%= `whoami` %>
+  database: caseflow-efolder-test
 
 production:
   <<: *default
-  database: db/production.sqlite3
+  username: <%= ENV["POSTGRES_USERNAME"] %>
+  password: <%= ENV["POSTGRES_PASSWORD"] %>
```
orischwartz: Don't we also need the url here? 

`url: <%= ENV['POSTGRES_URL'] %>`

PR #83, opened 2016-06-03
S3

### LINKED ISSUE #74: Abstract file storage to allow for storage in s3 in stead of local FS
`DownloadDocuments` should be able to write to local FS in `development`or to S3 in `production`. 

> Amazon S3 supports bucket policies that you can use if you require server-side encryption for all objects that are stored in your bucket. For example, the following bucket policy denies upload object (s3:PutObject) permission to everyone if the request does not include the x-amz-server-side-encryption header requesting server-side encryption.

http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingServerSideEncryption.html

S3 also allows for rules that will e.g., delete files older than 3 days to enforce a document cleanup policy.

shanear:
fixes #74


orischwartz:
* S3 will use the document-id as a prefix to save the file whereas on local FS we use the index (relative to downloaded case). Should we be consistent? Or better to leave existing unchanged? 
* `config/secrets.yml` and `config/secrets.yml` appear to have not been changed but are included in the diff.

PR #89, opened 2016-06-08
update status after inserting document records, closes #88

### LINKED ISSUE #88: Race condition: front-end app reads manifest documents before backend has finished inserting all records
This leads to the count being inaccurate with large cases.

orischwartz:


orischwartz:
blocked on #90

PR #90, opened 2016-06-08
update gemfile for nokogiri vulnerability

orischwartz:


Gemfile.lock
```diff
@@ -165,42 +167,53 @@ GEM
...
+    parser (2.3.1.2)
       ast (~> 2.2)
+    pkg-config (1.1.7)
     poltergeist (1.9.0)
       capybara (~> 2.1)
       cliver (~> 0.3.1)
       multi_json (~> 1.0)
       websocket-driver (>= 0.2.0)
     powerpack (0.1.1)
+    pry (0.10.3)
```
paultag: Seems like we added a pry

I think it's OK

PR #105, opened 2016-06-14
use secure cookies in production, closes #95

### LINKED ISSUE #95: Enable Secure cookies

orischwartz:


orischwartz:
See also #69 -- we disable the global `force_ssl` configuration option for the ELB health check.

config/initializers/session_store.rb
```diff
@@ -1,3 +1,3 @@
 # Be sure to restart your server when you modify this file.
 
-Rails.application.config.session_store :cookie_store, key: '_efolder-express_session', expire_after: 2.weeks
+Rails.application.config.session_store :cookie_store, key: '_efolder-express_session', expire_after: 2.weeks, secure: Rails.env.production?
```
paultag: Will this break the healthcheck endpoint?

orischwartz: No. This only affects the ability to set a cookie in the response, unlike the `force_ssl` config option which enforces both secure cookies and forces an http->https redirect on all requests.

orischwartz:
I'm going to merge this myself -- per @gmorgana-va we need this today.

Issue #107, opened 2016-06-22
Make database pool size configurable


orischwartz:
Currently set to 25 for all environments.

This can be a problem when deploying to our staging environment which uses a `db.t2.micro` with `1GB` of memory. The default postgres RDS config for `max_connections` is 

```
memory/31457280 = 1073741824/31457280 = 34.13
```

This can be exceeded with just two instances running with full connections pools.

PR #113, opened 2016-06-29
Smaller default pool size for non-prod envs

orischwartz:


orischwartz:
Alleviates contention for limited db connections until #107 is properly addressed.

PR #114, opened 2016-06-30
Use C-Ruby and the ruby version of Connect VBMS

shanear:


orischwartz:
#### This change requires the following runtime modifications:

* `java-1.8.0-openjdk-devel` no longer required (for jRuby)
* `postgresql-devel` required (previously JDBC)
* Must now specify the following connect VBMS environment variables:
    - `CONNECT_VBMS_SERVER_CERT`
    - `CONNECT_VBMS_CLIENT_KEYFILE`

orischwartz:
Looks like this will also need to update travis config:

https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/.travis.yml#L7

PR #118, opened 2016-07-05
Add support for restarting stalled jobs

### LINKED ISSUE #115: Create "janitor" process for stopped jobs
- Store download times
- Make ruby jobs self destroying on duplicate
- add #stalled? method to download
- add restart hook to GET /download 
- add restart hook to cron job (possibly with sidekiq)

shanear:
fixes #115


orischwartz:
Tests failing on 

```
 1) DownloadDocuments#download_contents when one file errors saves download state for each document
     Failure/Error: expect(successful_document.filepath).to eq((Rails.root + "tmp/files/#{download.id}/0-filename.pdf").to_s)
     
       expected: "/home/travis/build/department-of-veterans-affairs/caseflow-efolder/tmp/files/27/0-filename.pdf"
            got: "/home/travis/build/department-of-veterans-affairs/caseflow-efolder/tmp/files/27/1-filename.pdf"
     
       (compared using ==)
```

app/controllers/downloads_controller.rb
```diff
@@ -71,7 +82,7 @@ def downloads
   end
 
   def recent_downloads
-    @recent_downloads ||= downloads.where(status: [3, 4])
+    @recent_downloads ||= downloads.where(status: [3, 4, 5])
```
orischwartz: Can we use the enum names? e.g., 

```ruby
downloads.where(status: [:pending_documents, :packaging_contents, :complete])
```

spec/services/download_documents_spec.rb
```diff
@@ -90,9 +93,12 @@
         successful_document = Document.first
         expect(successful_document).to be_success
         expect(successful_document.filepath).to eq((Rails.root + "tmp/files/#{download.id}/0-filename.pdf").to_s)
```
orischwartz: I think this test was always wrong -- if the only file to succeed is the one with id `1` then the first successful document should have ID 1. Unclear why this test didn't fail before this change.

app/controllers/downloads_controller.rb
shanear: I wanted to do that, but it wasn't working. Not sure why, and I decided to just go with this.

I agree though, it's not ideal.

spec/services/download_documents_spec.rb
shanear: it's working locally for me 😉

the number has to do with its index in the list, not the id, so i'm actually not sure why this is failing.

shanear: I think it has to do with the default ordering of records being random on some postgres versions, and not random on my local version?

anyways, i added a specified ordering. That should fix it.

orischwartz: Weird, works locally for me too...

PR #122, opened 2016-07-07
Add time estimates to download progress

### LINKED ISSUE #53: Add rough time estimates to eFolder Express
When we conducted usability testing, many didn't not have a clear estimate for how long the download would take, or that they could close the fetching page while the eFolder was downloading. Even if we don't have an accurate time estimate, we should set user's expectations about how much time they'll have to wait. 

- Option 1: Give an actual time estimate, if possible 
- Option 2: Tell the user how many files are left (e.g. `Fetching 3 of 240 files`)

shanear:
fixes #53

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -12,6 +12,8 @@ class Download < ActiveRecord::Base
 
   TIMEOUT = 10.minutes
 
+  TIMEOUT = 10.minutes
+
   def demo?
     file_number =~ /DEMO/
   end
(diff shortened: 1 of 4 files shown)


shanear:
@orischwartz this one got kind of messed up from weird rebasing

app/models/download.rb
```diff
@@ -12,6 +12,8 @@ class Download < ActiveRecord::Base
 
   TIMEOUT = 10.minutes
 
+  TIMEOUT = 10.minutes
```
orischwartz: Is this a merge conflict?

PR #123, opened 2016-07-07
Change numbering scheme on filename downloads

### LINKED ISSUE #48: Rename/Download Files in Upload Date Order
During our usability test we observed most (all?) users requested that the downloaded files be ordered by the date they were uploaded. An important consideration is what we do when we have to skip a file because it can't be downloaded. We could do this in a couple ways...

- Name files to 01_FILENAME.pdf, 02_FILENAME.pdf, etc and skip over numbers when a file is missing. For example if File 03 could not be downloaded we'd go from 02_FILENAME.pdf to 04_FILENAME.pdf
- Name files to UPLOAD-DATE-TIMESTAMP_FILENAME.PDF. The upside is we wouldn't have to do anything tricky with skipping file name numbers but could be more confusing for the user.

We should also add copy indicating how the files will be named when downloaded to the download page.

Me and @shanear will discuss.

shanear:
fixes #48

### DIFF
--- a/app/services/download_documents.rb
+++ b/app/services/download_documents.rb
@@ -71,7 +71,7 @@ class DownloadDocuments
   end
 
   def unique_filename(document, index)
-    "#{index}-#{document.filename}"
+    "#{format('%04d', index)}0-#{document.filename}"
   end
 
   def fetch_from_s3(document)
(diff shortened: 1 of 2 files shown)


app/services/download_documents.rb
```diff
@@ -71,7 +71,7 @@ def save_document_file(document, content, index)
   end
 
   def unique_filename(document, index)
-    "#{index}-#{document.filename}"
+    "#{format('%04d', index)}0-#{document.filename}"
```
orischwartz: Why the trailing 0 after the index? (i.e., why `00040-xx.txt` vs `0004-xx.txt`?)

shanear: so users can interject VirtualVA files in-between the eFolder files.

PR #129, opened 2016-07-08
log user's id and email on each line (not entire user hash)

### LINKED ISSUE #128: Less verbose user logging in tags
Instead of [dumping the user hash](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/config/initializers/logger.rb#L9) in a logger tag, just log the user's email/id e.g.,

```ruby
user = session["user"]
["id", "email"].map { | attr | user[attr] }.join(" ") if user
```

orischwartz:
Closes #128

### DIFF
--- a/config/initializers/logger.rb
+++ b/config/initializers/logger.rb
@@ -6,7 +6,8 @@ end
 log_tags = []
 log_tags << lambda { |req|
   session = get_session(req)
-  session["user"]
+  user = session["user"]
+  ["id", "email"].map { | attr | user[attr] }.join(" ") if user
 }
 
 # :nocov:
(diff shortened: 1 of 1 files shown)


config/initializers/logger.rb
```diff
@@ -6,7 +6,8 @@ def get_session(req)
 log_tags = []
 log_tags << lambda { |req|
   session = get_session(req)
-  session["user"]
+  user = session["user"]
+  ["id", "email"].map { | attr | user[attr] }.join(" ") if user
```
shanear: might be helpful to log station id as well.

orischwartz:
Merging per Shane's comment and review

PR #153, opened 2016-07-13
Modify SSO logic to reroute auth request through SSO domain

shanear:
requires addition of SSO_HOST environment variable to deployment

ie:
```
export SSO_HOST="https://cf.preprod.ds.va.gov"
```


orischwartz:
:+1: @shanear to merge when ready. Also can you please open a corresponding issue in the deployment repo so if you haven't already?

Issue #126, opened 2016-07-07
Add google analytics


orischwartz:
* Include support for events
* Define a funnel so we can track how users are getting from search to listing to zip download

orischwartz:
```
<script>
  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');

  ga('create', 'UA-74789258-2', 'auto');
  ga('send', 'pageview');

</script>
```
See https://github.com/department-of-veterans-affairs/caseflow-certification/pull/229 and https://github.com/department-of-veterans-affairs/caseflow-certification/pull/121 for the implementation details in Certification app.

NickHeiner:
![image](https://user-images.githubusercontent.com/829827/28538346-f9a521ec-707b-11e7-88ee-79813d0fbc95.png)

This definitely worked initially, but has not worked in a long time.

orischwartz:
Looks like I screwed up 47e8423456949bf41218f9b5a62d3321a1ba6565 -- the config attribute is named `analytics_account`, not `google_analytics_account` which is being set in production.rb.

NickHeiner:
Yeah, I was thinking that may be problematic. :smile:

Fixing in #549.

shanear:
hi @orischwartz 👋

orischwartz:
hi everyone! :wave:

PR #174, opened 2016-07-26
Add Google Analytics support

orischwartz:
* Configure account # via `Rails.configuration.analytics_account`
* Emit download events for `start` and `create` download 
* Due to redirects on `download#show`, pass analytics events via `ga_event` querystring parameter.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -158,4 +158,19 @@ module ApplicationHelper
     return "#{format('%.2f', seconds)} <span class=\"ee-stat-unit\">sec</span>".html_safe if seconds < 60
     "#{format('%.2f', seconds / 60)} <span class=\"ee-stat-unit\">min</span>".html_safe
   end
+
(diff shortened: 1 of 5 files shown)


orischwartz:
Closes #126

app/controllers/downloads_controller.rb
```diff
@@ -20,20 +20,24 @@ def create
...
+    redirect_to download_url(@download, ga_event: "Create")
   end
 
   def start
     @download = downloads.find(params[:id])
     @download.update_attributes!(status: :pending_documents)
 
     start_download_files
-    redirect_to download_url(@download)
+    redirect_to download_url(@download, ga_event: "Start")
```
shanear: since we have an acutal database, i'd rather rely on it to deal with back-end events.

We should keep GA strictly for pageviews.

What do you think?

orischwartz:
Removed event-emitting logic, just records pageviews now

PR #175, opened 2016-07-27
Record searches, factor logic into search object

### LINKED ISSUE #148: Log whenever a user sees an error
We should log whenever a user sees an error on the home page or the download page like those below so that we can diagnose issues that users may encounter.

image

image

shanear:
fixes #148


orischwartz:
:+1: just waiting on a fix for https://github.com/department-of-veterans-affairs/appeals-deployment/issues/64 so we can run the migration ahead of deployment

PR #180, opened 2016-08-01
Require authorization to view stats page; add ActiveJob for pre-calculating expensive stats queries and restarting stalled jobs

shanear:


app/controllers/health_checks_controller.rb
```diff
@@ -1,6 +1,5 @@
 class HealthChecksController < ApplicationController
   skip_before_action :authenticate
-  skip_before_action :authorize
```
orischwartz: I think we still need to skip `:authorize` here.

config/sidekiq_cron.yml
```diff
@@ -1,5 +1,11 @@
-schedule_example_job:
+restart_stalled_downloads_job:
   cron: "*/1 * * * *"
   class: "RestartStalledDownloadsJob"
   queue: default
+  active_job: true
+
+calculate_stats_job:
+  cron: "*/2 * * * *"
```
orischwartz: Maybe add comments so readers don't have to parse the cron syntax like "every two minutes"?

Issue #109, opened 2016-06-23
Set Alerts for Suspicious/Unexpected User Behavior


KHarshawat:
Audit monitoring.

KHarshawat:
Here are some ideas for suspicious behavior we should look out for in eFolder Express. In general, we should be more conservative at first, triggering warnings more than is necessary. We can adjust the criteria once we have more confidence in the system and we've observed actual user behavior. 

- If someone tries fetching cases more frequently than they normally do (e.g. normally someone downloads 2 a day, then they start downloading 20 a day)
- If someone tries to access a case they don't have permission to view
- If someone tries to access a case they don't have permission to view X times in Y timeframe
- If someones started incrementing numbers in search/guessing eFolder IDs.

shanear:
I think we need to start recording eFolder search attempts. Then we can log search counts (successful, permission denied, not found), on the login page.

Thoughts @KHarshawat ?

orischwartz:
Update from standup today: We're going to start this by exposing download stats in the dashboard (#124). This way we see e.g., 

> the top 10 heaviest users per day and how many cases each downloaded.

This will give us a sense of what constitutes normal usage vs excessive/dangerous usage.

shanear:
Added some base monitoring in #185. We should reassess in 1 to 2 months.

shanear:
@cmgiven do you have time to discuss this today?

shanear:
The current monitoring has been deemed sufficent

PR #185, opened 2016-08-02
Add most active users to stats page

shanear:


orischwartz:
This addresses #109 . I wouldn't close that issue out yet, but this is a start.

app/models/stats.rb
```diff
@@ -38,6 +38,10 @@ class Stats
 
     document_errors: lambda do |range|
       Document.failed.where(created_at: range).count
+    end,
+
+    top_users: lambda do |range|
+      Download.top_users(downloads: Download.where(completed_at: range))
```
orischwartz: Is ActiveRecord smart enough not to repeat all these `Download.where(completed_at: range)` calls? Or are we deliberately ignoring that because calculation is on the backend?

PR #193, opened 2016-08-11
Make database connection pool size configurable

### LINKED ISSUE #107: Make database pool size configurable

orischwartz:
Closes #107 

5 is most definitely too low for production and possibly too low for lower environments too.

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -7,7 +7,7 @@
 default: &default
   adapter: postgresql
   encoding: unicode
-  pool: 5
+  pool: <%= ENV["DB_CONN_POOL_MAX_SIZE"] || 5 %>
   timeout: 5000
 
 development:


shanear:
@orischwartz how are we setting this environment variable?

orischwartz:
@shanear We're not setting it yet but it would be set as part of our deployment somewhere here: https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/roles/efolder-express/meta/main.yml#L7

PR #201, opened 2016-08-24
Only check active downloads for cached searches

### LINKED ISSUE #200: Inactive downloads are causing future searches to return "Not Found"
This is the cause of the mysterious "Not Found" errors that are happening.

shanear:
fixes #200

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -40,7 +40,7 @@ class Search < ActiveRecord::Base
   private
 
   def download_scope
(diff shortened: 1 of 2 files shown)


orischwartz:
Makes sense, this is definitely the cause...Note `active` defined as

```
 def self.active
    where(created_at: Download::HOURS_UNTIL_EXPIRY.hours.ago..Time.zone.now)
  end
```

PR #206, opened 2016-08-25
Toggle between median and 95th percentile statistics

cmgiven:


shanear:
❤️ the code, but you need to update the feature spec to account for the new stuff (that's why the build is failing).

bonus points if you write a spec that toggles between median and 95%ile and checks that everything works 💯

cmgiven:
Yup, that's what I get for rushing to get it in before I took off. Will update the specs.

cmgiven:
@shanear: Back to you. Now without breaking all the things.

shanear:
👍 🇺🇸

PR #208, opened 2016-08-26
CSS changes requested by Kavi

cmgiven:
Positioning as an extension to USWDS, given @KHarshawat's view of the general applicability of this. Maybe that's not appropriate? Also note that the class has been named `usa-table-no-header` as an indication of when it is appropriate to use.

### DIFF
--- a/app/assets/stylesheets/_main.scss
+++ b/app/assets/stylesheets/_main.scss
@@ -9,6 +9,27 @@
 @import 'download';
 
 
+// Candidate for USWDS
+.usa-table-no-header {
+  @extend .usa-table-borderless;
+
+  thead {
+    th {
+      background-color: $color-white;
+    }
+  }
+
+  th:first-child,
+  td:first-child {
+    padding-left: 0;
+  }
+
+  th:last-child,
+  td:last-child {
+    padding-right: 0;
+  }
+}
+
 // Candidate for Caseflow Commons
 .cf-loading-indicator {
   display: none !important;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/_main.scss
```diff
@@ -9,6 +9,27 @@
 @import 'download';
 
 
+// Candidate for USWDS
+.usa-table-no-header {
+  @extend .usa-table-borderless;
+
+  thead {
```
shanear: why do we need this section. The plan is that this would be used only for tables without a heading.

cmgiven: User-proofing. If there is a header, it'll appear as just another row.

cmgiven:
Now with no underlines on tab links.

cmgiven:
And a different red. That rounds out this PR. Now ready for you, @shanear.

app/assets/stylesheets/_main.scss
shanear: i like the idea, however, from an accessibility standpoint, it's a little weird. The header will still read as a header, but look like another row.

plus i'm always a fan of less code.

thoughts?

cmgiven: From an accessibility perspective, I can't see someone using a thead unless that was its semantic function, perhaps with some specific stylistic goal. It's not as though someone would use a header for the first row of data arbitrarily.

From my perspective, an additional rule is worth it if it guarantees that the style won't ever result a broken appearance. But that's also probably a function of treating this as something that will be used by more than just you and I.

shanear: Sounds good. Let's keep it in.

shanear:
👍

PR #209, opened 2016-08-27
Stats v2

cmgiven:
Opening this PR primarily to ask Shane a question, although these changes could be merged.

### DIFF
--- a/app/assets/javascripts/stats.js
+++ b/app/assets/javascripts/stats.js
@@ -16,5 +16,4 @@
 
 $(function() {
   window.StatsPanelToggle.bind();
-  window.Dashboard.init();
 });
--- a/app/assets/javascripts/stats.js
+++ b/app/assets/javascripts/stats.js
@@ -11,8 +11,10 @@
 // about supported directives.
 //
 //= require d3
+//= require moment
 //= require_tree ./stats
 
 $(function() {
   window.StatsPanelToggle.bind();
+  window.Dashboard.init();
 });
--- a/app/assets/javascripts/stats.js
+++ b/app/assets/javascripts/stats.js
@@ -10,6 +10,7 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
+//= require d3
 //= require_tree ./stats
 
 $(function() {
--- a/app/assets/javascripts/stats/dashboard.js
+++ b/app/assets/javascripts/stats/dashboard.js
@@ -74,7 +74,7 @@ window.Dashboard = (function(d3, moment) {
       })
 
     charts.on('mousemove', function () {
-      index = maxIndex - Math.floor((d3.event.offsetX) / x.step())
+      index = maxIndex - Math.max(Math.floor((d3.event.offsetX) / x.step()), 0)
       update()
     })
   }
--- a/app/assets/javascripts/stats/dashboard.js
+++ b/app/assets/javascripts/stats/dashboard.js
@@ -137,9 +137,9 @@ window.Dashboard = (function(d3, moment) {
 
   var format = d3.format('.2f')
   function timeFormat(seconds) {
-    return (!seconds ? '??' : format(seconds)) + (seconds > 60 ?
-      ' <span class="ee-stat-unit">min</span>' :
-      ' <span class="ee-stat-unit">sec</span>')
+    return !seconds ? '?? <span class="ee-stat-unit">sec</span>' :
+      seconds > 60 ? format(seconds / 60) + ' <span class="ee-stat-unit">min</span>' :
+      format(seconds) + ' <span class="ee-stat-unit">sec</span>'
   }
 
   // public
--- a/app/assets/javascripts/stats/dashboard.js
+++ b/app/assets/javascripts/stats/dashboard.js
@@ -1,0 +1,149 @@
+window.Dashboard = (function(d3, moment) {
+  // private
+
+  var index = 0
+  var maxIndex = data.length - 1
+  var dateStrings = data.map(function (d, i) {
+    var date = moment.unix(d.key)
+    var str = ''
+    switch (interval) {
+    case 'hourly': str += date.format('HH:mm') + '–' + date.add(59, 'm').format('HH:mm'); break
+    case 'daily': str += date.utc().format('MMMM D'); break
+    case 'weekly': str += 'the week of ' + date.utc().format('MMMM D'); break
+    case 'monthly': str += date.utc().format(i < 12 ? 'MMMM' : 'MMMM YYYY'); break
+    default: str += date.utc().format('X')
+    }
+
+    if (i === 0) { str += ' (so far)'}
+
+    return str
+  })
+  var charts, values, times, dates, table, x, y, bars
+
+  function init() {
+    charts = d3.selectAll('.data-chart')
+      .call(setDataByKey)
+      .append('svg')
+    values = d3.selectAll('.data-value')
+      .call(setDataByKey)
+    times = d3.selectAll('.data-time')
+      .call(setDataByKey)
+    dates = d3.selectAll('.data-date')
+    table = d3.select('#top-users-table')
+
+    x = d3.scaleBand()
+      .domain(d3.range(0, data.length))
+      .paddingInner(0.1)
+
+    y = d3.local()
+    charts.property(y, function (d) {
+      var max
+
+      switch (d.key) {
+      default: max = d3.max(d.values)
+      }
+
+      max = max > 1 ? max : 1
+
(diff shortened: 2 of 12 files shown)


spec/services/download_documents_spec.rb
```diff
@@ -101,7 +101,7 @@
         expect(successful_document.started_at).to eq(Time.zone.now)
         expect(successful_document.completed_at).to eq(Time.zone.now)
 
-        errored_document = Document.all[1]
+        errored_document = Document.last
         expect(errored_document).to be_failed
```
cmgiven: The creation of documents in `dashboard_spec.rb` somehow breaks this test. For some reason, the ordering of the records given by `Document.all` is now reversed, with the failed document located at index 0. But `.first` and `.last` continue to work as expected. Does this make sense to you, @shanear?

cmgiven:
@shanear: Ready for your review.

app/controllers/stats_controller.rb
```diff
@@ -1,11 +1,19 @@
...
+    @stats = {
+      hourly: 0...24,
+      daily: 0...30,
+      weekly: 0...26,
+      monthly: 0...24
+    }[interval].map { |i| Stats.offset(time: Time.zone.now, interval: interval, offset: i) }
+
+    @json = @stats.map { |d| { key: d.range_start.to_f, value: d.values } }.to_json
+
+    @interval = interval
```
shanear: We could memoize @interval and just add `helper_method :interval` to make it accessible in the template (like it's used here: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/downloads_controller.rb#L95)

app/assets/stylesheets/_main.scss
```diff
@@ -77,23 +80,31 @@ table td {
...
+.-timing {
+  // .ee-stat-figure { color: $color-green-light; }
+  svg .bar { fill: $color-green-light; }
+}
 
-  &.-errors {
-    color: $color-secondary;
-  }
+.-errors {
+  // .ee-stat-figure { color: $color-secondary; }
```
shanear: why comment these and not delete them?

app/models/download.rb
```diff
@@ -72,6 +72,10 @@ def complete_documents
     documents.select { |d| !d.pending? }
   end
 
+  def successful_documents
+    documents.select(&:success?)
```
shanear: this should be done via scopes (which will take advantage of SQL speed as opposed to doing the filtering in ruby)

see: http://guides.rubyonrails.org/active_record_querying.html#scopes

app/models/download.rb
```diff
@@ -108,16 +112,25 @@ def user_id_string
     "#{user_id} (Station #{user_station_id})"
   end
 
-  def self.top_users(downloads:)
-    users = downloads.each_with_object({}) do |download, result|
+  def self.all_users(downloads:)
```
shanear: This name of this function is not very clear. It returns a map of the number of downloads for each user in the list of the passed downloads. Maybe change the name to `count_by_user` or something like that? or maybe add a comment explaining the function?

app/models/download.rb
```diff
@@ -108,16 +112,25 @@ def user_id_string
...
     end
+  end
 
+  def self.top_users(downloads:)
+    users = all_users(downloads: downloads)
     sorted = users.sort_by { |_k, v| -v }
     sorted.map { |values| { id: values[0], count: values[1] } }.first(3)
   end
 
+  def self.document_count(downloads:)
```
shanear: documents also have a `completed_at` field. You could filter based on that and remove the need for that function

app/models/stats.rb
```diff
@@ -8,10 +8,18 @@ class Stats
...
+    user_count: lambda do |range|
+      Download.all_users(downloads: Download.where(completed_at: range)).count
+    end,
+
     completed_download_count: lambda do |range|
       Download.where(completed_at: range).count
     end,
 
+    document_count: lambda do |range|
+      Download.document_count(downloads: Download.where(completed_at: range))
```
shanear: Something like this would be simpler/faster
`Document.where(completed_at: range).count`

app/models/stats.rb
```diff
@@ -62,15 +70,65 @@ def values
...
+
   def calculate_and_save_values!
+    return true if complete?
     calculated_values = calculate_values
+    calculated_values[:complete] = Time.zone.now >= range_finish
     Rails.cache.write(cache_id, calculated_values)
     calculated_values
   end
 
+  def self.offset(interval:, time:, offset:)
```
shanear: This function should have a unit spec.

app/assets/javascripts/stats.js
```diff
@@ -10,8 +10,11 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
+//= require d3
+//= require moment
 //= require_tree ./stats
 
 $(function() {
   window.StatsPanelToggle.bind();
+  window.Dashboard.init();
```
shanear: Expecting `data` as a global variable is kind of confusing. We should call this function inside of the erb, and pass data to it as an argument, which would better encapsulate the chart data, and make this object more reusable.

app/views/stats/show.html.erb
```diff
@@ -1,4 +1,8 @@
 <% content_for :head do %>
+  <script>
+    var data = <%= raw @json %>
```
shanear: You could initialize the Dashboard directly here, and be able to pass in the data variable like so:

```
$(function() {
  window.Dashboard.init(data);
});
```

app/assets/javascripts/stats.js
cmgiven: Could we do it here in stats.js? Key is to wait for the onload event to be fired.

shanear: As long as you wrap it in the jQuery block it will wait.

```
$(function() {
  // anything you put here will be called after onload
})
```

app/assets/javascripts/stats/dashboard.js
```diff
@@ -0,0 +1,149 @@
+window.Dashboard = (function(d3, moment) {
```
shanear: There's a lot going on here, and it's pretty tough for me to follow/read through. I'm not really an expert with D3 though, so maybe these are just standard conventions. Not suggesting you change anything, just letting you know.

app/controllers/stats_controller.rb
```diff
@@ -1,11 +1,19 @@
...
-      interval: interval
-    )
+    @stats = {
+      hourly: 0...24,
+      daily: 0...30,
+      weekly: 0...26,
+      monthly: 0...24
+    }[interval].map { |i| Stats.offset(time: Time.zone.now, interval: interval, offset: i) }
+
+    @json = @stats.map { |d| { key: d.range_start.to_f, value: d.values } }.to_json
```
shanear: Since the format of the JSON here is pretty tightly coupled to your D3 implementation, I'd create a helper function for the JSON formatting here something like `d3_json(stats: stats)`. It makes that more clear.

shanear:
The D3 stuff is sweet! I added my notes. After you deploy, you can go over the UI with Kavi.

cmgiven:
@shanear: Commits in response to your comments now in.

app/models/stats.rb
shanear: don't see a unit spec for this function

cmgiven: You have suggestions for what you want to see tested? The code is covered by additions to the calculate_all context.

orischwartz:
Deployed this branch to UAT so we can evaluate better
image

cmgiven:
Hmm. These numbers look odd. 0 users and 0 completed downloads, yet 3,397 documents retrieved? Possible that we just need to clear Redis or something? Or maybe there are deeper issues, at least with the historical data.

spec/services/download_documents_spec.rb
orischwartz: This was driving me nuts too. ActiveRecord does different things behind the scenes.

* [first/last](http://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-last): If no order is defined it will order by primary key.
* [all](http://stackoverflow.com/questions/4986496/activerecord-find-all-not-sorting-by-id/4986613#4986613): Order not guaranteed unless the model defines default_scope (which we don't)

Your way of using first and last is deterministic.

cmgiven: Glad to know there is some logic in play here. Thanks.

orischwartz:
@cmgiven I think that's expected because we didn't add the relevant columns for tracking this until https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/178/ so any data from before then would be unreliable.

PR #214, opened 2016-09-08
Add support for filtering vbms doctypes

### LINKED ISSUE #211: Filter out XML document types
[DBQs](http://www.benefits.va.gov/compensation/dbq_disabilityexams.asp) in XML format should not be returned to the user. 

These are sent back with a docType of `999981 - C&P Exam XML` and have a corresponding PDF with a docType of `356 - C&P Exam`.

orischwartz:
And filter out the XML versions of `999981 - C&P Exam` docs that are also sent as PDFs.

Closes #211

### DIFF
--- a/app/services/download_documents.rb
+++ b/app/services/download_documents.rb
@@ -130,12 +130,10 @@ class DownloadDocuments
   end
 
   def self.ignored_doc_types
-    ignored_types = []
-
-    # C&P Exam (DBQ) sent back as both XML and PDF, ignore the XML 999981
-    ignored_types << "999981"
-
-    ignored_types
+    [
+      # C&P Exam (DBQ) sent back as both XML and PDF, ignore the XML 999981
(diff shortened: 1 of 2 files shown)


app/services/download_documents.rb
```diff
@@ -128,4 +128,20 @@ def before_document_download(document)
   def before_package_contents
     # Test hook for testing race conditions
   end
+
+  def self.ignored_doc_types
+    ignored_types = []
+
+    # C&P Exam (DBQ) sent back as both XML and PDF, ignore the XML 999981
+    ignored_types << "999981"
```
cmgiven: Reason not to just do this?

```ruby
def self.ignored_doc_types
  ignored_types = [
    # C&P Exam (DBQ) sent back as both XML and PDF, ignore the XML 999981
    "999981"
  ]
end
```

PR #215, opened 2016-09-08
Secondary sort by id for docs with same date

### LINKED ISSUE #213: User reports that documents with same receipt date are not ordered the same as VBMS
We store receipt date as a date with no time so documents with the same value will be ordered arbitrarily. 

Possible solutions:
* ~~Determine if VBMS stores the exact timestamp and use that instead~~ Per https://github.com/department-of-veterans-affairs/connect_vbms/issues/120, VBMS doesn't send the exact timestamp. 
* If VBMS returns the files in the expected order, their ids will be sequential. Use a compound ordering on the document sorting as follows:
  - Receipt date 
  - If equal, sort by document id

See #130 for original implementation.

orischwartz:
closes #213 ...I hope. If not, there's no reliable way to reproduce the order of documents in VBMS.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -12,7 +12,7 @@ class Download < ActiveRecord::Base
   TIMEOUT = 10.minutes
   HOURS_UNTIL_EXPIRY = 24
 
-  has_many :documents, -> { order(received_at: :desc) }
+  has_many :documents, -> { order(received_at: :desc, id: :desc) }
 
   after_initialize do |download|
     if download.file_number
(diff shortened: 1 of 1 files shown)


app/models/download.rb
```diff
@@ -12,7 +12,7 @@ class Download < ActiveRecord::Base
   TIMEOUT = 10.minutes
   HOURS_UNTIL_EXPIRY = 24
 
-  has_many :documents, -> { order(received_at: :desc) }
+  has_many :documents, -> { order(received_at: :desc, id: :desc) }
```
cmgiven: Can we add a comment referencing the original issue?

orischwartz:
Confirmed after deployment using data supplied by test user: VBMS will send them in order of receipt so this works.

PR #218, opened 2016-09-16
Match VBMS ordering of documents

### LINKED ISSUE #213: User reports that documents with same receipt date are not ordered the same as VBMS

orischwartz:
fixes #213, apparently we're displaying docs with same receipt date backwards from VBMS.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -14,7 +14,7 @@ class Download < ActiveRecord::Base
 
   # sort by receipt date; documents with same date ordered as sent by vbms; see
   # https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/213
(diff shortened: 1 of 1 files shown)


shanear:
👍

PR #220, opened 2016-09-16
Increase history from 1->3 days

### LINKED ISSUE #219: Show 3 days of download history (currently just 1)
- [x] Edit [expiry duration](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/download.rb#L13) in the `Download` model

> This is the only required change. 

- [x] Determine whether it makes sense to [clear temporary](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/config/sidekiq_cron.yml#L16) files less frequently

> Leaving this as is for now -- the application server's file system is treated as ephemeral anyway. S3 cache will be used instead.

- [x] Determine whether it makes sense to expire S3 cache less frequently (currently set to 3 days).

> Leaving this as is. The expiry applies from the time the download was `created_at`, which will elapse *before* the 3 day S3 expiry (files are saved after download creation).

orischwartz:
Can't decide if this would be more readable as `3.days/1.hour`...Certainly more rubyist. 

Kinda surprised there's no Ruby equivalent to Java's `TimeUnit.DAYS.toHours(3)`

Closes #219


orischwartz:
Un-assigning, need to updated specs

orischwartz:
Forgot to assign a reviewer...

shanear:
👍

PR #227, opened 2016-09-21
Use correct screenreader class

shanear:


shanear:
I'm not sure what happened to the other PR. It auto merged or something... I think it was due to something wonky with git on my localbox

PR #232, opened 2016-09-29
Update README to explain what efolder express does

shanear:


KHarshawat:
A few changes but then looks good to go!

PR #235, opened 2016-10-06
instructions to run redis & pg on startup

amprokop:
Instruct devs to use brew info to run services on startup.

h/t @shanear


shanear:
:+1:

PR #242, opened 2016-10-10
Updates .zip filename to match format  "LastName, Firstname - Last4SSN"

### LINKED ISSUE #55: Use a unique and more descriptive filename than "documents.zip"

### LINKED ISSUE #223: Use "lastName, firstName - lastfour" convention for packaged zip files
See also #55.

For now just change the zip naming convention. 

- [ ] Modify BGSService to return e.g., array of [firstName, lastName]
- [ ] Store first and last names separately in `Download` model.
- [ ] Add convenience method to Download for full `veteran_name`.
- [ ] Update zip-naming format in `Download#package_filename`

amprokop:
Changes download filename to "LastName, Firstname - Last4SSN". This was significantly more difficult than first anticipated! Adds a DB migration to store the last four digits of a veteran's ssn, and the first and last name in separate columns. A bit of tech debt is also cleaned up here and there for good measure.

Fixes #223
Fixes #55 

Reviewer: @shanear 

Test plan:
modified and added unit tests
locally tested that the file download now occurs with the expected format
tested the migration and rollback several times
image
TODO: currently blocked on testing this with the live DB. @shanear, can you help me figure out how to proceed?

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -18,7 +18,10 @@ class Download < ActiveRecord::Base
 
   after_initialize do |download|
     if download.file_number
-      download.veteran_name ||= download.demo? ? "TEST" : Download.bgs_service.fetch_veteran_name(download.file_number)
+      demo_info = veteran_info = { "first_name": "Test", "last_name": "User", "last_4_ssn": "1224" }
+      veteran_info = download.demo? ? demo_info : Download.bgs_service.fetch_veteran_info(download.file_number)
+      veteran_info ||= {}
+      download.veteran_name ||= "#{veteran_info[:first_name]} #{veteran_info[:last_name]}"
     end
   end
 
@@ -57,7 +60,7 @@ class Download < ActiveRecord::Base
   end
 
   def case_exists?
-    !Download.bgs_service.fetch_veteran_name(file_number).nil?
+    !Download.bgs_service.fetch_veteran_info(file_number).nil?
   rescue
     false
   end
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -20,7 +20,7 @@ class Download < ActiveRecord::Base
     if download.file_number
       veteran_info = Download.bgs_service.fetch_veteran_info(download.file_number)
       if veteran_info
-        download.veteran_name ||= "#{veteran_info[:veteran_first_name]} #{veteran_info[:veteran_last_name]}"
+        download.veteran_name ||= "#{veteran_info['veteran_first_name']} #{veteran_info['veteran_last_name']}"
       end
     end
   end
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -19,7 +19,9 @@ class Download < ActiveRecord::Base
   after_initialize do |download|
     if download.file_number
       veteran_info = Download.bgs_service.fetch_veteran_info(download.file_number)
-      download.veteran_name ||= "#{veteran_info[:first_name]} #{veteran_info[:last_name]}"
+      if veteran_info
+        download.veteran_name ||= "#{veteran_info[:veteran_first_name]} #{veteran_info[:veteran_last_name]}"
+      end
     end
   end
 
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -30,7 +30,7 @@ class Download < ActiveRecord::Base
   end
 
   def demo?
-    file_number =~ /DEMO/
+    Download.bgs_service.demo?(file_number)
   end
 
   def time_to_fetch_manifest
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -17,12 +17,26 @@ class Download < ActiveRecord::Base
   has_many :documents, -> { order(received_at: :desc, id: :asc) }
 
   after_initialize do |download|
-    if download.file_number
+    # TODO(alex): it doesn't appear as though the fake is used except for tests.
+    # is that right?
+    # TODO(alex): when running a migration on the Downloads table, this hook will
+    # be called for every row in the table. kind of hacky but
+    # can we prevent lots of unneccessary requests if we
+    # only call the bgs service for downloads which haven't yet been started?
+    if download.file_number && !download.started_at
       veteran_info = Download.bgs_service.fetch_veteran_info(download.file_number)
       if veteran_info
-        download.veteran_name ||= "#{veteran_info['veteran_first_name']} #{veteran_info['veteran_last_name']}"
+        download.veteran_first_name = veteran_info["veteran_first_name"]
+        download.veteran_last_name = veteran_info["veteran_last_name"]
+        download.veteran_last_four_ssn = veteran_info["veteran_last_four_ssn"]
       end
     end
+  end
+
+  # TODO: (alex) is this confusing to switch this from data to a dynamic method
+  # ...or is it good Ruby?
+  def veteran_name
+    "#{veteran_first_name} #{veteran_last_name}" if veteran_last_name
   end
 
   def self.active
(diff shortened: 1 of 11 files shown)


app/models/download.rb
```diff
@@ -17,17 +17,34 @@ class Download < ActiveRecord::Base
   has_many :documents, -> { order(received_at: :desc, id: :asc) }
 
   after_initialize do |download|
-    if download.file_number
-      download.veteran_name ||= download.demo? ? "TEST" : Download.bgs_service.fetch_veteran_name(download.file_number)
+    # TODO(alex): it doesn't appear as though the fake is used except for tests.
+    # is that right?
```
shanear: the fake is also used for running the app in "demo" mode

app/models/download.rb
```diff
@@ -17,17 +17,34 @@ class Download < ActiveRecord::Base
...
+      if veteran_info
+        download.veteran_first_name = veteran_info["veteran_first_name"]
+        download.veteran_last_name = veteran_info["veteran_last_name"]
+        download.veteran_last_four_ssn = veteran_info["veteran_last_four_ssn"]
+      end
     end
   end
 
+  # TODO: (alex) is this confusing to switch this from data to a dynamic method
+  # ...or is it good Ruby?
```
shanear: i think its fine. Technically, they're both just methods (even the ActiveRecord setters), so that's not changing, just the implementation of those methods, which is fine.

app/models/download.rb
```diff
@@ -89,7 +106,10 @@ def s3_filename
   end
 
   def package_filename
-    "#{veteran_name.gsub(/\s*/, '').downcase}-#{created_at.to_formatted_s(:filename)}.zip"
+    # TODO: (alex): should we still downcase the names?
```
shanear: Nope, it looks like we want the cap names in there

db/migrate/20161011020910_add_first_name_last_name_ssn_to_downloads.rb
```diff
@@ -0,0 +1,28 @@
+class AddFirstNameLastNameSsnToDownloads < ActiveRecord::Migration
+  def up
+    # we could be more clever here by renaming a name column
+    # instead of adding two columns, but this seems fine.
+    add_column :downloads, :veteran_last_name, :string
+    add_column :downloads, :veteran_first_name, :string
+    add_column :downloads, :veteran_last_four_ssn, :string
+
+    Download.all.each do |download|
```
joofsh: My Rails info may be old so @shanear can correct me, but `all` would load all records into memory. I think you'd want to use `Download.find_each` here to batch load the records.

app/models/download.rb
amprokop: sweet will resolve

app/models/download.rb
amprokop: word thanks

app/models/download.rb
```diff
@@ -17,17 +17,34 @@ class Download < ActiveRecord::Base
...
 
   after_initialize do |download|
-    if download.file_number
-      download.veteran_name ||= download.demo? ? "TEST" : Download.bgs_service.fetch_veteran_name(download.file_number)
+    # TODO(alex): it doesn't appear as though the fake is used except for tests.
+    # is that right?
+    # TODO(alex): when running a migration on the Downloads table, this hook will
+    # be called for every row in the table. kind of hacky but
+    # can we prevent lots of unneccessary requests if we
+    # only call the bgs service for downloads which haven't yet been started?
```
amprokop: yeah, that's sensible. will do.

amprokop: `if download.file_number && !download.started_at` 
this line ^ is problematic because when we run the migration, any orphan rows in the downloads table (like, if the sidekiq worker wasn't started when the download was requested) will kick off BGS service calls. do you think that's ok? i'll try to think of a better solution in the meantime

app/models/download.rb
shanear: @amprokop yeah

app/models/download.rb
shanear: `veteran_info?` could be a method that checks for the presence of all those fields. Something like this:

```
def missing_veteran_info?
  file_number && (!veteran_first_name || !veteran_last_name || !veteran_ssn_last_four)
end
```

app/models/download.rb
amprokop: looking into a fix now, no need to look into this.

app/models/download.rb
amprokop: @shanear — sorry i don't think i was clear. 

i think the problem with that solution is that if we have any sort of orphan rows in the production database, where a download was created and didn't get kicked off for whatever reason, we'll then fire off BGS service calls for all of those rows when we run the migration (since they will have a value for `file_number` but none of the veteran information). this might not be a problem if a. we don't think we have many or any of those rows in prod or b. we don't mind maybe kicking off lots of BGS requests.

shanear: Good thought.

Looking at the stats, there should be little to no orphan downloads in prod, and I also don't mind kicking off some BGS requests if there happen to be a few. Should be negligible.

shanear: @amprokop I like that idea. I never liked using `after_initialize` anyways. It's hard to control.

joofsh: Yea seems excessive to do BGS API calls  on every initialization of the row unless we were purposely doing that to keep the veteran's name/ssn up-to-date?

Based on the newest code seems like `before_create` might be be best? Running this in the `before_create` ensures they immediately get persisted to the DB.

amprokop: Looks like `before_create` worked in manual testing a few times. After looking at `Search#perform`, i don't see any obvious pitfalls.

shanear:
@amprokop yep! i can do that.

Also, I think your build is failing because you have a test that is starting a sidekiq job. You'll need to stub that out. There are some other examples where we do that in the code.

amprokop:
@shanear — huh. i don't quite understand why the test I added would cause a failure (looks like line 160 of `features/download_spec`. it's pretty similar to other tests in that file that don't appear to have anything stubbed out. any thoughts on why this one might be causing the failure?

amprokop:
@shanear — otherwise, this is ready for review!

shanear:
@amprokop we stub out the non-demo jobs here https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/spec/features/download_spec.rb#L12-L13.

But since your triggering `DemoGetDownloadManifestJob`, the stubs aren't taking effect, and it's trying to add the sidekiq job to redis, which doesn't exist in travis.

I committed a fix, feel free to change it if you feel like there's a better way

db/migrate/20161011020910_add_first_name_last_name_ssn_to_downloads.rb
```diff
@@ -0,0 +1,35 @@
...
+        # space. However, historical Download rows are kept
+        # for record-keeping, so it's probably acceptable.
+        first, last = download.veteran_name.split(' ', 2)
+        download.update(veteran_first_name: first, veteran_last_name: last)
+      end
+    end
+
+    remove_column :downloads, :veteran_name
+  end
+  def down
```
shanear: have you tested this up/down migration manually? no need for an automated test but we should have tested it

amprokop: yep, i've tested it

app/models/download.rb
```diff
@@ -30,6 +48,10 @@ def demo?
     file_number =~ /DEMO/
   end
 
+  def missing_veteran_info?
```
amprokop: hmm i think it does

amprokop: maybe not. will do!

app/services/bgs_service.rb
```diff
@@ -4,10 +4,21 @@
 class BGSService
   cattr_accessor :user
 
-  def self.fetch_veteran_name(file_number)
+  def self.parse_veteran_info(veteran_data)
```
amprokop: ok, will create!

amprokop:
Acknowledged. Will keep that in mind. I'll ask you for more details next
week and let's add this to the README

PR #249, opened 2016-10-13
Decrease session duration from 2 weeks->24 hours

### LINKED ISSUE #182: Enforce session timeout of 24 hours
As per findings of SCA

orischwartz:
Closes #182 (required per SCA audit findings).


orischwartz:
@shanear I think @gmorgana-va is reporting that we will have completed this by end of day so we should deploy this and https://github.com/department-of-veterans-affairs/caseflow-certification/pull/336 as soon as possible.

PR #251, opened 2016-10-13
Return 404 on non-existing download attempt

joofsh:
Testing:
- All tests pass
- Added tests that specifically test a non-existing download 
- Manually testing in the browser works as expected:
screen shot 2016-10-13 at 5 54 15 pm

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -58,11 +58,12 @@ class DownloadsController < ApplicationController
     @download_documents = DownloadDocuments.new(download: downloads.find(params[:id]))
     @download_documents.fetch_zip_from_s3
 
-    send_file @download_documents.zip_path
+    file_exists = @download_documents.zip_exists_locally?
+    file_exists ? send_file(@download_documents.zip_path) : record_not_found
   end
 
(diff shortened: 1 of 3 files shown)


app/controllers/downloads_controller.rb
```diff
@@ -58,11 +58,12 @@ def download
     @download_documents = DownloadDocuments.new(download: downloads.find(params[:id]))
     @download_documents.fetch_zip_from_s3
 
-    send_file @download_documents.zip_path
+    file_exists = @download_documents.zip_exists_locally?
+    file_exists ? send_file(@download_documents.zip_path) : record_not_found
```
joofsh: This might be overkill. I was trying to figure out if there was a scenario where the download could exist in the database but the zip wouldn't exist locally or in S3?

Looking at the code this seems like a feasible path if you manually navigated to `/downloads/123/download`, since the [download method](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/251/files#diff-5b1717b5c25ba511288b5bb06c831667R81) doesn't scope to downloads that are in the `complete_success` or `complete_with_errors` states

PR #253, opened 2016-10-14
Use document type names instead of vbms filenames

### LINKED ISSUE #237: Include document type in filename
Use the [VBMS docType listing](https://github.com/department-of-veterans-affairs/appeals-pm/blob/master/other%20systems/VBMS/document_types.txt) to include the docTypes in the file names (the human-readable version, not the docType ordinals).

This can be tricky because that listing contains lots of characters that can't be used in filenames. Would be difficult to exactly match the VBMS naming conventions.

A few VBMS examples (from sanitized screenshots):
image
image

More discussion in #158

shanear:
screen shot 2016-10-14 at 11 48 31 am

screen shot 2016-10-14 at 11 48 52 am

screen shot 2016-10-14 at 11 49 30 am

Also, update number ordering to start with 00010 instead of 00000.

fixes #237


.rubocop.yml
```diff
@@ -24,7 +24,7 @@ Style/Documentation:
   Enabled: false
 
 Metrics/LineLength:
-  Max: 120
+  Max: 200
```
shanear: These long Document type names are making it hard to keep the 120 limit. Increased to 200, but we should still aim for small line sizes.

app/models/document.rb
```diff
@@ -1,14 +1,529 @@
...
+    92  => "SF 502 Clinical Record - Narrative Summary",
+    93  => "SSA/SSI Letter",
+    95  => "Statement of Case (SOC)",
+    96  => "Summaries of Inpatient Treatment",
+    97  => "Supplemental Statement of Case (SSOC)",
+    98  => "Surgeon Generals Office (SGO) Extracts",
+    99  => "VA 10-1000 Hospital Summary and/or the Compensation and Pension Exam Report",
+    999_981 => "C&P Exam XML"
+  }.freeze
+
```
shanear: agreed. added that change

.rubocop.yml
joofsh: Agreed it would be ideal to stay below 120. Let's internally try and enforce that in PRs when reasonable

PR #254, opened 2016-10-14
Show document ID for errored downloads

### LINKED ISSUE #169: When a document isn't able to be downloaded, provide Doc ID
The most useful information for a user to find a missing document is the VBMS document ID. We should include this in the errors tab with the filename.

shanear:
screen shot 2016-10-14 at 2 51 33 pm

fixes #169

### DIFF
--- a/app/views/downloads/_progress.html.erb
+++ b/app/views/downloads/_progress.html.erb
@@ -86,6 +86,9 @@
         <tr>
           <th scope="col"><span class="usa-sr-only">Status</span></th>
           <th scope="col">Filename</th>
+          <% if current_tab == "errored" %>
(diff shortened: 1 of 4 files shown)


app/views/downloads/_progress.html.erb
```diff
@@ -97,6 +100,9 @@
               <%= svg_icon(document.download_status_icon) %>
             </td>
             <td class="ee-filename-row"><%= document.type_name %></td>
+            <% if current_tab == "errored" %>
+              <td class="ee-doc-id-row"><%= document.filename_doc_id %></td>
```
joofsh: Nvm I see it on master, didn't exist on my branch

PR #255, opened 2016-10-17
Show modal warning users when document fetch results in some errors

### LINKED ISSUE #160: Add modal warning users when they "Download Anyway" if a Download has errored documents.
broken off from #51

amprokop:
Shows modal for the error case when some documents couldn't be fetched.

Fixes #160.

final_efolder_modal

@lakohl — added this GIF to give you an idea of what it currently looks like, and to obtain real content. wanna get that to me when you can?

Test plan: fake a download error 

```
<div class="cf-app-segment cf-app-segment--alt">
  <% if @download.complete? %>
    <% if true %>
```

tab through, check that focus is trapped
check that modal info is marked by aria-labelledby and aria-describedby
check that esc key closes modal
check that focus returns to the previously focused element when modal is closed

cc @KHarshawat

### DIFF
--- a/app/assets/javascripts/application/modal.js
+++ b/app/assets/javascripts/application/modal.js
@@ -3,7 +3,6 @@
 window.Modal = (function($) {
 
   function openModal(e) {
-    debugger;
     e.preventDefault();
     var target = $(e.target).attr("href");
     $(target).addClass("active");
@@ -30,7 +29,6 @@ window.Modal = (function($) {
   // public
   return {
     bind: function() {
-      debugger;
       $('.cf-action-openmodal').on('click', openModal);
       $('.cf-modal').on('click', closeModal);
       $(window).on('keydown', onKeyDown);
--- a/app/assets/javascripts/application/modal.js
+++ b/app/assets/javascripts/application/modal.js
@@ -2,10 +2,17 @@
 
 window.Modal = (function($) {
 
+  // set by openModal. focus will be returned to this element
+  // after the modal is closed.
+  // important for a11y, so keyboard users have a sensible flow.
+  var returnEl;
+
   function openModal(e) {
     e.preventDefault();
     var target = $(e.target).attr("href");
     $(target).addClass("active");
+    $('.cf-modal-title').focus();
+    returnEl = e.target;
   }
 
   function closeModal(e) {
@@ -15,6 +22,9 @@ window.Modal = (function($) {
     if ($(e.target).hasClass("cf-modal") || $(e.target).hasClass("cf-action-closemodal")) {
       e.preventDefault();
       $(e.currentTarget).removeClass("active");
+    }
+    if (returnEl) {
+      $(returnEl).focus();
     }
   }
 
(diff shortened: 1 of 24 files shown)


amprokop:
@shanear — waiting on content from @lakohl @abbyraskin on this, but should be ready for review. thanks!

amprokop:
@shanear — just noticed that CI fails a check. weird bc it passes on my machine. i'll take a look, but the application code itself should still be ready for review.

shanear:
@amprokop my gut tells me it could have to do with a version discrepancy with PhantomJS

spec/features/download_spec.rb
```diff
@@ -293,6 +293,36 @@
...
+    page.execute_script("window.Modal.bind();")
+    expect(page).to have_content "Download anyway"
+    expect(page).to have_no_content "Download incomplete eFolder?"
+
+    within first(".usa-alert-body") do
+      click_on "Download anyway"
+    end
+
+    # check that focus is correctly moved to the modal
+    is_focus_on_modal = page.evaluate_script("document.activeElement.className").include? "cf-modal-startfocus"
```
amprokop: @joofsh — this is the test i was considering removing

spec/features/download_spec.rb
```diff
@@ -293,6 +293,36 @@
...
+    expect(is_focus_on_modal).to be true
+    expect(page).to have_content "Download incomplete eFolder?"
+
+    click_on "Go back"
+
+    # check that focus is correctly moved back to the "Download Anyway" button
+    # after the modal is closed
+    is_focus_on_button = page.evaluate_script("document.activeElement.className").include? "cf-action-openmodal"
```
amprokop: @joofsh @shanear — i also propose to remove this test

amprokop:
interesting. I updated PhantomJS and removed the tests relating to focused elements — and the Travis CI build still fails. It looks like the modal isn't popping up in the phantomJS browser maybe. More to investigate.

lakohl:
@amprokop here copy for the modal:
We encountered errors when retrieving some documents and they won’t be included in the eFolder download. If you elect to 'Download anyway', you may want to retrieve these files individually from VBMS.

amprokop:
@lakohl — one more question too. The content above ^^ uses single quotes. I assume we want double quotes for this, right?

app/assets/javascripts/application/modal.js
```diff
@@ -0,0 +1,92 @@
...
+    var escKey = (e.which === 27);
+    var tabKey = (e.which === 9);
+    var tabShift = (e.shiftKey && e.keyCode == 9);
+
+    if (escKey) {
+      $('.cf-modal').trigger('click');
+    }
+
+    if (tabKey) {
+      if ($('.cf-modal-endfocus').is(':focus')) {
```
amprokop: ahh this actually isn't scoped, you're right

amprokop:
@shanear — i made some changes to scope this more tightly. going to update from master, retest, then merge since you approved this.

amprokop:
image

PR #256, opened 2016-10-17
Improve time estimation for reamining downloads

### LINKED ISSUE #221: More accurate time remaining estimates
[Currently](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/download.rb#L136) we're just taking the average time it took to fetch completed files _in the current case_ and multiplying by the number of files outstanding.

Could make more accurate by keeping track of all files downloaded since e.g., application startup. (Could distribute the calculation across all app servers by storing sum and count values in a distributed cache like Redis.)

joofsh:
- Update logic for calculating remaining time for downloading files to
  reference the latest 100 downloads
- Cache calculation of average download rate to redis. Update
  calculations on the 30sec mark
- Add TimeUtil with floor method for calculating closest window for
  caching values

Fixes #221

Testing:
- Updating download unit tests
- Added document unit tests
- All existing tests pass
- Manually tested in browser and works as expected. In the gif below, the historical average download rate is set to ~1 minute. This way you can see it quickly updating as it downloads. Notice the steady rate that the estimation changes:
document_time_estimates


app/models/document.rb
```diff
@@ -1,10 +1,19 @@
+# frozen_string_literal: true
 class Document < ActiveRecord::Base
   include DocumentTypes
 
+  AVERAGE_DOWNLOAD_RATE_LIMIT = 100
+  AVERAGE_DOWNLOAD_RATE_CACHE_EXPIRATION = 30.seconds
+  AVERAGE_DOWNLOAD_RATE_CACHE_KEY = "historical-average-download-rate".freeze
```
joofsh: I tried to mimicking the redis key naming convention I see for `stats.rb` in caseflow-common. Does this work for you?  If we go with underscore syntax I could switch this to a symbol (feels a little weird as a lowercase string?)

spec/models/document_spec.rb
```diff
@@ -1,6 +1,11 @@
 require "rails_helper"
 
 describe Document do
+  before do
+    Document.delete_all
+    Rails.cache.clear
```
joofsh: todo: eventually I'd like to move this up to a higher level. We should be wiping the db & redis between each test run

spec/models/document_spec.rb
```diff
@@ -35,4 +40,59 @@
...
+
+      it "returns value from redis" do
+        # Initial call, saving value to cahce
+        expect(subject).to eq(4.minutes.round(2))
+        expect(@rails_cache_write_called).to be_truthy
+
+        @rails_cache_write_called = false
+        # call again, this time reading from cache
+        subject
+        expect(@rails_cache_write_called).to be_falsey
```
joofsh: Is there a prettier way to test this with rspec? I basically want to test that it does write to cache the first time, but does not on subsequent calls

app/models/document.rb
```diff
@@ -32,8 +41,46 @@ def type_name
     TYPES[doc_type.to_i] || vbms_filename
   end
 
+  def self.historical_average_download_rate
+    time_floor = TimeUtil.floor(AVERAGE_DOWNLOAD_RATE_CACHE_EXPIRATION)
+    cache = Rails.cache.read(AVERAGE_DOWNLOAD_RATE_CACHE_KEY)
+
+    return cache[:value] if cache && cache[:updated_at] == time_floor
```
joofsh: This 100% works properly, but just to offer you another potential solution: we could use the redis built in expire feature.

This way when we write to redis we would simply set the numeric value directly (rather than an object) and use the `expires_in` option (`Rails.cache.write(key, value, { expires_in: X })` ). Calculating the `expires_in` value is a little more complex but it would be more "the redis way"

spec/models/document_spec.rb
shanear: I agree. The only reason I haven't is that we only really want to do it for feature tests. Most unit tests don't use the DB and making a ton of DB calls every test run is expensive.

Maybe a good compromise is to make a `reset_application!` test helper method?

app/models/document.rb
shanear: Ooh, I like that cause it's less code. But we're also leaning pretty heavily on that logic working as we expect that we aren't testing (we don't bring in Redis in the tests).

I'm leaning towards keeping this code, but I'll defer to whatever you think is best.

spec/models/document_spec.rb
shanear: You could seperate this into two tests:

- The first test writes a key/value pair to the cache (maybe an outlandish value), with a time stamp of `29.seconds.ago`, and verifies that it uses that value.
- The second test test writes the same value except with a timestamp of `30.seconds.ago` and verifies that the correct value is computed, written and used instead of the outlandish value.

What do you think?

Another side note: we should configure the `Rails.cache` in the tests to use an in-memory cache.

app/models/document.rb
shanear: I think we should choose symbol or string keys now, and not worry about matching stats if that's not the best way.

Strings are nice because Redis will eventually use a string anyway and it's nice not to have the cognitive dissonance. However, symbols are nice because they feel more appropriate as hash keys in rails (and maybe have some performance gains?).

app/models/document.rb
```diff
@@ -32,8 +41,46 @@ def type_name
     TYPES[doc_type.to_i] || vbms_filename
   end
 
+  def self.historical_average_download_rate
```
shanear: any reason for having this method be:

`def self.method_name`

and the other class methods be like this?
```
class << self
  def method_name
end 
```

spec/models/document_spec.rb
```diff
@@ -35,4 +40,59 @@
...
+      end
+    end
+
+    context "#calculate_historical_average_download_rate" do
+      subject { Document.calculate_historical_average_download_rate }
+
+      it { is_expected.to eq(4.minutes.round(2)) }
+    end
+
+    context "#calculate_and_save_historical_average_download_rate" do
```
shanear: Also, the format for method names in specs is:

```
context ".class_method" do
end

context "#instance_method" do
end
```

see: http://betterspecs.org/#describe

spec/models/document_spec.rb
```diff
@@ -35,4 +40,59 @@
       it { is_expected.to eq("VA 10-1000 Hospital Summary andor the Compensation and Pension Exam Report-20150103-ABC123-DEF123-GHI456.pdf") }
     end
   end
+
+  context "calculate_historical_average_download_rate" do
+    before do
+      5.times do
+        # 3 minute average
+        Document.create!(started_at: 4.minutes.ago,
```
shanear: This is a lot of DB IO for these tests. Is there a way we can inject unsaved documents into the calculation methods, and only have 1 test actually test the DB call integration?

ie:
```
def calculate_average_download_rate(downloads)
  # all the maths
end

def calculate_historical_average_download_rate
  calculate_average_download_rate(ordered_by_completed_at.limit(AVERAGE_DOWNLOAD_RATE_LIMIT))
end
```

spec/models/document_spec.rb
shanear: This needs a `!`

spec/models/document_spec.rb
shanear: then the tests for `.calculate_and_save_historical_average_download_rate` and `.historical_average_download_rate` can just use 1 document in the DB (because you aren't real testing the algorithm in those tests)

lib/utils/time_util.rb
```diff
@@ -0,0 +1,9 @@
+module TimeUtil
```
shanear: The general idea behind `/lib` is that it should contain code that "transcends" the app and needs to be used in tests, CI, or something like that.

If we start adding app functionality to lib (even though it's sort of generic), we'll constantly have to be asking ourselves "is this generic enough for lib?". I'm willing to discuss further though if you don't agree.

app/models/document.rb
```diff
@@ -32,8 +41,46 @@ def type_name
...
+    # This calculates the historical average rate at which files are downloaded
+    # This takes the time spend downloading all files over the past hour and
+    # divides by the total number of files in that period
+    def calculate_historical_average_download_rate
+      documents = ordered_by_completed_at.limit(AVERAGE_DOWNLOAD_RATE_LIMIT)
+
+      count = documents.count
+      return nil unless count
+
+      total_time = documents.inject(0) do |total, document|
```
shanear: `#inject` is so pro 😎

spec/models/document_spec.rb
joofsh: ✔️

app/models/document.rb
joofsh: Especially if we aren't using redis for tests, I'm inclined to agree with you and leave the code be 👍

spec/models/document_spec.rb
joofsh: ✔️ https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/256/commits/9e23493e0fd24e1010c53bd6d67a7f1e711670a4

lib/utils/time_util.rb
shanear: Ohh ok, I see. `/app/util` would work, but this seems fine in `/lib/util` as well.

PR #268, opened 2016-10-19
Allow APP_ENV to be overridden through ENV

askldjd:
By allowing APP_ENV to be overridden through environment variables, Jenkins can inject custom properties in the pipeline. This way, we can use the same Jenkinsfile for different AWS environment.

jenkins-prop


shanear:
awesome 👍

I think we may also want to allow it to deploy simultaneously to all non-prod environments too.

PR #270, opened 2016-10-20
Only poll for download status when fetching manifest

### LINKED ISSUE #33: page should stop polling (refresh) after failing to find a file

joofsh:
Fixes #33 

Testing:
- Added feature tests to verify it does and does not call  `window.DownloadStatus.init()` as expected
- All current tests pass
- Manual browser testing, everything worked as expected

### DIFF
--- a/app/assets/javascripts/application/download_status.js
+++ b/app/assets/javascripts/application/download_status.js
@@ -3,6 +3,7 @@ window.DownloadStatus = (function($) {
 
   // public
   return {
+    intervalID: null,
     recheck: function() {
       if (id) {
         $.getJSON("/downloads/" + id).then(function(download) {
@@ -15,7 +16,7 @@ window.DownloadStatus = (function($) {
       id = downloadId;
       currentStatus = downloadStatus;
 
-      window.setInterval(this.recheck, 1000);
+      this.intervalID = window.setInterval(this.recheck, 1000);
     }
   };
 })($);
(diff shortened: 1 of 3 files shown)


spec/features/download_spec.rb
```diff
@@ -68,6 +68,7 @@
     expect(page).to have_content "Stan Lee (1234)"
     expect(page).to have_content "We are gathering the list of files in the eFolder now"
     expect(page).to have_current_path(download_path(@download))
+    expect(page.evaluate_script("window.DownloadStatus.intervalID")).to be_truthy
```
joofsh: Do you know a better way to test this? It feels a little awkward to grab values from JS and test them in rspec, but maybe this is normal?

app/assets/javascripts/application/download_status.js
```diff
@@ -15,7 +16,7 @@ window.DownloadStatus = (function($) {
       id = downloadId;
       currentStatus = downloadStatus;
 
-      window.setInterval(this.recheck, 1000);
+      this.intervalID = window.setInterval(this.recheck, 1000);
```
joofsh: Updates to this file are purely for testing purposes

app/views/downloads/_confirm.html.erb
```diff
@@ -1,7 +1,9 @@
-<script>
-  // Set up automatic download status updates
-  window.DownloadStatus.init(<%= @download.id %>, "<%= @download.status %>");
-</script>
+<% if @download.fetching_manifest? %>
```
joofsh: As far as I can tell, this is the only state that we want to poll for download status, right? 

When we're downloading documents we use the `DownloadProgress` js module, and we stop polling via a different mechanism

PR #275, opened 2016-10-24
Updated the text on the download eFolder page.

### LINKED ISSUE #267: Emphasize to users that VVA docs are not included in eX downloads
context: https://github.com/department-of-veterans-affairs/appeals-pm/issues/824

mdbenjam:
Added: "(Remember that Virtual VA documents are not included in this download)" on the download eFolder page.

Fixes #267 

**Test Plan**
-Ran test suite locally.
-Verify the new text is correct: screen shot 2016-10-24 at 4 44 20 pm

### DIFF
--- a/app/views/downloads/_progress.html.erb
+++ b/app/views/downloads/_progress.html.erb
@@ -24,8 +24,8 @@
           <p class="usa-alert-text">
             All of the documents in the VBMS eFolder for #<%= @download.file_number %> are ready to download.
             Click the "Download eFolder" button below.
-            <br/>
-            <br/>
+          </p>
+          <p>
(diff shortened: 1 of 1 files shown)


app/views/downloads/_progress.html.erb
```diff
@@ -24,6 +24,9 @@
           <p class="usa-alert-text">
             All of the documents in the VBMS eFolder for #<%= @download.file_number %> are ready to download.
             Click the "Download eFolder" button below.
+            <br/>
+            <br/>
+            (Remember that Virtual VA documents are not included in this download)
```
joofsh: Can we try wrapping this in a separate <p> instead?

PR #280, opened 2016-10-25
logs user in sentry

amprokop:
untested, need sentry access and the sentry DSN so i can see these coming through

@shanear @orischwartz — are you able to set me up with those?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -22,7 +22,7 @@ class ApplicationController < ActionController::Base
         id: current_user.id,
         email: current_user.email,
         ip_address: current_user.ip_address,
-        station_id: station_id
+        station_id: current_user.station_id
       )
     end
   end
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -14,6 +15,18 @@ def authenticate
...
   end
 
+  def set_raven_user
+    if ENV['SENTRY_DSN']
+      # Raven sends error info to Sentry.
+      Raven.user_context(
+        id: current_user.id,
+        email: current_user.email,
+        ip_address: current_user.ip_address,
+        station_id: station_id
```
orischwartz: Should this be `station_id: current_user.station_id` ? Where is the `station_id` value coming from?

shanear:
tested it. When i fix the bug that Ori pointed out, it works 😎 

this is what shows up in sentry. very cool
screen shot 2016-10-26 at 2 45 33 pm

amprokop:
cool, updated to fix the pointer @shanear

shanear:
@orischwartz this is good to go from my perspective

PR #284, opened 2016-11-10
Fix for ref to now squashed commit

cmgiven:


askldjd:
Thanks for the quick response.

PR #285, opened 2016-11-10
only log sentry user if there's a current_user

orischwartz:
Otherwise causes 

```
NoMethodError (undefined method `id' for nil:NilClass):
  app/controllers/application_controller.rb:22:in `set_raven_user'
```

See https://github.com/department-of-veterans-affairs/caseflow-certification/blob/master/app/controllers/application_controller.rb#L39-L47 for similar implementation of this functionality.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -16,7 +16,7 @@ class ApplicationController < ActionController::Base
   end
 
   def set_raven_user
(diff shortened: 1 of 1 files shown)


orischwartz:
Ironically I think the only action affected by this bug was the health-check itself. All other actions require a logged-in user and `before_action :authenticate` runs before `before_action :set_raven_user`.

PR #286, opened 2016-11-11
Allow Gem source to be overridden by env

askldjd:
Allow gem source to be overridden by `GEM_SERVER_URL`. This URL will port to an internal gem server in deployment scripts.

See also https://github.com/department-of-veterans-affairs/appeals-deployment/pull/198

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,4 @@
-if ENV['GEM_SERVER_URL'] == nil
-  source 'https://rubygems.org'
-else
-  source ENV['GEM_SERVER_URL']
-end
+source ENV['GEM_SERVER_URL'] || 'https://rubygems.org'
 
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -1,4 +1,8 @@
-source 'https://rubygems.org'
+if ENV['GEM_SERVER_URL'] == nil
+  source 'https://rubygems.org'
+else
+  source ENV['GEM_SERVER_URL']
```
joofsh: Super minor, but you want you can condense this to:
 `source ENV['GEM_SERVER_URL'] || 'https://rubygems.org'`

askldjd: Thanks! Complete ruby n00b here.

amprokop:
cool @askldjd.

PR #305, opened 2016-12-02
Adds link to Caseflow Feedback

amprokop:
Links point to Caseflow Feedback in the dropdown and on bottom right if an env variable is set.

To enable this in prod and uat, we'll need to set `ENV["CASEFLOW_FEEDBACK_URL"]`

Testing:
Link works locally, points to demo.
image


amprokop:
Hold on validating this until we enable the Feedback app in eFolder dev. Need to set the env var.

PR #307, opened 2016-12-06
Added a download to search history as soon as the search is started. …

### LINKED ISSUE #298: History Section:  Section not showing up for the first retrival request
While performing a search for a file, the entire History section including the requested veteran name and ID information is not showing up for the first veteran on the eFolder Express landing page.  The History section does show up when I perform a second file retrieval request.   

### Reproduction Steps
1. Enter Vet ID number in the search.
1. Click the Search button.
1. Click on the Start Retrieving eFolder button.
1. While file is being retrieved, click on the eFolder Express logo.  
1. The landing page is displayed.  There is no History section showing on the eFolder landing page nor does the first file being requested show.

Refer to images below in comments

nemodjuric:
…Fixed test for recent downloads.
To test: Do a search and before it finishes click on eFolder Express to get back to the Search page. The search should be present in History section.
fixes #298

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -86,7 +86,7 @@ class DownloadsController < ApplicationController
   end
 
   def recent_downloads
-    @recent_downloads ||= downloads.where(status: [3, 4, 5, 6])
+    @recent_downloads ||= downloads.where(status: [0, 1, 2, 3, 4, 5, 6])
   end
   helper_method :recent_downloads
 
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -415,7 +415,7 @@ RSpec.feature "Downloads" do
 
     visit "/"
 
-    expect(page).to_not have_content(pending_confirmation.file_number)
+    expect(page).to have_content(pending_confirmation.file_number)
     expect(page).to_not have_content(another_user.file_number)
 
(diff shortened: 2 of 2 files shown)


amprokop:
I don't quite understand this fix— would you mind adding a screenshot and some more explanation about what was wrong, @nemodjuric? Thanks!

nemodjuric:
The issue was referenced in the description #298. Screenshots are there as well...
The problem was that once you do a search and go back to the Search page, it wouldn't show that download since it was not completed.

app/controllers/downloads_controller.rb
```diff
@@ -86,7 +86,7 @@ def downloads
   end
 
   def recent_downloads
-    @recent_downloads ||= downloads.where(status: [3, 4, 5, 6])
+    @recent_downloads ||= downloads.where(status: [0, 1, 2, 3, 4, 5, 6])
```
shanear: We didn't include these for a reason. Because the download has not started in these states.

amprokop:
@nemodjuric — sure, I see that there's a description in the issue of the problem from a user-facing perspective. In the PR for a fix like this, I'd like to see an explanation of what was technically incorrect about the code and how your changeset fixes it. That way, whoever is reviewing can get some quick context and evaluate it better.

shanear:
this issue has been closed

Issue #314, opened 2016-12-08
eFolder uses stale user id / station id in BGSService


askldjd:
This is a serious bug. The way BGSService is [caching](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/services/bgs_service.rb#L30) the user object is wrong, and that leads to the app using the wrong user information in the BGS client.

@orischwartz has confirmed that once the user information is cached in the BGS Client, it will stay the same forever.

This has serious security implications because we are not validating the sensitivity level with the right user. See https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/services/bgs_service.rb#L24

askldjd:
As a follow-up to this bug: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/315

PR #316, opened 2016-12-08
Make `BGSService` not a singleton.

shanear:
Fakes::BGSService is still a singleton, need to assess whether
we want to change that in a follow up tech-improvement issue.

#314

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -51,7 +51,7 @@ class ApplicationController < ActionController::Base
   end
 
   def configure_bgs
-    BGSService.user = current_user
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -51,7 +51,7 @@ def strict_transport_security
   end
 
   def configure_bgs
-    BGSService.user = current_user
+    Download.bgs_service = BGSService.new(user: current_user) unless Rails.env.test?
```
joofsh: Doesn't this cause the same issue? Now we're caching a class attr on `Download` instead of within `BGSService`

PR #320, opened 2016-12-13
Log updates

### LINKED ISSUE #190: Log to stdout instead of rotating log files as part of application
> A twelve-factor app never concerns itself with routing or storage of its output stream. It should not attempt to write to or manage logfiles. Instead, each running process writes its event stream, unbuffered, to stdout.

~~To be deployed in conjunction with https://github.com/department-of-veterans-affairs/ansible-common-roles/issues/17~~ Deployed

Should also remove all logic for rolling over log files in Rails. 

Relevant library https://github.com/heroku/rails_stdout_logging (so small it may not be worth adding an additional dependency).

sharonwarner:
This PR sends logs to stdout. 

Fixes #190

### DIFF
--- a/config/initializers/logger.rb
+++ b/config/initializers/logger.rb
@@ -13,6 +13,11 @@ log_tags << lambda { |req|
 # :nocov:
 config = Rails.application.config
 
+# don't mix worker and RAILS http logs
+if !ENV["IS_WORKER"].blank?
+  log_tags << "jobs-worker"
+end
+
 config.log_tags = log_tags
 
 # log sidekiq to application logger (defaults to stdout)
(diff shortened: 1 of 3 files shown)


config/initializers/logger.rb
```diff
@@ -13,27 +13,8 @@ def get_session(req)
 # :nocov:
 config = Rails.application.config
 
-# don't mix worker and RAILS http logs
-if !ENV["IS_WORKER"].blank?
-  config.paths["log"] = "log/efolder-express-worker.log"
-  log_tags << "jobs-worker"
```
orischwartz: doh, I messed up, we still need this line. So something like 

```ruby
if !ENV["IS_WORKER"].blank?
  log_tags << "jobs-worker"
end
```

Issue #301, opened 2016-11-29
Record email addresses of users that downloaded each file


shanear:
As  a site administrator, I want to know the email addresses of the user of each eFolder download

### Acceptance Criteria
- Verify that after initiating a download, the user's email address is recorded on that download for the logged in user
- Verify that you can see the users email address in the "Most Active Users" part of stats page.

### Technical Notes
- We may want a users table. Or we may not, because that's a lot of notes. Part of this story will be weighting the pros and cons of that.

nemodjuric:
@shanear We can add a Users table or we can add it to Searches table (that already contains user_id column). What are your thoughts?

shanear:
I think i'd be easier to just add it to searches. @amprokop what do you think?

aroltsch:
Decided to add `email` to the `searches` table. In the future, we might create `users` table.

aroltsch:
"Most Active Users" tab will include user's email, id and station ID. If email is not recorded, display "No Email Recorded". Example:
`test@example.com (CSFLOW3567 - Station 283)`
`No Email Recorded (CSFLOW7654 - Station 765)`

kierachell:
Shows email when present; No email validation; Empty emails might be saved as blank in the DB so the "No Email Recorded" never shows.

shanear:
@kierachell emails come directly from CSS in UAT, and thus they will not be blank. So we will go ahead and deploy this and test the "No Email Recorded" issue in UAT.

shanear:
Confirmed that "No Email Recorded" is working correctly

PR #324, opened 2016-12-16
Add user's Email to the `searches` table

aroltsch:
1. Add email to the `searches` table
2. Add `database_cleaner` to rspec

connected #301 

Note: I think it would be better to create a follow up issue and create a `users` table.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -16,6 +16,7 @@ class Download < ActiveRecord::Base
   # sort by receipt date; documents with same date ordered as sent by vbms; see
   # https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/213
   has_many :documents, -> { order(received_at: :desc, id: :asc) }
+  has_many :searches
 
   before_create do |download|
     # This fake is used in the test suite, but let's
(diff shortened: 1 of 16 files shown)


app/models/download.rb
```diff
@@ -130,7 +131,13 @@ def complete!
   end
 
   def user_id_string
-    "#{user_id} (Station #{user_station_id})"
+    "(#{user_id} - Station #{user_station_id})"
+  end
+
+  def email
+    Search.where.not(email: nil)
```
shanear: I think you should just be able to use:

```
searches.last.try(:email)
```

app/models/download.rb
```diff
@@ -147,15 +154,17 @@ def number_of_documents
 
   def self.downloads_by_user(downloads:)
     downloads.each_with_object({}) do |download, result|
-      result[download.user_id_string] ||= 0
-      result[download.user_id_string] += 1
+      result[download.user_id] ||= {}
+      result[download.user_id_string][:email] ||= download.email || User::NO_EMAIL
```
aroltsch: Yes. That's why I did what I did on line 138: search through ALL searches to find the email. This way we will call the query only once to find the email, instead of going through every download until we find the email. @shanear

shanear: Ohhh, okay I see now. Looks like I rushed a little to fast through reviewing this PR 😄 

It seems like we're adding a lot of logic in order to avoid adding a users table. I'm good with this for now, but after looking at this PR, it seems like we should really just have a users table. What do you think?

Issue #325, opened 2016-12-19
Create "users" table in eFolder Express


aroltsch:
1. Add `css_id` in `searches` and `downloads` table. Duplicate `user_id` to `css_id`.
2. Create `users` table with the following columns: id, station_id, css_id, email. Migrate data into the `users` table. Populate `user_id` with the id from the `users` table
3. Change code to use `users` table

aroltsch:
Tech Details: 
1st PR should:

add css_id to both searches and downloads
migrate all values in user_id to css_id
write both css_id and user_id in new downloads and searches.
The next PR should:

remove all references to user_id in the code, replacing them with css_id
The 3rd PR should:

remove the user_id column.
The 4th PR should:

Add the users table.
Add user_id column back to both searches and downloads as an integer
Begin writing new users to the users table, and adding the appropriate user_id value to new searches and downloads.
Migrate all css_id and station_id on downloads and searches to the users table, and add the appropriate user_id to each row.
The 5th PR should:

Remove all references to css_id and station_id on the downloads or searches table.
The 6th PR should:

Remove the css_id and station_id column from downloads and searches

kierachell:
The functionality has not changed for `dev`/`demo`; however, (as you've already experienced) `prod` can be a different story.
Needs @shanear 's validation.

shanear:
validated in prod. Nice work @aroltsch !

PR #326, opened 2016-12-20
Add 'users' table, rename user_id to css_id in downloads and searches

aroltsch:
1. Rename `user_id` to `css_id` in `searches` and `downloads`
2. Add `users` table, add `user_id` to `searches` and `downloads`
3. Fix code to use `css_id` instead of `user_id`

Next steps:
1. Remove columns: css_id, email, user_station_id from `searches` and `downloads`
2. Re-factor code to access user data through the `users` table 

connects #325

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -34,6 +34,7 @@ class Search < ActiveRecord::Base
 
   def user=(user)
     self.user_id = user.id
+    self.css_id = user.id
     self.user_station_id = user.station_id
     self.email = user.email
   end
@@ -44,6 +45,7 @@ class Search < ActiveRecord::Base
     Download.active.where(
       file_number: sanitized_file_number,
       user_id: user_id,
+      css_id: css_id,
       user_station_id: user_station_id
     ).where.not(status: [1, 7])
   end
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -17,6 +17,8 @@ class Search < ActiveRecord::Base
     return true if match_existing_download
 
     self.download = download_scope.new
+    # set css_id if it is a new download
+    download.css_id = css_id
 
     return false unless validate!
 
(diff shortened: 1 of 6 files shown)


aroltsch:
@shanear First PR is ready.

app/models/search.rb
```diff
@@ -44,6 +45,7 @@ def download_scope
     Download.active.where(
       file_number: sanitized_file_number,
       user_id: user_id,
+      css_id: css_id,
```
shanear: while the migration is running, there will be some downloads that don't yet have `css_id` set. To be safe, this might be best in PR 2. On the other hand, the migration will probably be pretty fast.

spec/rails_helper.rb
```diff
@@ -29,6 +29,8 @@
 Sniffybara::Driver.configuration_file = File.expand_path("../support/VA-axe-configuration.json", __FILE__)
 Sniffybara::Driver.issue_id_exceptions += []
 
+ActiveRecord::Migration.maintain_test_schema!
```
shanear: ❤️ 

we needed this!

db/migrate/20161220180421_add_css_id_to_downloads_and_searches.rb
```diff
@@ -0,0 +1,10 @@
+class AddCssIdToDownloadsAndSearches < ActiveRecord::Migration
+  def change
+    add_column :searches, :css_id, :string
+    add_column :downloads, :css_id, :string
+
+    (Search.all + Download.all).each do |record|
```
shanear: cool ruby-fu 👍

app/models/search.rb
aroltsch: Yes, good catch! I will set it in PR 2.

aroltsch: Actually, this piece is also used to create a new download if it doesn't exist. If I remove we will run into another problems: downloads with nil css_id @shanear

PR #331, opened 2016-12-21
Remove all refences to user_id in the code replacing them with css_id

aroltsch:
Connects #326

This is PR 2 to replace user_id with css_id everywhere in the code.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -17,8 +17,6 @@ class Search < ActiveRecord::Base
     return true if match_existing_download
 
     self.download = download_scope.new
-    # set css_id if it is a new download
-    download.css_id = css_id
 
     return false unless validate!
 
@@ -35,8 +33,7 @@ class Search < ActiveRecord::Base
   end
 
   def user=(user)
-    self.user_id = user.id
-    self.css_id = user.id
+    self.css_id = user.css_id
     self.user_station_id = user.station_id
     self.email = user.email
   end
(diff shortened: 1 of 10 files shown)


app/models/search.rb
```diff
@@ -35,8 +33,7 @@ def sanitized_file_number
   end
 
   def user=(user)
-    self.user_id = user.id
-    self.css_id = user.id
```
amprokop: hmm.

`self.css_id = user.id`

`self.css_id = user.css_id`

it's likely that only one of these two lines is correct. i'm betting it's the one we deleted here :)

what do you think?

aroltsch: Alex, the code is below on line 36 that sets css_id. @amprokop

amprokop: @aroltsch — my question was that you replaced
`self.css_id = user.id`
with
`self.css_id = user.css_id`

Why the change there?

I can see how this might be correct— maybe we used to (confusingly) store the CSS id in the user model as `user.id` before we persisted the user object, but don't anymore? I'd look to you to explain this change though, since I'm not sure.

aroltsch: Oh okay, I see. In User.rb, I changed `id` to be `css_id` because in the subsequent PRs I will be adding `users` table and `id` will represent a primary key in the users table so I renamed `id` to `css_id` in User. 
@amprokop

amprokop: Ok, sounds good. Thanks for the explanation.

PR #333, opened 2016-12-23
Add users table

aroltsch:
This is the 4th PR:
* Add the users table.
* Add user_id column back to both searches and downloads as an integer
* Begin writing new users to the users table, and adding the appropriate user_id value to new searches and downloads.
* Migrate all css_id and station_id on downloads and searches to the users table, and add the appropriate user_id to each row.

connected #325


amprokop:
i'm off for two weeks, @aroltsch. think you could find someone else to review this? lmk if you can't and i'll take a look, though! don't want you to be blocked.

aroltsch:
@amprokop thanks! I think @nemodjuric  can review it.

PR #334, opened 2016-12-28
Remove all references to css_id and station_id

aroltsch:
This is the 5th PR that removes all references to css_id and station_id on the downloads or searches table.

Removal of the css_id and station_id column from downloads and searches will be in the next PR.

connects #325

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ group :test do
   # to save and open specific page in capybara tests
   gem 'launchy'
   gem 'database_cleaner'
+  gem 'factory_girl_rails'
 end
 
 group :development do
--- a/Gemfile
+++ b/Gemfile
@@ -55,7 +55,7 @@ gem 'rubyzip'
 gem 'zaru'
 
 gem 'connect_vbms', git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "6cde66023ccbd09676b851bdb97f5b8ac383960b"
-gem 'bgs', git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git", :branch => 'ssl'
+gem 'bgs', git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git", :branch => 'master'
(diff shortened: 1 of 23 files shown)


Gemfile
```diff
@@ -95,6 +95,7 @@ group :test do
   # to save and open specific page in capybara tests
   gem 'launchy'
   gem 'database_cleaner'
+  gem 'factory_girl_rails'
```
shanear: we should have a broader discussion about whether we want to use factory girl (in general, we should probably have a broader discussion when introducing big gems like this). I'm afraid that there might be too much of a learning curve for the factory-girl api with not enough benefits. For now could we just create a method that creates a user, then ping #appeals-engineering on tuesday to get thoughts on using factory girl?

in `spec_helper.rb`
```
def create_user(attrs)
  # ...
end

def create_download(attrs)
  # ...
end
```

app/models/download.rb
```diff
@@ -19,6 +19,8 @@ class Download < ActiveRecord::Base
   has_many :searches
   belongs_to :user
 
+  validates :user, presence: :true
```
shanear: i don't think we need this. The user isn't responsible for adding the user here, so our tests should cover the logic.

app/models/search.rb
```diff
@@ -7,6 +7,8 @@ class Search < ActiveRecord::Base
   belongs_to :download
   belongs_to :user
 
+  validates :user, presence: :true
```
shanear: see the comment above

Gemfile
aroltsch: Sounds reasonable!

app/models/download.rb
aroltsch: I personally like to have an additional layer of defense so we won't create bad records in the DB. However, if this is the direction you have been taking, I will remove it.

shanear: My opinion is that we should rely on automated tests as our layer of defense.

This would be a good discussion with the rest of engineering.

PR #344, opened 2017-01-03
Embed doctype, receipt date, and doc ID as PDF metadata

orischwartz:
### New Dependencies 
* Runtime dependency on `pdftk` to facilitate embedding document metadata in PDFs.

### Overview
* Introduce a class attr `pdf_service`  to `DownloadDocuments`.
* Delegate file writing to `pdf_service` when saving documents with `pdf` extension.
* Default `PdfService` implementation invokes `pdftk - update_info`, reading PDF contents from stdin, document metadata from a temporary text file, and writing the result to the specified document path.
* Modify specs to use an actual PDF document contents so pdtk can be invoked.

### Limitations
* No tests were added because affected codepaths already tested via `spec/features/download_spec.rb` and `spec/services/download_documents_spec.rb`.
* Throws an exception if `pdftk` outputs anything to stderr.

**Note:** Cannot be merged until a corresponding change to deployment is made adding the pdftk dependency (https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/344).


.travis.yml
```diff
@@ -19,6 +19,7 @@ before_install:
 before_script:
   - RAILS_ENV=test bundle exec rake db:create
   - RAILS_ENV=test bundle exec rake db:schema:load
+  - sudo apt-get install pdftk
```
shanear: our good friend pdftk :)

app/services/download_documents.rb
```diff
@@ -68,8 +68,19 @@ def download_dir
 
   def save_document_file(document, content, index)
     filename = File.join(download_dir, unique_filename(document, index))
-    File.open(filename, "wb") do |f|
-      f.write(content)
+
+    if document.preferred_extension == "pdf"
+      pdf_attributes = {
```
shanear: this could be a method instead of a variable assignment. Would make the code a little cleaner

app/services/pdf_service.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+
+      # open returns the value of the block
+      file
+    end
+  end
+
+  # see https://github.com/blambeau/pdf-toolkit/blob/master/lib/pdf/toolkit/coercions.rb
+  def self.format_field(field)
+    if field.is_a?(Time)
```
shanear: the terse, "ruby way" 😄
`field.is_a?(Time) ? format_time(field) : field`

app/services/pdf_service.rb
```diff
@@ -0,0 +1,46 @@
...
+      format_time(field)
+    else
+      field
+    end
+  end
+
+  def self.format_time(time)
```
shanear: this seems like it could use a unit test or two

orischwartz:
Thanks @shanear.

 > Do we have PDFTK installing on efolder workers on our ansible scripts
Yes, merged in https://github.com/department-of-veterans-affairs/appeals-deployment/pull/318

 >This is a pretty beefy change that should be tested on a variety of eFolder file data. We should test this pretty extensively on "the efolders from hell". We need to find out those #'s.
Good call...testing 4 cases from `R11.0_DS_VBMS Initial Validation Checklist UAT 4` in UAT environment:
* `/downloads/266`
* `/downloads/267`
* `/downloads/268`
* `/downloads/269`

I'll add the suggested code changes above shortly.

app/services/download_documents.rb
```diff
@@ -68,8 +68,19 @@ def download_dir
...
+        "Document Type" => document.type_name,
+        "Receipt Date" => document.received_at,
+        "Document ID" => document.document_id
+      }
+
+      DownloadDocuments.pdf_service.write(filename, content, pdf_attributes)
+    else
```
joofsh: Not to pile on the test bandwagon but now that this has 2 pathways, can we make sure we add tests for both pdf and non-pdf?

app/services/pdf_service.rb
joofsh: Maybe test this whole service class, if possible

app/services/pdf_service.rb
```diff
@@ -0,0 +1,46 @@
+require "tempfile"
+require "open3"
+
+class PdfService
+  def self.write(filename, contents, pdf_attributes)
+    # write attributes file to tmpfile
+    pdf_attributes_file = write_attributes_file(pdf_attributes)
+
+    # run pdftk, read pdf from stdin, update_info from above file, write to new
+    command = "pdftk - update_info '#{pdf_attributes_file.path}' output '#{filename}'"
```
joofsh: can we make sure we manually test this, that it doesn't change/affect the original pdf's content

app/services/pdf_service.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    _stdout, stderr, _process = Open3.capture3(command, stdin_data: contents, binmode: true)
+
+    fail "cannot write PDF with attributes #{pdf_attributes} to #{filename}: #{stderr}" unless stderr.empty?
+
+    filename
+  end
+
+  def self.write_attributes_file(pdf_attributes)
+    Tempfile.open("pdf-attributes") do |file|
```
joofsh: Just to sanity check, does `Tempfile.open` create a unique filename each time? Otherwise 2 web workers could clash on the same `/tmp/pdf-attributes`

app/services/pdf_service.rb
```diff
@@ -0,0 +1,46 @@
...
+      pdf_attributes.each do |key, value|
+        file.puts "InfoBegin"
```
joofsh: Is there any documentation on the pdf metadata format used here? It might be useful to add a comment with a link that people could reference in the future. I would have expected this needed to be XML

joofsh:
Looks great in general, not much to add beyond what Shane commented already 👍

app/services/pdf_service.rb
orischwartz: Good news -- this is the format for PDF dates but it is only applicable to the created and modified date attributes, which we don't use. 

I'm going to leave date formatting up to the caller, remove these methods entirely, and simplify the format as follows:

```ruby
"Receipt Date" => document.received_at ? document.received_at.iso8601 : ""
```

app/services/pdf_service.rb
orischwartz: Confirmed on the 4 large efolders listed in https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/344#issuecomment-270397289

Also adding a fallback on corrupt PDFs that pdftk cannot handle which writes the bytes without modification.

app/services/pdf_service.rb
orischwartz: [Correct](http://ruby-doc.org/stdlib-2.0.0/libdoc/tempfile/rdoc/Tempfile.html):

> When you create a Tempfile object, it will create a temporary file with a unique filename.

app/services/pdf_service.rb
orischwartz: Unfortunately not, it is pdftk-specific and there's not much documentation on it.

orischwartz:
@joofsh @shanear ready for re-review.

Issue #306, opened 2016-12-06
Create a test endpoint that deletes a cached eFolder


shanear:
In order to write a repeatable external automated test against UAT, we'll need to have an endpoint that deletes an eFolder from the cache, so we can ensure that we are testing the VBMS upload, and not just retrieving a cached file.

This pattern of Data resets is one that will have to be repeated for other automated tests. Would be good to have opinions on how we should tackle this.

shanear:
After speaking with @askldjd, the current plan is to set a test user id in ansible, that user is allowed to delete all data associated with themselves.

kierachell:
Should this just be limited to `dev` and `test` - that way we don't have to hardcode the user.
The access to a specific download is already authorized by session, right?

kierachell:
@ashmahm or someone can check the functionality on `UAT`/`Prod`

kierachell:
Ok; I just checked it against our testing eFolder in UAT AND IT WORKS

PR #347, opened 2017-01-05
Test user download cache delete

kierachell:
This will allow a test user to delete a cached download from the table - this way we can re-retrieve same folders for testing.

connects #306 

I added a route for deleting a download based on a set up shell variable `TEST_USER_ACCOUNT`
Added a link and a controller method (for deleting a specific cached download for the test user)

Since this a dangerous action outside of `dev` I put redundant checks on every step;

I am not totally clear on the cache (if there is one on S3) and how all that works - so let me know what to change for that: does `download_documents` need changes as well?


app/controllers/downloads_controller.rb
```diff
@@ -66,6 +66,14 @@ def record_not_found
     render "not_found", locals: { id: params[:id] }, layout: "application", status: 404
   end
 
+  # test user delete download method
+  def delete
+    if current_user.email == ENV['TEST_USER_ACCOUNT']
+      Download.destroy_all(id: params[:id], user_id: current_user.id)
```
shanear: `downloads.where(id: params[:id]).delete` also does the trick

app/controllers/downloads_controller.rb
```diff
@@ -66,6 +66,14 @@ def record_not_found
     render "not_found", locals: { id: params[:id] }, layout: "application", status: 404
   end
 
+  # test user delete download method
+  def delete
+    if current_user.email == ENV['TEST_USER_ACCOUNT']
```
shanear: it'd be nice to have a test that non-test users cannot do this.

shanear: it's unclear that `TEST_USER_ACCOUNT` needs to be an email address. Maybe call it `TEST_USER_EMAIL`?

app/controllers/downloads_controller.rb
kierachell: 👍

app/controllers/downloads_controller.rb
kierachell: Will have to check in with @askldjd  to see how you guys want it set up in the `ENV`. I do not see `email` as required for all users.

askldjd: This should probably be `current_user.id` (which is CSS ID). We are guaranteed to have the ID field from CSS.

app/controllers/downloads_controller.rb
askldjd: You mentioned S3. I thought we have an aging policy set on the bucket, so it will just clear itself out.

app/controllers/downloads_controller.rb
kierachell: Test added; the route should only be available with the ENV set and visible/accessible only with ENV set to proper value

kierachell:
@shanear 
- modified methods as requested
- added test for availability of the route and functionality of the method
- defeated "rubocop"

.gitignore
```diff
@@ -22,3 +22,5 @@
 *swp
 
 .idea/*
+
+Gemfile.lock
```
kierachell: Ah I was checking `commons` integration on another branch 🤦‍♂️

Issue #299, opened 2016-11-28
Download eFolder:  Error message received when downloading


lakohl:
OGC Contractors reported that they had trouble downloading the zip file for two cases. More than one contractor had the same issue and I was able to replicate it on my machine as well. Will send screenshots over lync.  Suspect it might have something to do with the size of the efolder (large) and causing a time out issue.

### Reproduce Steps
1. Enter Veteran ID.
1. Click on the Search button
1. Wait for the retrieval process to be completed.
1. Click on the View Results link.
1. Click on the Download eFolder button.
1. The following message is displayed:
eFolder Message:  The [Veteran Name].zip download was interrupted. 
Or
Browser Message:  "Source file could not be read."
Or
Browser Message:  "No documents found"

Refer to images below.  You will need to contact @lakohl to get specific IDs.

lakohl:
![lk_reproduces error](https://cloud.githubusercontent.com/assets/4975959/20687902/56d4b734-b58c-11e6-8a34-d1aa096391ea.png)

lakohl:
I have real vet ID #s I can share over Lync for debugging

nemodjuric:
@shanear What are the reproduction steps for local instance?

shanear:
This is only reproducible on prod with certain eFolders, youll need to pair with Alan or me. Save this one for next week.

askldjd:
I looked into this with @nemodjuric, @lakohl, and @joofsh 

@lakohl demonstrated this with a prod case, and the we found out that the zipped file is > 1.4GB. During the download, the browser displays "Unknown File Size" with a spinner, which indicates that the `content-length` header is not set.

At a glance, we see two possibilities

1. Because of the scary looking "Unknown File Size" message, this large download gives the user the feeling of an endless download. User ends the download early, and gets a corrupted zip file.

1. Efolder are prematurely killing the download connection early and lead to a corrupted zip file.

At minimum, I would suggest to add `content-length` [header](http://stackoverflow.com/questions/29485042/how-to-set-content-length-by-send-file) to [send_file](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/downloads_controller.rb#L62) to make sure we display download progress correctly. Afterward, we should re-test to see if it addresses the issue.

shanear:
Added "Content-Type" header. See #350 for validation info.

kierachell:
Can't test large sizes in `demo` but the header does show right now:
[...]

PR #350, opened 2017-01-06
Add Content-Length header to zip download

shanear:
Also, store the filesize to the database as a way of caching it because `File.size` can
take several seconds on files of ~1GB. Additionally allows us to query
the size of downloads.

connects #299

Before:
screen shot 2017-01-06 at 9 32 43 am

After:
screen shot 2017-01-06 at 10 26 17 am

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -62,7 +62,7 @@ class DownloadsController < ApplicationController
     file_exists = @download_documents.zip_exists_locally?
     return record_not_found unless file_exists
 
-    response.headers['Content-Length'] = (download.zipfile_size || File.size(@download_documents.zip_path)).to_s
+    response.headers["Content-Length"] = (download.zipfile_size || File.size(@download_documents.zip_path)).to_s
     send_file @download_documents.zip_path
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/downloads_controller.rb
```diff
@@ -55,11 +55,15 @@ def retry
...
-    @download_documents = DownloadDocuments.new(download: downloads.find(params[:id]))
+    download = downloads.find(params[:id])
+    @download_documents = DownloadDocuments.new(download: download)
     @download_documents.fetch_zip_from_s3
 
     file_exists = @download_documents.zip_exists_locally?
-    file_exists ? send_file(@download_documents.zip_path) : record_not_found
+    return record_not_found unless file_exists
+
+    response.headers["Content-Length"] = (download.zipfile_size || File.size(@download_documents.zip_path)).to_s
```
aroltsch: I think you should be able to set it as an integer and it should be okay.

aroltsch:
Other than my minor comment, looks good!

app/controllers/downloads_controller.rb
shanear: I tried that originally and it didn't work. It causes the "Content-Type" to change to "text/html".

screen shot 2017-01-09 at 9 34 01 am

Issue #346, opened 2017-01-04
Memory consumption is very high when downloading large efolder


askldjd:
When downloading large efolders, memory consumption is unreasonably high.

This issue is seen frequently in Prod. When we reach peak traffic (usually around 6PM EST), efolder crashes.

@orischwartz was able to replicate this in UAT environment with 4 simultaneous large Efolder download. When the EC2 ran out of memory, it was terminated by the ASG.
![image](https://cloud.githubusercontent.com/assets/3258724/21651237/dec3a46a-d275-11e6-9228-dca3376044c1.png)

Here's a relevant conversation with @joofsh 
![image](https://cloud.githubusercontent.com/assets/3258724/21651491/c90a34c6-d276-11e6-8060-65ee63148859.png)

askldjd:
My setup is running efolder locally against production settings. I am downloading a single efolder (efolder from hell). The efolder has ~700 files, so it is a very long running job.

During this download, by default, efolder memory usage will creep up slowly over time. If I take out the restart stalled download cron job in efolder, the memory usage is stable.

It appears that there is a bug in the code where it is mistakenly restarting stalled jobs ... every minute.

![image](https://cloud.githubusercontent.com/assets/3258724/21729474/49b94f00-d419-11e6-9ef2-34ab0acb03a6.png)

askldjd:
Retested in UAT environment with 4 concurrent downloads, and efolder survived. Before this fix, even 2 concurrent downloads would break the server. It should be okay now.

Validation team should double check my work.

PR #352, opened 2017-01-06
Update download updated_at on file download success/fail

joofsh:
connects https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/346


shanear:
& most of all, thanks to you rubocop :robot: 😄

Issue #353, opened 2017-01-09
File Retrival | Download counts is inaccurate on re-download


askldjd:
Steps to reproduce:

1. Download a efolder.
2. After the efolder is downloaded, redownload the efolder. On second download, the remaining download count is inaccurate.

E.g. In this download, 48 files have been completed, yet it shows 26 completed on the status.
<img width="1094" alt="screen shot 2017-01-09 at 3 35 13 pm" src="https://cloud.githubusercontent.com/assets/3258724/21782511/b17d0266-d681-11e6-9c8b-2f2774d472a3.png">

aroltsch:
re: testing, on Demo, you can use "DEMO3" download. It will download documents with errors.

kierachell:
As far as `DEMO3` folder goes the retrieved amounts are the same upon repeated attempts (20 total, 15 retrieved, 5 errors):
<img width="1017" alt="screen shot 2017-01-31 at 08 18 28" src="https://cloud.githubusercontent.com/assets/18075411/22466425/30c1e2d6-e78e-11e6-9c81-99ec3b855c01.png">

If downloaded, the resulting .zip file has 16 (not 15) documents:
<img width="228" alt="screen shot 2017-01-31 at 08 19 25" src="https://cloud.githubusercontent.com/assets/18075411/22466444/482b22ca-e78e-11e6-90ba-059d8a48869f.png">

Subsequent retrieval will update the amount of files to 16 out of 20:
<img width="1023" alt="screen shot 2017-01-31 at 08 21 08" src="https://cloud.githubusercontent.com/assets/18075411/22466466/580f887a-e78e-11e6-9a1b-3aaa7fdc9043.png">

Seems like a bug.

kierachell:
Now with `DEMO2` folder (100 documents):
<img width="996" alt="screen shot 2017-01-31 at 08 28 08" src="https://cloud.githubusercontent.com/assets/18075411/22466694/7233f28a-e78f-11e6-8be6-52d997e826dc.png">
<img width="466" alt="screen shot 2017-01-31 at 08 28 28" src="https://cloud.githubusercontent.com/assets/18075411/22466695/7233f0f0-e78f-11e6-9e46-e9d8d4d4daf7.png">
Redownloading the folder produces the same results 100 documents.

kierachell:
Now for `DEMO4`:
[...]

PR #368, opened 2017-01-27
Progress bar: fix the progress on the second and later download attempts

aroltsch:
For a progress bar "X of XXX files remaining", use `download_status` to count in-progress documents instead of `completed_at` timestamp. Using `completed_at` would work on the first attempt to download the documents but it would not work on the second attempt because documents with `success` status were marked as completed on the first attempt. 

image

connects #353

### DIFF
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -277,6 +277,36 @@ RSpec.feature "Downloads" do
     expect(page).to_not have_css ".document-pending", text: "yawn.pdf"
   end
 
+  scenario "Progress bar shows correct information" do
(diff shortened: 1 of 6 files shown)


spec/features/download_spec.rb
```diff
@@ -277,6 +277,36 @@
     expect(page).to_not have_css ".document-pending", text: "yawn.pdf"
   end
 
+  scenario "Progress bar shows correct information" do
```
shanear: Did this test fail before the above code changes? I'm having a hard time seeing where the bug was.

aroltsch:
@shanear I made changes based on our discussion.

PR #412, opened 2017-03-12
Sandra/efolder changes

Sjones352:
**Review**
Please ignore PR videos #411 as I have updated and fixed the failed tests.
https://github.com/department-of-veterans-affairs/appeals-design-research/issues/103

After cleaning my main.scss files for the failed tests, I am still having issues with this below. When I run rake, that's the only error I get.
screen shot 2017-03-13 at 1 53 59 pm


Sjones352:
I need to get rid of this PR since all my tests are now passing.

screen shot 2017-03-13 at 2 40 57 pm

test
```diff
@@ -0,0 +1 @@
+plain text
```
shanear: this file seems like it shouldn't have been added.

app/views/help/show.html.erb
```diff
@@ -136,8 +186,8 @@
   <h2 id="need-help">18. What if I still need help?</h2>
 
   <p>
-    If you still encounter issues or have questions that aren't answered here, reach out to the National Service Desk by calling
-    1-855-NSD-HELP (1-855-673-4357).
+     If you still encounter issues or have questions that aren't answered here, reach out to the Caseflow Product Support Team by calling 1-844-876-5548.
   </p>
 
+<div class="scrollToTop"></div>
```
shanear: This link is not accessible in many ways:

1. We should avoid using images to display the text inside of the link. And the background too ideally.
2. It is a `<div>` and not an `<a>`. The browser will not read this div as being clickable.

Additionally, we shouldn't need all that fancy Javascript for this one link. We can use the built in "target location" for this.

see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a
> A URL fragment is a name preceded by a hash mark (#), which specifies an internal target location (an ID of an HTML element) within the current document.

After all of that, I think this should look more like:
`<a href="#top">Back</a>`

There are other changes required that I will add in other comments.

app/assets/stylesheets/_main.scss
```diff
@@ -171,3 +171,40 @@ table td {
...
+  margin-top: 12px;
+}
+
+.Title {
+  border-bottom: 1px solid;
+  margin-bottom: 3rem;
+  border-bottom-color: $color-gray-light;
+}
+
+.scrollToTop {
```
shanear: we should be using the USWDS to style this button. See: https://standards.usa.gov/components/buttons/

app/assets/stylesheets/_main.scss
```diff
@@ -171,3 +171,40 @@ table td {
...
+}
+
+.text-area {
+  width: 265px;
+  float: left;
+  line-height: 1.25;
+  margin-top: 12px;
+}
+
+.Title {
```
shanear: This should be styled by the US Web Design standards:
https://standards.usa.gov/components/typography/#typesetting

app/assets/stylesheets/_main.scss
```diff
@@ -171,3 +171,40 @@ table td {
 .ee-documents-table {
   margin-top: 0;
 }
+
```
shanear: Most of the CSS in this document is unnecessary. We should be leaning on the US Web Design Standards for how text is styled and placed:
https://standards.usa.gov/components/typography/

app/assets/javascripts/application/scroll.js
```diff
@@ -0,0 +1,18 @@
+$(document).ready(function(){
```
shanear: We shouldn't need all this fancy JS here. See my comment below.

app/assets/stylesheets/_main.scss
Sjones352: When I tested it, this is the error I was getting so I had to use the existing variable.
screen shot 2017-03-14 at 11 55 48 am
This is actually a grey horizontal divider between the different sections.USWDS to style
screen shot 2017-03-14 at 11 58 13 am

test
Sjones352: Deleted the file

app/assets/stylesheets/_main.scss
Sjones352: The design team had requested for the back to top button

app/assets/stylesheets/_main.scss
Sjones352: This is actually horizontal-line that divides the sections. I called the name of it to indicate horizontal line. There is no  horizontal line style in the US Web Design Standards

app/assets/javascripts/application/scroll.js
Sjones352: Deleted Javascript

Sjones352:
@shanear Please review PR
screen shot 2017-03-17 at 4 13 10 pm

app/assets/stylesheets/_main.scss
```diff
@@ -171,3 +171,12 @@ table td {
...
+.training-videos {
```
shanear: Now that you have the columns, what happens if you delete this? Do the videos automatically become the right size.

app/assets/javascripts/application.js
```diff
@@ -17,4 +17,5 @@
 $(function() {
   window.Dropdown.bind();
   window.LoadingIndicator.bind();
+  window.scroll.bind();
```
shanear: This can be deleted.

app/assets/stylesheets/_main.scss
Sjones352: @shanear 
when I took out the class "training-videos" they don't have enough space in between.
screen shot 2017-03-21 at 1 03 57 pm
## Screenshot2
So I set the heights and widths of the videos in there iFrames.
screen shot 2017-03-21 at 1 23 58 pm

app/assets/javascripts/application.js
Sjones352: Deleted it

PR #417, opened 2017-03-14
Reference the document types from commons

mdbenjam:
Use the document types from commons and remove the copy in efolder.

**Test Plan**
- [x] Smoke test, to make sure document type named files still show up correctly.


mdbenjam:
For sure, I already checked it!

Issue #415, opened 2017-03-14
Remove DocumentTypes class and use `type_description` from VBMS eFolder Version 1.0 Service


aroltsch:
The new VBMS service returns type_description so we no longer need `DocumentTypes` map. 
1. Update the existing DocumentTypes map with the latest document types.
2. Create a migration to set `type_description` on already existing records. 
3. Remove `doc_type` column from `documents` and use `type_id` when appropriate.
4. Remove DocumentTypes class.

PR #418, opened 2017-03-15
Use VBMS eFolder service v1

aroltsch:
1. Connect to VBMS eFolder service v1 if `USE_VBMS_V5` is enabled
2. Add `type_description` column to `documents`
3. Make the code work for both versions. 
4. Create subsequent tickets that will be implemented when we switch to a new service completely.
[#416](https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/416)
[#415](https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/415)

connects #418

### DIFF
--- a/app/services/vbms_service.rb
+++ b/app/services/vbms_service.rb
@@ -5,7 +5,7 @@ class VBMSService
   def self.fetch_documents_for(download)
     @client ||= init_client
 
-    request = if ENV["USE_VBMS_V5"].to_b
+    request = if FeatureToggle.enabled?(:vbms_efolder_service_v1)
                 VBMS::Requests::FindDocumentSeriesReference.new(download.file_number)
               else
                 VBMS::Requests::ListDocuments.new(download.file_number)
@@ -19,7 +19,7 @@ class VBMSService
   def self.fetch_document_file(document)
     @client ||= init_client
 
-    request = if ENV["USE_VBMS_V5"].to_b
+    request = if FeatureToggle.enabled?(:vbms_efolder_service_v1)
                 VBMS::Requests::GetDocumentContent.new(document.document_id)
               else
                 VBMS::Requests::FetchDocumentById.new(document.document_id)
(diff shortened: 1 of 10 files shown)


app/services/vbms_service.rb
```diff
@@ -5,7 +5,11 @@ class VBMSService
   def self.fetch_documents_for(download)
     @client ||= init_client
 
-    request = VBMS::Requests::ListDocuments.new(download.file_number)
+    request = if ENV["USE_VBMS_V5"].to_b
```
shanear: ENV vars aren't the best here because we have 4 instances in prod and changing them is a pain.

Can we port your beautiful feature flag stuff to commons and use it here?

spec/models/document_spec.rb
```diff
@@ -17,10 +17,10 @@
     subject { document.s3_filename }
 
     let(:document) do
-      Document.new(vbms_filename: "keep-stamping.pdf", mime_type: "application/pdf", download_id: 45)
+      Document.new(id: 5, mime_type: "application/pdf", download_id: 45)
     end
 
-    it { is_expected.to eq("45-#{document.id}-keep-stamping.pdf") }
+    it { is_expected.to eq("#{document.download_id}-#{document.id}.pdf") }
```
shanear: We want to be hard coding as much of the expected value as possible in our tests. It make sure that we know what we are testing. Thoughts?

app/services/vbms_service.rb
aroltsch: Haha, yes, I need to do that.

aroltsch:
@shanear I don't think we should be removing Document#type_name with this PR (it should be done in the subsequent ticket https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/415). We need a migration to update type_description for already existing records.

spec/models/document_spec.rb
aroltsch: Ok, make sense.

aroltsch:
Yes, I was waiting on the Commons PR to be approved :)

Issue #49, opened 2016-05-24
Make error messages more noticible


KHarshawat:
During usability testing, most users didn't notice when a download had errors right away. We should explore how to make these errors more obvious. This may be solved if we do a good job with #46 which will allow the users to see the error more easily. 

Otherwise we could make the error a popup. Doing so would be tricky if the user revisits the status page from the homepage.

I'll take a pass at the design.

shanear:
https://marvelapp.com/h141bc#13388413

shanear:
@KHarshawat not sure if there's any dev work here not captured by #51. Let me know if so.

KHarshawat:
@shanear I think we're able to accomplish this mostly by making the list results in a scroll window so the user can always see the top of the page and the error messages. I would also like to show the following pop up if a user tries to click the download button if there were errors with fetching the documents from VBMS. 

![image](https://cloud.githubusercontent.com/assets/6335546/16717646/bd3fcd2a-46e6-11e6-85d6-7540f39e8716.png)


When you're actually implementing this let me know. I'd like to make sure it's consistent with the other modal we have in Certification but also clean up the design. I'm a little unsure of the cancel placement and phasing.

PR #430, opened 2017-03-22
copy from caseflow

kierachell:
this will collect the latest tag and pass it to the common pipeline for prod deployments
connects [QA #49](https://github.com/department-of-veterans-affairs/appeals-qa/issues/49)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,7 +9,7 @@ def APP_NAME = 'efolder';
 
 // The application version to checkout.
 // See http://docs.ansible.com/ansible/git_module.html version field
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -9,7 +9,7 @@ def APP_NAME = 'efolder';
 
 // The application version to checkout.
 // See http://docs.ansible.com/ansible/git_module.html version field
-def APP_VERSION = 'HEAD'
+def APP_VERSION = 'HEAD';
```
askldjd: haha, i don't think groovy needs semicolons...

Issue #431, opened 2017-03-23
Rename doc_type to type_id in documents table


aroltsch:

aroltsch:
PostMortem: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/443

kierachell:
**PASSED**

```
#<Document id: 1, download_id: nil, download_status: 0, document_id: nil, vbms_filename: "", filepath: nil, source: nil, mime_type: nil, received_at: nil, created_at: "2017-04-06 18:52:19", updated_at: "2017-04-06 18:52:19", started_at: nil, completed_at: nil, lock_version: 0, type_description: nil, type_id: nil, error_message: nil>
```

PR #441, opened 2017-03-28
Remove read/write from doc_type

aroltsch:
This is PR 4

connects #431


aroltsch:
Waiting on the migration to deploy.

aroltsch:
@shanear this is no longer blocked.

Issue #427, opened 2017-03-20
Filter out all "Fiduciary" documents from eX downloads


amprokop:
As a eFolder Express user, I never should be able to see or download "Fiduciary" documents.

## Acceptance Criteria
- Validate that no documents of type Fiduciary are being shown in the documents list. Fiduciary documents are defined as having type id's in the following set:

```
552, 600, 607, 601, 602, 546, 603, 604, 545, 605, 606, 608, 609, 575, 543, 452, 547, 610, 611, 445, 574, 458, 535, 612, 614, 442, 595, 644, 615, 616, 541, 540, 456, 403, 617, 618, 620, 619, 715, 621, 622, 623, 624, 716, 625, 626, 628, 629, 630, 631, 443, 632, 633, 634, 635, 636, 439, 440, 438, 441, 551, 550, 637, 638, 639, 596, 640, 642, 643, 511, 402, 538, 645, 646, 647, 648, 544, 649, 650, 536, 539, 537, 576, 457, 455, 421, 422, 424, 594, 425, 426, 169, 404, 454, 128, 429, 430, 431, 432, 433, 434, 435, 436, 453, 437, 657, 651, 542
```

## Notes
- Once this is complete, we should validate that we aren't getting any more sensitive document errors in prod (this issue: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/395)

## Background
Currently, we fetch all documents from VBMS. We've received guidance from the chief of the fiduciary service:

> Re: Fiduciary: It appears that, per VA policy, the documents are not to be released. As such, we need to come up with a way to turn off the spigot for documents that fall into the fiduciary category.

In the `ListDocuments` call, we should filter out all documents with type "Fiduciary".

See, https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/410
Also, https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/304

shanear:
This can come with @aroltsch's feature flag on eFolder Service v1

kierachell:
I am requesting VBMS team to provide us with with some file #s to fully validate this.

aroltsch:
@kierachell I can give you two cases that are in production.

kierachell:
**PASSED**
I uploaded Doctype 552 to File # 561349912.
[...]

PR #442, opened 2017-03-29
Filter out fiduciary document types

aroltsch:
Filter out fiduciary document types

connects #427


amprokop:
Super awesome, thanks @aroltsch!!!

Issue #434, opened 2017-03-23
Show history with the most recent download at the top


shanear:
Currently eX History shows the files that have been searched for in order from oldest (at top of the list) to newest (at bottom of the list). It should be the opposite.

## Acceptance Criteria
- Verify that the most recently searched eFolder shows at the top of the History and the oldest search shows at the bottom of the list, and that the list is in reverse chronological order.

## Mocks
When this issue is implemented DEMO987 would represent the most recent search and DEMO123 represents the oldest search.
<img width="939" alt="screen shot 2017-03-23 at 12 22 57 pm" src="https://cloud.githubusercontent.com/assets/4975959/24258239/84cb8ba4-0fc3-11e7-80e2-f272bc42137e.png">

astewarttistatech:
**PASSED**

Validated in: UAT
Browsers Validated: FF and Chrome

Steps to Validate: 
1. Go to Efolder on UAT 
2. Retrieve about 4+ eFolders for applicable veterans. 
3. View the History and see that the retrievals are in most recent to oldest order.

Screenshots: 
<img width="1011" alt="screen shot 2017-04-04 at 8 48 27 am" src="https://cloud.githubusercontent.com/assets/23038505/24657407/17ecc552-1914-11e7-95f3-afddff895b77.png">

PR #445, opened 2017-03-30
Artem/reverse-downloads-and-logging

kierachell:
connects #434 

Also added VBMS request logging because we need it to debug some silent errors

### DIFF
--- a/app/services/vbms_service.rb
+++ b/app/services/vbms_service.rb
@@ -45,7 +45,7 @@ class VBMSService
     return VBMS::Client.from_env_vars(
       logger: RailsVBMSLogger.new,
       env_name: ENV["CONNECT_VBMS_ENV"]
-    ) if Rails.application.secrets.vbms["env"]
+    ) #if Rails.application.secrets.vbms["env"]
 
     VBMS::Client.new(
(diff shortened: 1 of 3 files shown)


kierachell:
Ensured that downloads are sorted by _created_at_ in reverse order
screen shot 2017-03-30 at 7 27 15 am

app/services/vbms_service.rb
```diff
@@ -68,6 +68,11 @@ def log(event, data)
             "VBMS HTTP Error #{data[:response_code]} " \
             "(#{data[:request].class.name}) #{data[:response_body]}"
           )
+        else
+          Rails.logger.info(
+            "VBMS Request Sent: #{data[:request].class.name}" \
+            "VBMS Reponse Code #{data[:response_code]}"
+          )
         end
```
nemodjuric: The format we are currently logging is response_code followed by request details. Do we want to keep that format for 200 response code as well?

kierachell:
@nemodjuric 
screen shot 2017-03-30 at 8 03 06 pm

and error:
screen shot 2017-03-30 at 8 01 47 pm

Whadayathink?

Issue #395, opened 2017-02-17
Solve with sensitive document errors in production


shanear:
So some of the document errors we've been getting are due to them being sensitive documents in VBMS.

Right now, we just call them errors (and don't tell them the cause). There might be a more user friendly way to deal with these, and also don't record them as errors in our stats page.

See original conversation here: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/330

aroltsch:
Looking at February 2017 because this month had the most document errors. There are 499 document errors for the month of February, however, there are only 20 actual downloads that are associated with these failed documents:
```
Document.where(download_status: 2, created_at: start_time..end_time).map(&:download_id).uniq.size
=> 20
``` 

Next step: Looking at the status for these downloads.

aroltsch:
#### Download id: 8126 ####
Failed with 1 document error due to the following: message="CMSERR0023 - The document ID <document_id> does not exist."
#### Download id: 8150 ####
No failure 
#### Download id: 9117 ####
Failed with 1 document error. Checked VBMS production and this document fails to preview there as well: Failed to load PDF document.
#### Download id: 9390 ####
Failed with 7 documents. Checked VBMS production and these documents fails to preview there as well: Failed to load PDF document.
#### Download id: 9417 ####
Failed with 1 document. Checked VBMS production and these documents fails to preview there as well: Failed to load PDF document.
#### Download id: 9646 #### 
No failure
#### Download id: 9997 ####
no failure 
#### Download id: 10016 ####
no failure 
#### Download id: 10017 #### 
no failure
#### Download id: 10018 #### 
no failure 
#### Download id:  10019 ####
no failure 
#### Download id:  9996 ####
no failure 
#### Download id:  10020 #### 
no failure 
#### Download id:  10222 ####
References the same file number as #4
#### Download id: 10308 ####
Failed with 1 document error. Checked VBMS production and this document fails to preview there as well: Failed to load PDF document.
#### Download id:  10551 ####
#### Download id:  10573 #### 
References the same file number as #16
#### Download id:  10600 ####
#### Download id:  10681 ####
Failed with 1 document error. Checked VBMS production and this document fails to preview there as well: Failed to load PDF document.
#### Download id:  11571 ####
No failure

aroltsch:
Solution: save document errors in the DB so we can easily track these.

aroltsch:
Some document errors are caused by:
```
[...]

PR #455, opened 2017-04-07
Anya/395 capture document errors

aroltsch:
1. Keep track of document errors in the documents table. 
2. Better handling of empty documents will lead to much less document errors. I created a ticket for Lara to determine what to do with empty documents. 

connects #395

### DIFF
--- a/app/services/vbms_service.rb
+++ b/app/services/vbms_service.rb
@@ -13,7 +13,7 @@ class VBMSService
     @client.send_request(request)
   rescue => e
     Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
-    raise VBMS::ClientError
+    raise VBMS::ClientError, e
   end
 
   def self.fetch_document_file(document)
@@ -28,7 +28,7 @@ class VBMSService
     result && result.content
   rescue => e
     Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
-    raise VBMS::ClientError
+    raise VBMS::ClientError, e
   end
 
   def self.vbms_config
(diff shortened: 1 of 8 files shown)


app/services/vbms_service.rb
```diff
@@ -13,7 +13,7 @@ def self.fetch_documents_for(download)
     @client.send_request(request)
   rescue => e
     Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
-    raise VBMS::ClientError
+    raise VBMS::ClientError, e
```
shanear: shouldn't this just be `raise e`?
What does passing two args to `raise` do?

spec/services/download_documents_spec.rb
```diff
@@ -118,10 +118,12 @@
         expect(successful_document.filepath).to eq((Rails.root + "tmp/files/#{download.id}/00010-VA 9 Appeal to Board of Appeals-20150101-1.pdf").to_s)
         expect(successful_document.started_at).to eq(Time.zone.now)
         expect(successful_document.completed_at).to eq(Time.zone.now)
+        expect(successful_document.error_message).to eq nil
```
shanear: Should we check to see that the message matches some text?
```
expect(successful_document.error_message).to match(/Failure/)
```

app/services/vbms_service.rb
aroltsch: The first arg is the exception type, the second argument is the message. If you don't specify the type, it will default to whatever the type it received (for example, `VBMS::HTTPError`). Then we are specific about what we are rescuing in get_download_manifest_job.rb and we rescue `VBMS::ClientError`. @shanear

spec/services/download_documents_spec.rb
aroltsch: The change is on line 126.

app/services/vbms_service.rb
shanear: 👍 thanks @aroltsch

Issue #404, opened 2017-03-02
eFolder Express is returning .bin files


shanear:
From Nick Holtz (i don't know his github tag). He's seeing that efolder express is downloading .bin files in some of the eFolders.

Get the file numbers where he is experiencing this issue, and figure out why this is happening, and if there's anything we can do to fix it.

# Notes
- Idea:  put monitors in the app.

nicholasholtz:
I have file number for whoever would like it.

nicholasholtz:
![image](https://cloud.githubusercontent.com/assets/16781398/23524080/30521d3a-ff57-11e6-8094-cd1470de25af.png)
@shanear

aroltsch:
Some documents are returned with mime_type: application/octet-stream which in turn becomes a document with a `.bin` extension . In VBMS, these documents are displayed without error. I suggest that we add a check for `application/octet-stream` mime_type and use `.pdf` as a preferred extension instead of `.bin`. Thoughts? 
@amprokop @shanear

amprokop:
@aroltsch — let's try to get an answer from VBMS first before we take more action. link up with @kierachell on this and add their response here?

aroltsch:
Waiting on the response from the VBMS team.

kierachell:
I sent them an email in the context of testing the new API:

```
- We were testing FindDocumentSeriesReference and found a few documents return with mime type "application/octet-stream" and file extension `.bin`. It seems like these documents can be opened with Adobe if the file extension is changed to `.pdf`. Is it expected that some of the documents in eFolder will be of mimetype "application/octet-stream" and safe to convert to PDF?
```

Meanwhile:

Here are some cases that are returning errors:
- 332345345   

`TypeError: no implicit conversion of Symbol into Integer` in xml_helper


- 112938488
[...]

PR #460, opened 2017-04-14
Anya/404 bin files

aroltsch:
connects #404

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -11,6 +11,14 @@ class Document < ActiveRecord::Base
 
   after_initialize { |document| document.vbms_filename ||= "" }
 
+  # It is expected that some of the documents may have a MIME type of "application/octet-stream".
+  # However, there is not a guarantee that all documents of this type can be opened as PDFs.
+  # The "application/octet-stream" MIME type could represent arbitrary data, mislabeled PDFs, etc.
+  # and there is no guarantee on file format without investigating the individual document.
+  # Convert mime type to "application/pdf" if it is a binary file
+  # then PDFkit will check if PDF is valid
+  after_initialize :adjust_mime_type
+
   def self.ordered_by_completed_at
     where.not(completed_at: nil).order(completed_at: :desc)
   end
@@ -93,5 +101,11 @@ class Document < ActiveRecord::Base
 
       (total_time / count).round(2)
     end
+  end
+
+  private
+
+  def adjust_mime_type
+    self.mime_type = "application/pdf" if self.mime_type == "application/octet-stream"
   end
 end
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -106,6 +106,6 @@ class Document < ActiveRecord::Base
   private
 
   def adjust_mime_type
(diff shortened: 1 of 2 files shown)


app/models/document.rb
```diff
@@ -11,6 +11,14 @@ class Document < ActiveRecord::Base
 
   after_initialize { |document| document.vbms_filename ||= "" }
 
+  # It is expected that some of the documents may have a MIME type of "application/octet-stream".
+  # However, there is not a guarantee that all documents of this type can be opened as PDFs.
+  # The "application/octet-stream" MIME type could represent arbitrary data, mislabeled PDFs, etc.
+  # and there is no guarantee on file format without investigating the individual document.
+  # Convert mime type to "application/pdf" if it is a binary file
+  # then PDFkit will check if PDF is valid
+  before_create { |document| document.mime_type = "application/pdf" if document.mime_type == "application/octet-stream" }
```
shanear: not sure we should wait for the Document to be saved to perform this conversion. How do you feel about making this a `after_initialize`.

Also, could we move it out to a private method like `adjust_mime_type`?

spec/models/document_spec.rb
```diff
@@ -13,6 +13,22 @@
     end
   end
 
+  context ".create" do
+    subject { Document.create(mime_type: mime_type).mime_type }
```
shanear: Database saves really slow down the unit tests, so it's nice to be able to avoid them when we can. Probably not a big deal here, but it adds up.

app/models/document.rb
```diff
@@ -11,6 +11,14 @@ class Document < ActiveRecord::Base
 
   after_initialize { |document| document.vbms_filename ||= "" }
 
+  # It is expected that some of the documents may have a MIME type of "application/octet-stream".
```
shanear: Amazingly helpful explaination 😄 💯

aroltsch:
I made the changes, please take a look @shanear

app/models/document.rb
```diff
@@ -94,4 +102,10 @@ def calculate_average_download_rate(documents)
       (total_time / count).round(2)
     end
   end
+
+  private
+
+  def adjust_mime_type
+    update(mime_type: "application/pdf") if mime_type == "application/octet-stream"
```
shanear: doesn't `update` save the record to the DB? I don't think we want that here.

aroltsch: Oh, we don't want to save it as an application/pdf mimetype in the DB? @shanear

shanear:
Sorry about the oversight, I do still think this is the right way forward in the long term, it just has some adverse affects with in progress downloads 😢 

I'm thinking we should revert https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/462 in about a week.

What do you think @aroltsch?

aroltsch:
No worries @shanear! 

Why do you think before_create is not a good way forward?

aroltsch:
We could definitely revert #462  though because by that time we should not have any documents with .bin extensions as they will all be deleted in a week.

shanear:
@aroltsch attaching functionality to the DB saves can cause issues and make things harder to test. Better to do in ruby I think.

Issue #475, opened 2017-04-27
Fetching large files from S3 may lead to download timeout


askldjd:
Efolder operates in the following matter when you press the **Download File** button

1. Checks if the bundled zip file is on disk. If so, go to step 3. If not, proceed to step 2.
2. Download the zip file from S3.
3. Call `send_file` with the file file path.

If the file is really large, step 2 may take considerable amount of time, and may exceed the HTTP timeout. So in user's perspective, they will see a 504 error.

Based on quick benchmark, EC2 can fetch from S3 at ~23MB/s. So in the case of https://github.com/department-of-veterans-affairs/appeals-support/issues/133, it will take 104s to fetch the 2.4GB file from S3. This easily exceed the HTTP timeout of 60s.

Possible Solutions:
1. ~~Modify Nginx to allow for larger timeout and hide the problem.~~
1. Allow Rails to stream the file from S3 (not sure how to stream data in Rails)
1. Nginx proxying S3 (how do we authenticate?)
1. AWS Presign URL (might have ATO issue)
1. Use a small Node.js streaming server with Rails

askldjd:
Attempted to resolve this bug by changing Nginx, and this path led to a lot of issues.
 - By increasing nginx timeout, we moved the timeout to ELB.
 - By increasing ELB idle timeout, we moved to the timeout to browser.

In the end, the timeout happened on the browser side. Therefore, solution #1 is a no-go.

askldjd:
After fiddling with https://github.com/vala/s3streamer, I was unable to find a working solution in Rails 4.

However, I have another solution working:

A much simpler approach is to have Rails generate a presigned S3 URL that expires in a short period of time (say 60s). After a quick prototype, I was able to get this to work with the following changes.

```ruby
# in caseflow-commons s3_service.rb
def self.get_presigned_url(filename)
  Aws::S3::Presigner.new.presigned_url(
    :get_object,
    bucket: bucket_name,
    key: filename,
    expires_in: 60
  )
end
```

```ruby
# efolder / download_documents.rb
def fetch_presigned_url_from_s3
  @s3.get_presigned_url(@download.s3_filename)
end

# efolder / downloads_controller.rb
def download
...
  redirect_to @download_documents.fetch_presigned_url_from_s3
end
```

I have discussed this with @shanear before and the ATO issue has come up. However, after giving it a good thought, I am not sure how this is any less secure than our current approach.

askldjd:
[...]

PR #486, opened 2017-05-17
Stream zip file from s3 to the client

aroltsch:
connects #475 

We need to test it on UAT.

Note: I will update Commons  once the commons PR is merged.


app/services/download_documents.rb
```diff
@@ -115,11 +115,9 @@ def zip_exists_locally?
     File.exist?(zip_path)
   end
 
-  def fetch_zip_from_s3
-    # if the file exists on the filesystem, skip
-    return if zip_exists_locally?
-
-    @s3.fetch_file(@download.s3_filename, zip_path)
+  def stream_zip_from_s3
+    key = Rails.application.config.s3_enabled ? @download.s3_filename : zip_path
```
shanear: this could be moved to a function `def streaming_s3_key`

aroltsch: good idea

aroltsch:
We will be testing it using this branch so I will not merge it yet.

askldjd:
Tested pretty extensively for an hour. 

Verified:
 - Download streams immediately
 - Zero files are in tmp folder
 - Little to zero CPU usage

Side note: I also found out that VA network routers is probably buffering our transfer. I started 4 500MB download, killed all EC2 within 1 min of download, and all files were transferred successfully. Our ELB connection draining parameters is probably not doing much because of this.

Issue #488, opened 2017-05-22
eX Display VVA docs in manifest


lakohl:
Now that eX is supporting VVA we need to show the VVA documents to the user and to distinguish between docs fetched from VVA and docs fetched from VBMS.

## AC

1. Verify that all docs display in date order with the most recent date at the top.

1. Verify that VVA and VBMS docs are show interspersed 

1. Verify that a new column with the header 'Source' is added to the table before the doc date column to indicate the database from which the doc was retrieved.

1. Verify that when necessary, the document name (type) is wrapped to the next line

## Mocks
<img width="636" alt="screen shot 2017-05-22 at 1 16 34 pm" src="https://cloud.githubusercontent.com/assets/4975959/26320451/f0e3025c-3ef0-11e7-8cfd-ce5650a13798.png">

<img width="530" alt="screen shot 2017-05-22 at 1 25 49 pm" src="https://cloud.githubusercontent.com/assets/4975959/26320797/36565b80-3ef2-11e7-97a2-3992aa225b00.png">

kierachell:
#### Verify that when necessary, the document name (type) is abridged with ellipses in the table to accommodate the new 'Source' column.

Currently it just wraps to 2nd line: @lakohl is that a suitable solution?

![newsource](https://user-images.githubusercontent.com/18075411/27254238-bb1dce38-5352-11e7-9ebd-93c6985dcf50.gif)

aroltsch:
@lakohl , could you please confirm that it is okay.

lakohl:
@aroltsch yes, let's try it with the document type wrapping to a second line

kierachell:
**PASSED**
Then we good!

PR #494, opened 2017-06-13
Implement FE changes for VVA

aroltsch:
This is an FE piece for VVA. VVA is hidden behind the Feature Flag `vva_service` in UAT and Prod. Per agreement with Lara, `source` column will not be hidden behind the feature flag so it will be visible once this is merged. 

connects #488 

Example when the VVA flag is enabled:
image


app/jobs/demo_get_download_files_job.rb
```diff
@@ -1,9 +1,10 @@
-class FakeVBMSService
+class FakeService
```
aroltsch: good call!

Issue #375, opened 2017-02-03
Error message sometimes showing up when searching for an eFolder


kierachell:
# Description
After clicking on the Search button from the eFolder Dashboard, an error message is received.  
However, if I go back to the Dashboard history and then to _View Progress_ for that eFolder, it shows that documents were, in fact, listed.

This might be a network issue of some sort; but I was able to isolate it to the original query sent by the `Search` button. For whatever reason, `listDocument` or another command return errors despite the query actually going through.

**Environment:**
- [x] Production
- [x] UAT
- [ ] Demo
- [ ] Development

**Type of Bug:**
- [ ] Performance
- [ ] Design
- [ ] Security
- [x] Functional

**Severity Level:** 
- [ ]    5 - Data loss, data corruption or system unavailable  
- [ ]    4 - Important functionality is unavailable with no workaround 
- [x]    3 - Important functionality is unavailable but has a reasonable workaround 
- [ ]    2 - Secondary functionality is unavailable but has a reasonable workaround 
- [ ]    1 - Cosmetic issues or some functionality unavailable but has a simple workaround 
 
# Reproductive Steps
**Scenario 1 - Start retrieving eFolder button**
1. Go to eFolder
1. Enter a value in the Search field.
1. Click on the Search button. 
1. Get **Uh Oh** message.  (maybe)
1. Click on the **_search again_** link in the `Uh Oh Error` page.  The 'Dashboard` page is displayed.  
1. Enter another Veteran ID to search for and click Search button.  The `Manifest` page is displayed with files listed.
1. Click on the **_Start retrieving eFolder_** button.
1. Get **Uh Oh** message is received. 
1. Click on the **_search again_** link in the `Uh Oh Error` page.  The 'Dashboard` page is displayed. 
With the previous items in History.  If not, click search again for that same Veteran ID and repeat steps.

## Screenshots
**_Scenario 1 - Error Message_**
<img width="1080" alt="screen shot 2017-02-27 at 9 32 05 am" src="https://cloud.githubusercontent.com/assets/18075411/23365330/ae6f21f6-fcd0-11e6-921f-b9c5d0e91d33.png">

# Related Issues
**_Support Tickets_**
Dashboard | Start retrieving eFolder | Error message displayed - uh oh [#60](https://github.com/department-of-veterans-affairs/appeals-support/issues/60)
**_Engineering Issues_**
Efolder client side polling is too aggressive, and leads to HTTP backup. #373
[...]

PR #498, opened 2017-06-16
Anya/375 sikekiq not found

aroltsch:
connects #375 

1. `Record not found` issue -  Sidekiq is finding out about the job before the database record has it committed, so use after_commit on: :create.
2. Add test_after_commit gem to actually test `after_commit` hook in rspec.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -40,8 +40,8 @@ class Download < ActiveRecord::Base
     end
   end
 
-  # Sidekiq is finding out about the job before the database record has it committed,
-  # so use after_commit on: :create.
+  # Wait for the record to be committed to the DB and only then start the Sidekiq job
+  # Here is the issue: https://github.com/mperham/sidekiq/issues/322
   after_commit :start_fetch_manifest, on: :create
 
   def veteran_name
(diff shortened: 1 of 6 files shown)


amprokop:
@aroltsch — can we add a little bit of context on how this connects to #375 ? is there a recommended way we can test that this addresses the issue/how did you discover the fix?

app/models/download.rb
```diff
@@ -40,6 +40,10 @@ class Download < ActiveRecord::Base
     end
   end
 
+  # Sidekiq is finding out about the job before the database record has it committed,
```
amprokop: nit: this verb tense may not clearly show the meaning after this is merged.

Suggest something more like `To prevent Sidekiq from finding out...`

And maybe we can post that SO link too?

aroltsch:
@amprokop 
I think here is what happens: Active Record starts to create a download in the DB. Then immediately we fire up the Sidekiq job to use that download record. The Sidekiq job is a separate process so it cannot use the same in-memory download from the rails application so it goes to the DB to look for it. When it goes to look for it, the download record is not committed yet in the DB so it fails to find the record. Hence we get these logs:
```Record not found... Exception ActiveRecord::RecordNotFound: Couldn't find Download with 'id'=xxx```
The fix is to fire up the Sidekiq job after the record has been committed to the DB. 
The issue was happening randomly so in order to test it, the QA person would need to try a bunch of downloads and see if it happens again.

PR #503, opened 2017-06-22
Implement fetch document content api endpoint

aroltsch:
connects #504 

Steps to validate:
1. Test eFolder Express that it functions as normal
2. Test API call /api/v1/documents/document_id:
    a) enable the feature `reader_api`
    b) document_id is the DB ID of the document that should already exist in the EX DB
    c) example using CURL: `curl -i -X GET http://localhost:3000/api/v1/documents/23548`
    d) disable the feature flag as it is the only thing that protects the API (no authentication in place)


app/jobs/demo_get_download_files_job.rb
```diff
@@ -1,28 +1,13 @@
-class FakeDocumentService
```
aroltsch: I moved this into its own file lib/fakes/document_service.rb

app/services/download_documents.rb
```diff
@@ -58,33 +56,13 @@ def create_documents
     end
   end
 
-  def fetch_document(document, index)
```
aroltsch: I moved this method into a Document.rb

app/services/download_documents.rb
```diff
@@ -97,65 +75,27 @@ def download_contents
     end
   end
 
-  def download_dir
```
aroltsch: Moved this to Download.rb

aroltsch:
Tests for the API endpoint are coming.

app/services/download_documents.rb
```diff
@@ -31,9 +32,6 @@ class DownloadDocuments
   def initialize(opts = {})
     @download = opts[:download]
     @external_documents = DownloadDocuments.filter_documents(opts[:external_documents] || [])
-    @vbms_service = opts[:vbms_service] || VBMSService
```
aroltsch: Determine VBMSService, VVAService and S3Service based on the environment.

app/jobs/demo_get_download_files_job.rb
```diff
@@ -1,28 +1,13 @@
...
-
-  def self.fetch_document_file(_document)
-    sleep(rand(FakeDocumentService.max_time))
-    fail VBMS::ClientError if FakeDocumentService.errors && rand(5) == 3
-    fail VVA::ClientError if FakeDocumentService.errors && rand(5) == 2
-    "this is some document, woah!"
-  end
-end
-
 class DemoGetDownloadFilesJob < ActiveJob::Base
```
mdbenjam: Should we move this class as well? Instead of calling it DemoGetDownloadFilesJob, make a Fakes directory in the job directory, and make this class Fakes::GetDownloadFilesJob. Just so we have a consistent naming pattern. Curious on your thoughts.

app/models/document.rb
```diff
@@ -65,6 +67,55 @@ def type_name
...
+  def fetch_and_cache_in_s3
+    content = external_service.fetch_document_file(self)
+    S3Service.store_file(s3_filename, content)
+    content
+  end
+
+  def fetch_content
+    update_attributes!(started_at: Time.zone.now)
+    content = S3Service.fetch_content(s3_filename) || fetch_and_cache_in_s3
+    update_attributes!(
```
mdbenjam: What is the purpose of recording the started_at and completed_at times? Just so we know how long jobs take? Is there a disadvantage of saving the started_at time locally, and then saving started_at and completed_at at the same time? So we don't do two DB writes?

app/models/document.rb
```diff
@@ -65,6 +67,55 @@ def type_name
...
+      completed_at: Time.zone.now,
+      download_status: :success
+    )
+    content
+  end
+
+  def fetch_content_and_save(index = 0)
```
mdbenjam: Can we stream the content directly from S3 instead of saving it locally?

mdbenjam: After looking ahead, it seems we'll still need this for zipping. But can we make another method for streaming directly?

app/services/download_documents.rb
mdbenjam: 👍

config/initializers/s3.rb
```diff
@@ -0,0 +1 @@
+S3Service = (Rails.application.config.s3_enabled ? Caseflow::S3Service : Caseflow::Fakes::S3Service)
```
mdbenjam: I really like this model. So clean.

app/models/document.rb
aroltsch: I asked myself the same questions when I saw this :). I think the point of saving these times is to ensure that the fetch content was successful. If for any reason the file download was not successful with VBMS, the exception will be thrown and completed_at will not be set. @mdbenjam

app/models/document.rb
aroltsch: Yes.

app/jobs/demo_get_download_files_job.rb
aroltsch: Yes, I can do. However, I don't want to over-whelm this PR with a ton of re-factors. It might become massive :)
@mdbenjam

app/models/document.rb
aroltsch: Mark, I am going to create a ticket for this ^^.

app/jobs/demo_get_download_files_job.rb
aroltsch: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/507

app/controllers/api/v1/documents_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Api::V1::DocumentsController < Api::V1::ApplicationController
+  before_action :verify_feature_enabled
+
+  def show
+    document = Document.find(params[:id])
+    streaming_headers(document.mime_type, document.filename)
+    self.response_body = document.stream_file
```
mdbenjam: why do we need the `self.` here?

app/controllers/api/v1/documents_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+        "status": "404",
+        "title": "Document not found",
+        "detail": "A document with that ID was not found in our systems."
+      ]
+    }, status: 404
+  end
+
+  def verify_feature_enabled
+    # TODO: scope this to a current user
+    unauthorized unless FeatureToggle.enabled?(:reader_api)
```
mdbenjam: Can't we accomplish scoping to the current user by just adding a second parameter `user: current_user`? That said, I'm not sure we need to scope on a per user basis. I'm not if we want to have to turn this on for each of our users, since we're already gating reader on a per user basis. Why did you want to scope per user?

app/jobs/demo_get_download_files_job.rb
mdbenjam: 💯

app/models/document.rb
mdbenjam: Based on our conversation yesterday, and the fact that we'll be calling the streaming version, I'm cool with leaving this.

app/models/document.rb
```diff
@@ -65,6 +67,59 @@ def type_name
     type_description || TYPES[type_id.to_i] || vbms_filename
   end
 
+  def external_service
+    from_vva? ? VVAService : VBMSService
+  end
+
+  def stream_file
+    S3Service.stream_content(s3_filename) || fetch_and_cache_in_s3
```
mdbenjam: For my own understanding, what do we return in the streaming case. It seems like we can return two very different things, a streamed file versus file content.

app/models/document.rb
```diff
@@ -65,6 +67,59 @@ def type_name
...
+    update_attributes!(started_at: Time.zone.now)
+    content = S3Service.fetch_content(s3_filename) || fetch_and_cache_in_s3
+    update_attributes!(
+      completed_at: Time.zone.now,
+      download_status: :success
+    )
+    content
+  end
+
+   def fetch_and_cache_in_s3
```
mdbenjam: weird indentation here.

app/models/document.rb
```diff
@@ -65,6 +67,59 @@ def type_name
     type_description || TYPES[type_id.to_i] || vbms_filename
   end
 
+  def external_service
```
mdbenjam: Can this be private?

app/models/document.rb
```diff
@@ -65,6 +67,59 @@ def type_name
...
+    content = external_service.fetch_document_file(self)
+    S3Service.store_file(s3_filename, content)
+    content
+  end
+
+  def fetch_content_and_save(index = 0)
+    save_locally(fetch_content, index)
+  end
+
+  def save_locally(content, index)
```
mdbenjam: Is this called outside this file? We could probably make this private as well.

app/models/download.rb
```diff
@@ -114,6 +114,18 @@ def progress_percentage
     end
   end
 
+  def download_dir
```
mdbenjam: Do we have a unit test?

app/controllers/api/v1/documents_controller.rb
aroltsch: current_user is not available here as the API has its own application controller. Once we implement the authentication and the setting of the current_user, we will scope the flag to the user. We can also make the documents_controller inherit from the main application_controller. Thoughts?

app/models/document.rb
aroltsch: I don't think we gain any benefits from making things private in the model. The controller is where we need to worry about the private methods.

app/models/document.rb
aroltsch: Yes, there are two cases here: S3Service.stream_content(s3_filename) will return chunks and fetch_and_cache_in_s3 will return one huge chunk. Do you see a problem with that? Would suggest to always stream from s3? For example:
1. check s3 
2. if not in s3, fetch from VBMS and save in s3
3. check s3 again and stream

app/models/document.rb
aroltsch: Good catch!

app/models/document.rb
aroltsch: When you make things private, you cannot unit test then separately. I have a unit test for this method.

app/models/download.rb
aroltsch: I moved this method from download_documents.rb and you are right there is no unit test for it!

app/models/document.rb
aroltsch: The only downside, it might look messy :)

app/controllers/api/v1/documents_controller.rb
aroltsch: Strangely, if you omit `self` it will try to render the view and throw an exception that view is not found. But if you set the self.response_body it will not render the view. I could not understand the reason.

app/models/document.rb
```diff
@@ -65,6 +67,59 @@ def type_name
...
+  def external_service
+    from_vva? ? VVAService : VBMSService
+  end
+
+  def stream_file
+    S3Service.stream_content(s3_filename) || fetch_and_cache_in_s3
+  end
+
+  def fetch_content
+    update_attributes!(started_at: Time.zone.now)
```
amprokop: don't we already do this here:
https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/503/files#diff-a4389ce312191ae68bc9183bd4c4fae5L81

app/controllers/api/v1/documents_controller.rb
mdbenjam: Probably some rails :sparkles:

app/controllers/api/v1/documents_controller.rb
mdbenjam: What's the disadvantage of inheriting `ApplicationController` in `Api::V1::ApplicationController`? It seems we could get some of these methods defined in the `Api::V1::ApplicationController` for free, as well as things like the current_user (assuming the session token is what we expect).

app/models/document.rb
mdbenjam: I was under the impression from Shane, that making whatever is not called from outside the class private is a good practice, since then you don't have to test those private methods. Instead you can just test the API of the class. But I'm not a Ruby expert, and will defer to your opinion here.

app/models/document.rb
mdbenjam: I don't see a problem with it, I'm just trying to get a better handle of how this all works. So in the streaming case does rails send each chunk along as it receives it?

app/controllers/api/v1/documents_controller.rb
mdbenjam: Alex we will hit this directly from the browser.

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  private
+
+  def streaming_headers(mime_type, filename)
+    headers["Content-Type"] = mime_type
+    headers["Content-disposition"] = "attachment; filename=\"#{filename}\""
+
+    # Setting this to "no" will allow unbuffered responses for HTTP streaming applications
+    headers["X-Accel-Buffering"] = "no"
+    headers["Cache-Control"] ||= "no-cache"
```
aroltsch: for streaming yes

app/controllers/api/v1/documents_controller.rb
aroltsch: The disadvantage is that it will inherit all before_action methods from ApplicationController and in some cases it is not what we want.

app/models/document.rb
aroltsch: Yes @mdbenjam, however, Alex brought up a good point that these files are not too large and we don't necessarily need streaming. Thoughts?

app/models/document.rb
aroltsch: Yes, that's a good reason to make something private.

aroltsch:
@amprokop yes it will affect the old api as well. My goal was to make all of this cleaner because download_documents.rb is a mess :). I am also working on moving some methods out from the Document.rb into a new class to make it even cleaner.

app/controllers/api/v1/application_controller.rb
aroltsch: Great question! Buffering is when the server saves up data so as to send it out in fewer transactions. But in our case we would rather just start sending stuff. We also tell the server to stop caching because caching will cause the output to be sent in larger amounts and that's not what we want.

app/services/download_documents.rb
aroltsch: These methods are now in Fetcher class.

app/models/download.rb
aroltsch: I added the unit test.

app/controllers/api/v1/documents_controller.rb
mdbenjam: Okay, let's leave it for now. A possible future refactor would be to have a ApplicationController that everything can inherit from, and move the stuff we don't want into an eFolderExpress ApplicationController.

spec/requests/api/v1/documents_spec.rb
```diff
@@ -40,5 +39,20 @@
       expect(response.code).to eq("200")
       expect(response.body).to eq("hello there")
     end
+
+    it "returns 500 on any other error" do
+      allow_any_instance_of(Fetcher).to receive(:stream).and_raise("Much random error")
```
mdbenjam: nit: What's "Much random error"? Maybe just "Random Error"

spec/models/fetcher_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+      before do
+        allow(S3Service).to receive(:fetch_content).and_return("hello there")
+      end
+
+      it "should return the content from S3" do
+        expect(subject).to eq "hello there"
+      end
+    end
+
+    context "when file is not in S3" do
```
mdbenjam: Maybe add one more test here so that it ensures it only makes one call to VBMS if you call it twice (i.e. that it's caching properly).

app/controllers/api/v1/documents_controller.rb
```diff
@@ -0,0 +1,28 @@
+class Api::V1::DocumentsController < Api::V1::ApplicationController
+  before_action :verify_feature_enabled
+
+  def show
+    document = Document.find(params[:id])
```
mdbenjam: One thing I just thought of, we want to tell the user browser to cache the response here. This way they don't have to download the document more than once. We have this line in the Caseflow version of this:

`expires_in 30.days, public: true`

Can we add that here?

aroltsch: we can!

app/controllers/api/v1/documents_controller.rb
aroltsch: Yes, I thought of the same thing. I can create an issue for that re-factoring.

spec/models/fetcher_spec.rb
aroltsch: Ok, I see what you mean.

app/controllers/api/v1/application_controller.rb
amprokop: huh, interesting. was just curious about the caching header. i don't quite get why it would cause the output to be sent in larger amounts, would love to hear more about that sometime, but cool

app/controllers/api/v1/documents_controller.rb
aroltsch: @mdbenjam I think I am going to create a separate ticket for this, if you are okay with that?

spec/models/fetcher_spec.rb
```diff
@@ -61,5 +61,18 @@
...
+
+    context "when the file is in s3 after it has been cached" do
+      before do
+        allow(S3Service).to receive(:stream_content).and_return(nil)
+        allow(Fakes::DocumentService).to receive(:fetch_document_file).and_return("from VBMS")
+      end
+
+      it "should cache in s3 from VBMS and then serve from s3" do
+        expect(subject).to eq "from VBMS"
+        allow(S3Service).to receive(:stream_content).and_return("from s3")
```
mdbenjam: There probably should be a check that it makes a call to actually cache it. That said, I think this is sufficient for now.

app/controllers/api/v1/documents_controller.rb
mdbenjam: Yup, sure am.

Issue #496, opened 2017-06-14
List Documents | File number is not validated with non alphanumeric characters


kierachell:
# Description
When I go search for a filenumber, the `C` at the end is not removed,
so the List Documents just hangs...

## Reproduction Steps

**Scenario 1 -**

1. Go to eFolder
1. Search for `879325490C`
1. Watch spinner
1. Question life

**Scenario 1 -**

1. Go to eFolder
1. Search for `0909+1293`
1. Cause Sentry error
1. Question life

**Validated in the following Envrionment(s)**
- [ ] Dev
- [X] UAT
- [ ] Preprod
- [ ] Prod

## Screenshots
**_Scenario 1 - [You spin me right round]_**
<img width="1048" alt="screen shot 2017-06-14 at 2 44 37 pm" src="https://user-images.githubusercontent.com/18075411/27148933-0329f9e4-5110-11e7-8016-c4d7a3f30faa.png">

## Sentry:

https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/250/

Acceptance Criteria:
1. Display error if when file number is less than 8 characters long
2. Display error if when file number contains letters
3. Works correctly when file_number is at least 8 characters long and all digits

Header text in the error should say: "The veteran ID you entered was not valid."
Body text in the header should say: "Check to make sure you entered the ID correctly and try again."

amprokop:
@kierachell — is this new?

amprokop:
the S and C are only in VACOLS, i'm not sure eFolder Express should accept them.

kierachell:
It should not, but it also should not hang when there's something other than `S` at the end. It should either error out or suggest removing all non-numerics.

kierachell:
So yeah, there's a fix, but what is it?
@shellicious or @amprokop  - can you provide "expected behavior" and/or error message mocks to the original issue so that I know what to expect?

shellicious:
@kierachell I don't have much context about this, but I did add copy to the description of this issue: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/516, which I believe was resolved by this PR: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/506

kierachell:
**PASSED**
![faceroll](https://user-images.githubusercontent.com/18075411/27806998-fc3d77c2-600b-11e7-86e0-b84df580c7bb.gif)

PR #506, opened 2017-06-26
Validates user input

amprokop:
Shows error message to users if Veteran ID is invalid.

Connects https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/496 , https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/516

efolder_error_message

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -112,6 +112,11 @@ end
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   #gem 'web-console', '~> 2.0'
+  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
+  gem 'byebug', platforms: :ruby
+  gem 'pry'
+  gem 'rb-readline'
+
 
   # For windows
   gem 'tzinfo-data'
--- a/Gemfile
+++ b/Gemfile
@@ -106,6 +106,7 @@ group :test do
   # to save and open specific page in capybara tests
   gem 'launchy'
   gem 'database_cleaner'
+  gem 'test_after_commit'
 end
 
 group :development do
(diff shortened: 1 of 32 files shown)


shellicious:
@amprokop:

Header text in the error should say: "The veteran ID you entered was not valid."
Body text in the header should say: "Check to make sure you entered the ID correctly and try again."

Are there specific reasons we can give to help the user enter a valid ID? No special characters, numbers instead of letters?

Gemfile
```diff
@@ -112,6 +112,11 @@ end
 group :development do
   # Access an IRB console on exception pages or by using <%= console %> in views
   #gem 'web-console', '~> 2.0'
+  # Call 'byebug' anywhere in the code to stop execution and get a debugger console
+  gem 'byebug', platforms: :ruby
```
amprokop: adding byebug since a lot of devs use it

aroltsch: fine! I like pry :)

app/models/search.rb
```diff
@@ -14,6 +14,19 @@ class Search < ActiveRecord::Base
...
 
+  def valid_file_number?
+    number = sanitized_file_number
+    return true if number =~ /DEMO/
+
+    return false if /\D+/ =~ number
+    # We don't want to render an error message on initial pageload.
+    # We don't pass in user until we actually submit the form,
+    # so only check length if the user attribute is present.
+    return false if user && number.length < 8
```
aroltsch: Would this work?
```
return false if /\D+/ =~ number ||  (user && number.length < 8)
```

app/models/search.rb
```diff
@@ -14,6 +14,19 @@ class Search < ActiveRecord::Base
     access_denied: 3
   }
 
+  def valid_file_number?
```
aroltsch: Is there a unit test for this?

app/models/search.rb
amprokop: sure! that sounds fine. i thought it was a bit clearer to break it out into two conditions so i can explain the second one. happy to crunch it down into one though

aroltsch: Oh okay, I am gonna leave it up to you then.

app/models/search.rb
amprokop: Ah I'll write one

aroltsch: sorry Alex :)

PR #515, opened 2017-06-30
efolder needs to be able to handle RAILS_ENV=demo

kierachell:
Connects https://github.com/department-of-veterans-affairs/appeals-deployment/issues/663

I made a change that sets RAILS_ENV based on deploy environment, so the rake tasks need to be configured to be able to handle that.

This will just compile assets in production mode for `demo` RAILS_ENV.
In the future, this file can be edited to change that based on our needs

### DIFF
--- a/config/environments/demo.rb
+++ b/config/environments/demo.rb
@@ -77,8 +77,8 @@ Rails.application.configure do
   config.bgs_environment = ENV["BGS_ENVIRONMENT"] || "beplinktest"
   config.vva_wsdl = ENV["VVA_WSDL"]
 
-  config.s3_enabled = true
+  # Demo doesn't use S3, but that might change
+  config.s3_enabled = ENV["AWS_BUCKET_NAME"].present?
   config.s3_bucket_name = ENV["AWS_BUCKET_NAME"]
 
-  config.google_analytics_account = "UA-74789258-2"
 end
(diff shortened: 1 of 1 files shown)


config/environments/demo.rb
```diff
@@ -0,0 +1,84 @@
...
+  # Use default logging formatter so that PID and timestamp are not suppressed.
+  config.log_formatter = ::Logger::Formatter.new
+
+  # Do not dump schema after migrations.
+  config.active_record.dump_schema_after_migration = false
+
+  config.bgs_environment = ENV["BGS_ENVIRONMENT"] || "beplinktest"
+  config.vva_wsdl = ENV["VVA_WSDL"]
+
+  config.s3_enabled = true
```
nemodjuric: While I believe this is a true at the moment, maybe we should have ENV["AWS_BUCKET_NAME"].present? like in caseflow demo.rb

config/environments/demo.rb
```diff
@@ -0,0 +1,84 @@
...
+  config.s3_bucket_name = ENV["AWS_BUCKET_NAME"]
+
+  config.google_analytics_account = "UA-74789258-2"
```
nemodjuric: This line should be removed.

kierachell:
@nemodjuric Thank you for the catch; let me know if anything else is to be changed

Issue #505, opened 2017-06-22
eFolder Express Integration | Adding /api/v1/files/:file_number endpoint


mdbenjam:
Read the full eFolder Express integration tech spec [here](https://github.com/department-of-veterans-affairs/appeals-pm/blob/master/caseflow/tech_specs/prefetching_files_for_reader.md).

We need to create a new endpoint at api/v1/files in eFolder Express to allow other apps to request the list of files associated with a veteran's file number.

# AC
1. Verify the user can send the file number in a header called: file-number
1. Verify the user must send the station and css id in headers called: station-id and css-id
1. Verify the downloads table is indexed by file_number, user_id.
1. Verify when the endpoint is hit, the DB records are returned that match the most recent download entry that has that user_id (associated with the css_id and station_id) and file_number.
1. Verify that if the manifest_fetched_at field is older than three hours, VBMS is queried for the new manifest file, the new documents from VBMS are added to the DB, and the updated_at field in the download DB record is updated.
1. Verify that if the query param `download` is sent, then the job to download all the documents is started.
1. Verify the format of the returned data is:
```
{
  "data": {
    "id": "7",
    "type": "file",
    "attributes": {
      "manifest_fetched_at": "2017-07-07T02:49:51.936Z",
      "vbms_error": false,
      "vva_error": false,
      "documents": {[
        {
          "id": "412",
          "type_id": "123",
          "received_at": "2017-04-...",
          "external_document_id": "1234"
        },
        ...
      ]
    }
  }
}
```

kierachell:
### Should missing user ID cause Sentry error or just 500/Unauthorized?
`param is missing or the value is empty: user_id`
https://sentry.uat.ds.va.gov/sentry/efolder-uat-u4/issues/311/
```
{
"errors": [
{
"status": "500",
"title": "Unknown error occured",
"detail": "param is missing or the value is empty: user_id (Sentry event id: 9b8ae9f9dc2742a4a6abc10cbde07b2a)"
}
]
}
```

### Does the other have to be logged in? To be validated against BGS?

Without that the VBMS won't happen
`/api/v1/files/123?user_id=123123`
https://sentry.uat.ds.va.gov/sentry/efolder-uat-u4/issues/312/
```
{
"errors": [
{
"status": "500",
"title": "Unknown error occured",
"detail": "undefined method `ip_address' for nil:NilClass (Sentry event id: ea8b7a75bc054bf89f1e9f07334a6534)"
}
[...]

PR #526, opened 2017-07-07
Add file endpoint

mdbenjam:
Connects: #505 

This PR updates eFolder Express to pull the list of documents from our new API.
1. We updated the DB to index the download table by user and file number
1. We added a new file endpoint
1. We check to see if the files have already been downloaded in the last three hours, otherwise we refresh it from VBMS.
1. If the user requests us to download the file content, we start the job to do so.

### DIFF
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -1,0 +1,31 @@
+class Api::V1::FilesController < Api::V1::ApplicationController
+  before_action :verify_feature_enabled
+
+  def show
+    documents = Download.where(user_id: current_user, file_number: params[:id]).documents
+    streaming_headers(document.mime_type, document.filename)
+    # By setting the response body directly to an enumerator
+    # Rails will use the enumerator to send the data element by element,
+    # calling next on the enumerator to get the next chunk of data.
+    self.response_body = document.fetcher.stream
+  rescue ActiveRecord::RecordNotFound
+    file_not_found
+  end
+
+  private
+
+  def file_not_found
+    render json: {
+      "errors": [
+        "status": "404",
+        "title": "File not found",
+        "detail": "A case file with that ID was not found in our system."
+      ]
+    }, status: 404
+  end
+
+  def verify_feature_enabled
+    # TODO: scope this to a current user
+    unauthorized unless FeatureToggle.enabled?(:reader_api)
+  end
+end
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -22,7 +22,7 @@ class Api::V1::FilesController < Api::V1::ApplicationController
 
   def download
     @download ||= Download.includes(:documents).where(user_id: user_id, file_number: id).last ||
-      Download.create(user_id: user_id, file_number: id)
+      Download.create(user_id: user_id, file_number: id, no_fetch: true)
   end
 
   def user_id
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -1,6 +1,4 @@
 class Api::V1::FilesController < Api::V1::ApplicationController
-  before_action :verify_feature_enabled
-
   def show
     render json: json_files
   rescue ActiveRecord::RecordNotFound
@@ -13,7 +11,7 @@ class Api::V1::FilesController < Api::V1::ApplicationController
     download.force_fetch_manifest if !download.manifest_fetched_at || download.manifest_fetched_at < 3.hours.ago
     fail ActiveRecord::RecordNotFound if download.documents.empty?
 
-    download.start_cache_documents if download?
+    download.start_save_files_in_s3 if download?
 
     ActiveModelSerializers::SerializableResource.new(
       download,
@@ -33,11 +31,6 @@ class Api::V1::FilesController < Api::V1::ApplicationController
         "detail": "A case file with that ID was not found in our system."
       ]
     }, status: 404
-  end
-
-  def verify_feature_enabled
-    # TODO: scope this to a current user
-    unauthorized unless FeatureToggle.enabled?(:reader_api)
   end
 
   def user_id
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -11,10 +11,9 @@ class Api::V1::FilesController < Api::V1::ApplicationController
 
   def json_files
     download.force_fetch_manifest if !download.manifest_fetched_at || download.manifest_fetched_at < 3.hours.ago
+    fail ActiveRecord::RecordNotFound if download.documents.empty?
 
     download.start_cache_documents if download?
-
-    fail ActiveRecord::RecordNotFound if download.documents.empty?
 
     ActiveModelSerializers::SerializableResource.new(
       download,
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -8,10 +8,7 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   private
 
   def json_files
-    download.force_fetch_manifest_if_expired
-
-    fail ActiveRecord::RecordNotFound if download.documents.empty?
-    download.start_save_files_in_s3 if download?
+    download.prepare_files_for_api!(start_download: download?)
 
     ActiveModelSerializers::SerializableResource.new(
       download,
@@ -19,6 +16,9 @@ class Api::V1::FilesController < Api::V1::ApplicationController
     ).as_json
   end
 
+  # A caller can pass a download parameter in the url: `?download=true`.
+  # If it is present, then we return true from download? to signal
+  # that we want to download the content of the files from VBMS to S3.
   def download?
     params[:download]
   end
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -14,23 +14,12 @@ class Api::V1::FilesController < Api::V1::ApplicationController
 
     download.start_cache_documents if download?
 
+    raise ActiveRecord::RecordNotFound if download.documents.empty?
+
     ActiveModelSerializers::SerializableResource.new(
       download,
       each_serializer: Serializers::V1::DownloadSerializer
     ).as_json
-  end
-
-  def download
-    @download ||= Download.includes(:documents).where(user_id: user_id, file_number: id).last ||
-      Download.create(user_id: user_id, file_number: id, no_fetch: true)
-  end
-
-  def user_id
-    params.require(:user_id)
-  end
-
-  def id
-    params[:id]
   end
 
   def download?
@@ -50,5 +39,19 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   def verify_feature_enabled
     # TODO: scope this to a current user
     unauthorized unless FeatureToggle.enabled?(:reader_api)
+  end
+
+  private
+
+  def user_id
+    params.require(:user_id)
+  end
+
+  def id
+    params[:id]
+  end
+
+  def download
+    @download ||= Download.find_or_create_by_user_and_file(user_id, id)
   end
 end
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -10,7 +10,7 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   private
 
   def json_files
-    download.force_fetch_manifest if (!download.manifest_fetched_at || download.manifest_fetched_at < 1.second.ago)
+    download.force_fetch_manifest if (!download.manifest_fetched_at || download.manifest_fetched_at < 3.hours.ago)
 
     download.start_cache_documents if download?
 
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -8,9 +8,9 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   private
 
   def json_files
-    download.force_fetch_manifest if !download.manifest_fetched_at || download.manifest_fetched_at < 3.hours.ago
-    fail ActiveRecord::RecordNotFound if download.documents.empty?
+    download.force_fetch_manifest_if_expired
 
+    fail ActiveRecord::RecordNotFound if download.documents.empty?
     download.start_save_files_in_s3 if download?
 
     ActiveModelSerializers::SerializableResource.new(
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -10,11 +10,11 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   private
 
   def json_files
-    download.force_fetch_manifest if (!download.manifest_fetched_at || download.manifest_fetched_at < 3.hours.ago)
+    download.force_fetch_manifest if !download.manifest_fetched_at || download.manifest_fetched_at < 3.hours.ago
 
     download.start_cache_documents if download?
 
-    raise ActiveRecord::RecordNotFound if download.documents.empty?
+    fail ActiveRecord::RecordNotFound if download.documents.empty?
 
     ActiveModelSerializers::SerializableResource.new(
       download,
@@ -40,8 +40,6 @@ class Api::V1::FilesController < Api::V1::ApplicationController
     # TODO: scope this to a current user
     unauthorized unless FeatureToggle.enabled?(:reader_api)
   end
-
-  private
 
   def user_id
     params.require(:user_id)
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -2,17 +2,40 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   before_action :verify_feature_enabled
 
   def show
-    documents = Download.where(user_id: current_user, file_number: params[:id]).documents
-    streaming_headers(document.mime_type, document.filename)
-    # By setting the response body directly to an enumerator
-    # Rails will use the enumerator to send the data element by element,
-    # calling next on the enumerator to get the next chunk of data.
-    self.response_body = document.fetcher.stream
+    render json: json_files
   rescue ActiveRecord::RecordNotFound
     file_not_found
   end
 
   private
+
+  def json_files
+    download.force_fetch_manifest if (!download.manifest_fetched_at || download.manifest_fetched_at < 1.second.ago)
+
+    download.start_cache_documents if download?
+
+    ActiveModelSerializers::SerializableResource.new(
+      download,
+      each_serializer: Serializers::V1::DownloadSerializer
+    ).as_json
+  end
+
+  def download
+    @download ||= Download.includes(:documents).where(user_id: user_id, file_number: id).last ||
+      Download.create(user_id: user_id, file_number: id)
+  end
+
+  def user_id
+    params.require(:user_id)
+  end
+
+  def id
+    params[:id]
+  end
+
+  def download?
+    params[:download]
+  end
 
   def file_not_found
     render json: {
--- a/app/jobs/save_files_in_s3_job.rb
+++ b/app/jobs/save_files_in_s3_job.rb
@@ -1,10 +1,10 @@
-class CacheFilesJob < ActiveJob::Base
+class SaveFilesInS3Job < ActiveJob::Base
   queue_as :default
 
   def perform(download)
     download_documents = DownloadDocuments.new(download: download)
 
-    download_documents.download_contents(only_cache: true)
+    download_documents.download_contents(save_locally: false)
   end
 
   def max_attempts
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -196,6 +196,15 @@ class Download < ActiveRecord::Base
     end
   end
 
+  # are we looking things up twice because of the after_commit hook?
+  def force_fetch_manifest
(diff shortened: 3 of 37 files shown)


spec/requests/api/v1/file_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+        allow(CacheFilesJob).to receive(:perform_later)
+
+        get "/api/v1/files/#{download.file_number}?user_id=#{user.id}&download=true"
+
+        expect(CacheFilesJob).to have_received(:perform_later)
+      end
+    end
+  end
+
+  it "returns 500 on any other error" do
```
mdbenjam: Can we somehow test this on the application controller, since this effectively duplicates a similar test in documents_spec.rb

app/services/download_documents.rb
```diff
@@ -38,31 +38,37 @@ def create_documents
...
 
-  def download_contents
+  def download_contents(only_cache: false)
     @download.update_attributes!(started_at: Time.zone.now)
     @download.documents.where(download_status: 0).each_with_index do |document, index|
       before_document_download(document)
       begin
-        document.save_locally(document.fetcher.content, index)
-        @download.touch
+        if only_cache
```
aroltsch: How about something like that:

```
content = document.fetcher.content
document.save_locally(content, index) unless only_cache
@download.touch
```

Also, `only_cache` is a bit unclear, maybe we can call it `save_locally`?

app/models/download.rb
```diff
@@ -194,12 +196,28 @@ def bgs_service
...
   end
 
   private
 
+  # Do not fetch the manifest on the after_create hook if no_fetch is true. This value is
+  # set to true when we create our record through find_or_create_by_user_and_file, since
+  # in that case we want to call force_fetch_manifest to do it synchronously.
   def start_fetch_manifest
-    demo? ? Fakes::DownloadManifestJob.perform_later(self) : DownloadManifestJob.perform_later(self)
+    (demo? ? Fakes::DownloadManifestJob.perform_later(self) : DownloadManifestJob.perform_later(self)) unless no_fetch
```
shanear: `perform_later` is used to get a more snappy response (and show the big loading icon) on the UI efolder express.

app/models/download.rb
```diff
@@ -194,12 +196,28 @@ def bgs_service
...
+
+  def start_cache_documents
+    CacheFilesJob.perform_later(self)
   end
 
   private
 
+  # Do not fetch the manifest on the after_create hook if no_fetch is true. This value is
+  # set to true when we create our record through find_or_create_by_user_and_file, since
+  # in that case we want to call force_fetch_manifest to do it synchronously.
```
shanear: I agree. The `after_commit` + `no_fetch` logic is janky af.

app/services/download_documents.rb
```diff
@@ -38,31 +38,37 @@ def create_documents
...
-          type_id: external_document.doc_type || external_document.type_id,
-          type_description: external_document.try(:type_description) || TYPES[external_document.doc_type.to_i],
-          source: external_document.source,
-          mime_type: external_document.mime_type,
-          received_at: external_document.received_at,
-          jro: external_document.try(:jro),
-          ssn: external_document.try(:ssn),
-          downloaded_from: external_document.try(:downloaded_from) || "VBMS"
-        )
+        @download.documents.find_or_initialize_by(document_id: external_document.document_id).tap do |t|
```
shanear: also curious 🍎

spec/models/download_spec.rb
```diff
@@ -238,6 +238,27 @@
...
+    let(:file_number) { "123" }
+
+    context "retrieves most recent existing record" do
+      let!(:old_download) { Download.create(user: user, file_number: file_number) }
+      let!(:new_download) { Download.create(user: user, file_number: file_number) }
+
+      it { is_expected.to eq(new_download) }
+    end
+
+    context "creates new download when no one exists" do
```
shanear: it would be more idomatic to write this:

```
context "when no download exists" do
  it "creates a new download" do
    ...
  end
end
```

spec/models/download_spec.rb
```diff
@@ -238,6 +238,27 @@
     end
   end
 
+  context ".find_or_create_by_user_and_file" do
+    subject { Download.find_or_create_by_user_and_file(user.id, file_number) }
+    let(:user) { User.create(css_id: "WALTER", station_id: "123") }
+    let(:file_number) { "123" }
+
+    context "retrieves most recent existing record" do
+      let!(:old_download) { Download.create(user: user, file_number: file_number) }
```
shanear: I would explicitly set the `created_at` values of these two objects to make this test more explicit. Plus stuff like this can cause race conditions and flakey tests.

IE:
```
let!(:old_download) { Download.create(user: user, file_number: file_number, created_at: 2.seconds.ago) }
let!(:new_download) { Download.create(user: user, file_number: file_number, created_at: 1.second.ago) }
```

spec/models/download_spec.rb
```diff
@@ -289,4 +310,20 @@
       it { is_expected.to eq(100) }
     end
   end
+
+  context "#force_fetch_manifest" do
+    it "starts the manifest job" do
+      allow(DownloadManifestJob).to receive(:perform_now)
+      download.force_fetch_manifest
+      expect(DownloadManifestJob).to have_received(:perform_now)
```
shanear: you can write:
```
expect(DownloadManifestJob).to receive(:perform_now)
```
on line 316 and eliminate this line.

spec/models/download_spec.rb
```diff
@@ -289,4 +310,20 @@
...
+    end
+  end
+
+  context "#start_cache_documents" do
+    it "starts the manifest job" do
+      allow(CacheFilesJob).to receive(:perform_later)
+      download.start_cache_documents
+      expect(CacheFilesJob).to have_received(:perform_later)
```
shanear: see the comment above, its applicable here too.

spec/requests/api/v1/file_spec.rb
```diff
@@ -0,0 +1,156 @@
...
+
+    let(:vbms_document2) do
+      OpenStruct.new(
+        document_id: "2",
+        received_at: "3/4/2017",
+        type_id: "124"
+      )
+    end
+
+    context "download object doesn't exist" do
```
shanear: all of this testing stuff gets way simpler if you move some of this logic into the model

spec/requests/api/v1/file_spec.rb
shanear: If we've already tested this functionality on another route, i don't think its necessary to duplicate it.

spec/services/download_documents_spec.rb
```diff
@@ -51,6 +51,30 @@
...
+      let(:external_documents2) do
+        [
+          OpenStruct.new(document_id: "1", filename: "filename.pdf", doc_type: "124",
+                         source: "SRC", received_at: Time.zone.now, type_id: "124",
+                         mime_type: "application/pdf", type_description: "VA 9 Appeal to Board of Appeals"),
+          OpenStruct.new(document_id: "5", received_at: 1.hour.ago)
+        ]
+      end
+
+      let(:download_documents2) do
```
shanear: `download_documents2` 😢 😢

spec/services/download_documents_spec.rb
```diff
@@ -101,6 +125,20 @@
       end
     end
 
+    context "when only_cache is true" do
+      before do
+        download_documents.create_documents
+      end
+
+      it "files are cached and not saved" do
```
shanear: ideally these should read like a sentence:
`it "caches files but does not save them"`

app/models/download.rb
mdbenjam: I was trying to avoid larger changes. Honestly, I don't see why not, though that would probably be a pretty large architecture change. I've been told (haven't looked into the code) that the front-end polls to get the status of this job.

app/services/download_documents.rb
mdbenjam: Soooo we need a way to update metadata. While we assume the documents themselves are immutable--the meatadata is not. People can change file names etc. In the old eFolder Express model, if a user downloaded the same case within three days of the previous download, then the download object would be looked up and used as is. In our new model, if you access it via the API then we only create one download object and use it forever. (If a user downloads the case via efolder express after hitting it from the API, then we might create new entries.) The idea is we only need one download entry to use a cached version of whatever is in VBMS, and we should keep that one entry up to date rather than creating new entries.

app/models/download.rb
shanear: Thats correct mark. Making this serial would make the UI in eFolder express way less responsive (and when VBMS is slow, we even risk timing out). We couldn't show our fancy loading icon.

app/services/download_documents.rb
shanear: why not just delete the old ones and start from scratch instead of going through the rigamarole of mutating existing docs? Seems simpler and safer.

Better yet, why delete the old ones? It's not like they are taking up all that much space, and it'd be nice to know when folks are downloading docs, and what the state of the docs were when they were downloaded.

mdbenjam: I disagree about not taking up much space. Yes the download object might not take up much space, but if we create a new entry, then we have to create new document entries for every single document in the eFolder. That means 100-1000 document entries for every time a user downloads it again. I really think there should only be one download per (case file id + user id). But we haven't been living under that assumption with eFolder Express.

spec/services/download_documents_spec.rb
mdbenjam: Are the 😢 just because of a non-descriptive name? I tried to make it better, but still might not be clear enough.

app/services/download_documents.rb
shanear: A download & document represents the state of an eFolder at the time it was downloaded. Maintaining that property is pretty important since we don't want to get in the business of maintaining sync'ed state between VBMS and Caseflow.

For the same reason Redux is easier to reason about (since is immutable), keeping downloads/documents immutable will make things much easier.

If the only downside is space (correct me if there are other downsides?), we can discuss that:

Each row is probably on the order of 100 bytes. Which means each duplicated download takes up 500*100 = 50kb. Lets say each of 50k appeals every year has 1 duplicate download (which is probably generous, since most cases are done within 3 days anyway). Then 50kb*50000 = 2.5gb per year. That's not that much space.

Additionally, if we are concerned about the space piling up. We can always clean up old downloads every month/3 months/year.

spec/services/download_documents_spec.rb
shanear: its almost always better to add more specificity to the variable name than to add a number

app/models/download.rb
mdbenjam: I really hate this too... I don't think it's as easy as putting it after download.save! But let me look into it.

app/models/download.rb
```diff
@@ -194,12 +196,29 @@ def bgs_service
...
+
+    def find_or_create_by_user_and_file(user_id, file_id)
+      Download.includes(:documents).where(user_id: user_id, file_number: file_id).last ||
+        Download.create(user_id: user_id, file_number: file_id, no_fetch: true)
+    end
+  end
+
+  def force_fetch_manifest_if_expired
+    (demo? ? Fakes::DownloadManifestJob.perform_now(self) : DownloadManifestJob.perform_now(self)) if
+      !manifest_fetched_at || manifest_fetched_at < 3.hours.ago
```
aroltsch: Would this achieve the same result:

```
demo? ? Fakes::DownloadManifestJob.perform_now(self) : DownloadManifestJob.perform_now(self)
```

app/services/download_documents.rb
mdbenjam: A few things here. There will be significantly more duplicated downloads than you just suggested. The API refresh logic is different than the eFolder Express logic on when we re-download. In this case we try to download again if the case is older than 3 hours, since it's really important that attorneys have every piece of evidence. On every refresh there will be a new download added. Your point about space is valid, even with these duplicates we're at most probably around 25gb/year. But I don't see how maintaining all of these downloads helps us reason about anything.

I agree that we're not and (should not try) going to maintain a perfectly sync'd state between VBMS and Caseflow, but for this application we need a pretty up to date cache of what is in VBMS. My ideal model for this is that there is one record for each user + file_number combination so that we can just find that one entry and serve it.

The only other concern I have is I think making new entries makes the VBMS down logic more complicated. Currently if VBMS is down, we can just serve the current entry we're trying to update. But if we are always creating new entries, we either create a new entry and serve it, or we return the most recently created one. It's not a huge deal, but I think it's a more complicated model.

shanear: > But I don't see how maintaining all of these downloads helps us reason about anything.
In the same way that maintaining a immutable state helps us reason about the front end code. We can see the state of a document at a particular point in time, which helps for diagnosing all sorts of issues and creating a history of how things happened on a folder.

Additionally, creating a new record from scratch is almost always less complicated and easier to understand than mutating an existing record, especially when your new record doesn't actually need any information from the old record.

Also, I'm not suggesting that we keep them forever. We could very well have a clean up job. Or clean them up immediately after.

> we need a pretty up to date cache of what is in VBMS. My ideal model for this is that there is one record for each user + file_number combination so that we can just find that one entry and serve it.

Having one record for each user + file number doesn't help keep the documents more up to date, decreasing the 3hr cache limit does. And we can do that no matter which route we choose.

> The only other concern I have is I think making new entries makes the VBMS down logic more complicated. Currently if VBMS is down, we can just serve the current entry we're trying to update. But if we are always creating new entries, we either create a new entry and serve it, or we return the most recently created one. It's not a huge deal, but I think it's a more complicated model.

Not sure I agree here. If VBMS is down we return the most recent download/documents (which is what we already do if the download is < 3 hours old). If it's not we create a new one. That seems very simple to me.

shanear: I'm down to chat about this over the ☎️ if that would be more helpful 😄

joofsh: > A download & document represents the state of an eFolder at the time it was downloaded.

@shanear what you're describing here makes sense from the perspective of eFolder Express as a UI, but it does not make sense for API design for a few reasons:

- Making a request `GET /files/1234` for a proxy/caching API like this should _not_ create hundreds of new rows. If I am making a request for the file 1234, then I expect to receive that 1 singular resource object. From a REST design standpoint, making new rows for each file/document is janky.

- It slows down the API request. On GET request creating a new download + hundreds of documents will have a serious performance impact. We should try to do this as infrequently as possible. This `find_or_initialize` and a later save will only do `UPDATE`s on the few documents that actually have their metadata change. 

- The physical space may not matter, but the impact on performance is real. Queries become slower as the tables grow in size. Even now the architecture for how quickly the download/documents table grow in eFolder Express is not scalable. Think how big these tables will be in 5 years. Eventually we are going to need to start pruning data. Why add the complexity of having a background job to delete data? Better if we can minimize the amount of data we create to begin with. 

- Creating new documents is not an extensible design.  What if in the future we want to change the cache rate to 1 hour or 10 minutes? All of a sudden we have _explosive_ growth in the number of documents we create. We would be designing ourselves into a corner.

- From a practical standpoint, attorneys review cases longer than 3 hours. This will definitely trigger a redownload for almost all cases

That said, what you're describing about immutability and helping to debug is valid, but we don't do that for any other objects. If we want to know the state of an object on every update/change, then we should add audit/trigger tables.  By being a separate table we gain the information you desire about changes over time, without impacting performance on the main documents table.

app/controllers/api/v1/files_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def show
+    render json: json_files
+  rescue ActiveRecord::RecordNotFound
+    file_not_found
+  end
+
+  private
+
+  def json_files
```
mdbenjam: Now my model is fat AF

spec/requests/api/v1/file_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      ]
+    end
+
+    context "retrieve the documents" do
+      before do
+        Timecop.freeze(Time.utc(2015, 1, 1, 17, 0, 0))
+        download.update_attributes!(manifest_fetched_at: Time.zone.now - 4.hours)
+      end
+
+      let(:response_body) do
```
mdbenjam: Any thoughts on how to better format this?

app/services/download_documents.rb
shanear: > Making a request GET /files/1234 for a proxy/caching API like this should not create hundreds of new rows. If I am making a request for the file 1234, then I expect to receive that 1 singular resource object. From a REST design standpoint, making new rows for each file/document is janky.

On the first fetch, it is creating hundreds of new rows though. I don't like that either (especially regarding REST design) and expressed my concern about that in the original API tech spec here https://github.com/department-of-veterans-affairs/appeals-pm/pull/1299#discussion_r119879270.

> It slows down the API request. On GET request creating a new download + hundreds of documents will have a serious performance impact. We should try to do this as infrequently as possible. This find_or_initialize and a later save will only do UPDATEs on the few documents that actually have their metadata change.

It's already quite slow. In this request you are:
- querying VBMS (can take > 10s)
- individually [finding](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/526/files#diff-a4389ce312191ae68bc9183bd4c4fae5R41) and [updating](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/526/files#diff-a4389ce312191ae68bc9183bd4c4fae5R53) each document in the Caseflow DB.

Is finding + updating that much faster than inserting? Debatable.

If we want to talk about speed. We need to talk about making work in this call asynchronous.

> The physical space may not matter, but the impact on performance is real. Queries become slower as the tables grow in size. Even now the architecture for how quickly the download/documents table grow in eFolder Express is not scalable. Think how big these tables will be in 5 years. Eventually we are going to need to start pruning data. Why add the complexity of having a background job to delete data? Better if we can minimize the amount of data we create to begin with.

I'm not suggesting we save old states forever. We can wipe them every year/month/week/day depending on our performance needs.

spec/requests/api/v1/file_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        "{\"data\":{\"id\":\"4\",\"type\":\"file\",\"attributes\":"\
+        "{\"manifest_fetched_at\":\"2015-01-01T17:00:00.000Z\","\
+        "\"documents\":[{\"document_id\":\"{3333-3333}\",\"type_id\":"\
+        "\"825\",\"received_at\":\"2015-09-06T01:00:00.000Z\"},"\
+        "{\"document_id\":\"1\",\"type_id\":\"123\",\"received_at\":"\
+        "\"2017-02-01T00:00:00.000Z\"},{\"document_id\":\"2\",\"type_id\":"\
+        "\"124\",\"received_at\":\"2017-04-03T00:00:00.000Z\"}]}}}"
+      end
+
+      it "returns existing and new files" do
```
mdbenjam: Now that we have unit tests for everything, I've updated the feature test to just look at the result of the api call. Does that seem like a good move?

shanear: The test pyramid thanks you
test-pyramid

spec/requests/api/v1/file_spec.rb
shanear:
```
{
  data: {
    things: "things"
  }
}.to_json
```
might do the trick

app/controllers/api/v1/files_controller.rb
shanear: a37iewzh8fx6bn27tpuy

spec/requests/api/v1/file_spec.rb
mdbenjam: How did the Egyptians build the ancient test pyramids without modern test frameworks? One of the mysteries of history.

kierachell: They packaged each pharaoh in a new non-backwards-compatible binary...duh

app/jobs/save_files_in_s3_job.rb
```diff
@@ -0,0 +1,13 @@
+class SaveFilesInS3Job < ActiveJob::Base
+  queue_as :default
+
+  def perform(download)
+    download_documents = DownloadDocuments.new(download: download)
+
+    download_documents.download_contents(save_locally: false)
+  end
```
aroltsch: @mdbenjam at what point are you creating document records as you are not passing external_documents to DownloadDocuments? And I don't see you calling VBMSService.fetch_documents_for from anywhere. Confused.

aroltsch: Duh! I see it now. In force_fetch_manifest_if_expired!

app/services/download_documents.rb
mdbenjam: I think I'd like to keep it the way it is. I went down the road of trying to delete and create new records, but it broke some tests, and got a little tricky to get working again. I also thought of some new questions. If VBMS is down but VVA is up, we want to update the records from VVA, but not delete the old VBMS records we have. I think updating is the right thing to do for now.

spec/models/download_spec.rb
```diff
@@ -289,4 +303,80 @@
...
+      end
+    end
+
+    context "when the manifest was fetched less than 3 hours ago" do
+      before do
+        download.update_attributes!(manifest_fetched_at: Time.zone.now - 2.hours)
+      end
+
+      it "does not start the manifest job" do
+        allow(DownloadManifestJob).to receive(:perform_now)
```
shanear: i don't think this is necessary

spec/models/download_spec.rb
```diff
@@ -289,4 +303,80 @@
...
+      end
+    end
+
+    context "when the manifest was fetched more than 3 hours ago" do
+      before do
+        download.update_attributes!(manifest_fetched_at: Time.zone.now - 4.hours)
+      end
+
+      it "starts the manifest job" do
+        allow(DownloadManifestJob).to receive(:perform_now)
```
shanear: you can probably put this in a `before` block at the top of the context

aroltsch:
Awesome, Mark!

PR #537, opened 2017-07-18
Fix files endpoint

mdbenjam:
Connects: #505

In UAT the files endpoint failed validation because we triggered a call to BGS, but lacked the required information to make a request. It turns out the call to BGS isn't even necessary, so we updated the call to only call BGS when needed.

We also updated the returned id from the endpoint to reflect the internal id, since that is what is needed to request the pdfs from the documents endpoint.

**Test Plan**
- [ ] Try the files endpoint when connected to UAT
- [ ] Try the normal UI download path when connected to UAT


app/models/download.rb
```diff
@@ -53,7 +52,30 @@ def demo?
   end
 
   def missing_veteran_info?
-    file_number && (!veteran_first_name || !veteran_last_name || !veteran_last_four_ssn)
+    file_number && (!self[:veteran_last_four_ssn] ||
```
mdbenjam: Since we overrode the original accessor methods if we call them, we'll call fetch_veteran_info which will call missing_veteran_info? so I think we'll get an infinite loop.

app/models/download.rb
```diff
@@ -53,7 +52,30 @@ def demo?
...
+    # TODO: (alex) we should maybe be setting the fake
+    # or real bgs service on an instance level, rather
+    # than a class. Refactor the class method `bgs_service`
+    # into an instance one.
+    if missing_veteran_info?
+      bgs_service = demo? ? Fakes::BGSService : Download.bgs_service
+
+      veteran_info = bgs_service.fetch_veteran_info(file_number)
+
+      reload.update_attributes!(
```
amprokop: why `reload`?

spec/models/download_spec.rb
```diff
@@ -380,4 +384,48 @@
...
+      end
+
+      it "doesn't fetch veteran info" do
+        download.fetch_veteran_info
+        expect(download.veteran_first_name).to eq(different_veteran_first_name)
+      end
+    end
+
+    context "doesn't have veteran info" do
+      it "savevs it" do
```
amprokop: typo

app/models/download.rb
mdbenjam: Oddly it said the record was out of sync, not sure how. But the reload fixed it.

app/models/download.rb
aroltsch: OK! Could you please add a comment @mdbenjam

Issue #535, opened 2017-07-18
API | Make secure


mdbenjam:
We need a way to authenticate callers to our new API, and we should not be passing sensitive information as a get parameter.

# AC
1. Verify the apis only respond with information when either the static API key is passed or when the user has a valid token.
1. Verify that for the files api the file_number is passed in the header.
1. Verify that for the files api the station_id and css_id are passed in the header.

astewarttistatech:
@mdbenjam ISo tried a few scenarios that have been tried are: 
 
1. Station ID not passed:  

{"errors":[{"status":"500","title":"Unknown error occured","detail":"ERROR:  null value in column \"station_id\" violates not-null constraint\nDETAIL:  Failing row contains (58, CASEFLOW_317, null, null, 2017-07-27 18:06:07.178415, 2017-07-27 18:06:07.178415, 0).\n (Sentry event id: cda3e1133bc448849b7ad52a5ed4aa98)"}]}


There's a Sentry error there and I'm not sure if this is to be expected

2. No file number passed: 

{"data":{"id":"659","type":"file","attributes":{"manifest_fetched_at":null,"vbms_error":true,"vva_error":false,"documents":[]}}}

Yet, this passed with a 200 and a vbms_error value of true. To be expected?

3. Invalid Token: 

{"status":"unauthorized"}  

Is this to be expected?

4. Invalid css_id: 

The response is too long for this ticket, but it does return a 200 and a response as if it were valid. I also do not know if this is to be expected.

mdbenjam:
#3 is correct. #1, 2, and 4, are undefined, but should probably return an error. I will post an update.

astewarttistatech:
@mdbenjam Did you post an update?

mdbenjam:
@astewarttistatech Sorry I merged a change to address the other issues. We return 400's if a header is not provided.

astewarttistatech:
Errors are now being passed for scenarios 1, 2, and 4.

**PASSED**
Env: Terminal/Console

Steps to Validate: 
1. Pass station_id, css_id, and file_numbers incorrectly to see if a correct error appears.
2. Pass station_id, css_id, and file_numbers correctly to see if the results touch the API correctly

PR #539, opened 2017-07-19
Add authentication to API

mdbenjam:
Connects: #535 

We need to add authentication to our two APIs. Unfortunately, since they are accessed differently, we need two different authentication mechanisms.

The /files endpoint is hit from the Caseflow backend, so we will use a static API key saved as an environment variable to authenticate.

The /documents endpoint is hit from the Caseflow frontend, so we will use the user token (since the token is shared between Caseflow and eFolder Express) to authenticate. We will just check that the token has the Reader role.

Furthermore, since the veteran id is PII we move it from a get parameter to a value passed in the header. We also pass the station_id and css_id in the header.

Also starting to return the vbms_document_id since we'll need that on the Caseflow side.

**Test Plan**
- [ ] Curl the API with `curl "http://localhost:3000/api/v1/files" -H "Authorization: Token token=token" -H "css_id: CSFLOW4" -H "station_id: 283" -H "veteran_id: DEMO123"`
- [ ] Test the documents api by loging into Caseflow locally and then hitting the endpoint in the browser.
- [ ] After merging, validate assumptions in UAT.

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -37,7 +37,11 @@ class Api::V1::ApplicationController < BaseController
     # authenticates, then we create the current user based on the css_id and station_id
     # passed in the header. Otherwise we try to create the current user from the session
     # and authorize based on the presence of the Reader role.
-    return unauthorized if !(authenticate_with_token || user_has_role)
+    if authenticate_with_token
+      @current_user = User.find_or_create_by(css_id: css_id, station_id: station_id) if authenticate_with_token
+    elsif !user_has_role
+      unauthorized
+    end
   end
 
   def station_id
@@ -46,10 +50,5 @@ class Api::V1::ApplicationController < BaseController
 
   def css_id
     request.headers["HTTP_CSS_ID"]
-  end
-
-  def current_user
-    return @current_user ||= User.find_or_create_by(css_id: css_id, station_id: station_id) if authenticate_with_token
-    super
   end
 end
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -18,15 +18,6 @@ class Api::V1::ApplicationController < ActionController::Base
 
   private
 
-  def streaming_headers(mime_type, filename)
-    headers["Content-Type"] = mime_type
-    headers["Content-disposition"] = "attachment; filename=\"#{filename}\""
-
-    # Setting this to "no" will allow unbuffered responses for HTTP streaming applications
-    headers["X-Accel-Buffering"] = "no"
-    headers["Cache-Control"] ||= "no-cache"
-  end
-
   def ssl_enabled?
     Rails.env.production?
   end
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -1,6 +1,6 @@
 class Api::V1::ApplicationController < BaseController
   protect_from_forgery with: :null_session
-  before_action :verify_reader_api_enabled
+  before_action :authorize
 
   rescue_from StandardError do |error|
     Raven.capture_exception(error)
@@ -20,8 +20,30 @@ class Api::V1::ApplicationController < BaseController
     render json: { status: "unauthorized" }, status: 401
   end
 
-  def verify_reader_api_enabled
-    # TODO: scope this to a current user
-    unauthorized unless FeatureToggle.enabled?(:reader_api)
+  def authenticate_with_token
+    @authenticate_with_token ||= authenticate_with_http_token do
+      |token, _options| token == Rails.application.config.api_key
+    end
+  end
+
+  def user_has_role
+     current_user && (current_user.can?("Reader") || current_user.can?("System Admin"))
+  end
+
+  def authorize
+    return unauthorized unless authenticate_with_token || user_has_role
+  end
+
+  def station_id
+    request.headers["HTTP_STATION_ID"]
+  end
+
+  def css_id
+    request.headers["HTTP_CSS_ID"]
+  end
+
+  def current_user
+    return @current_user ||= User.find_or_create_by(css_id: css_id, station_id: station_id) if authenticate_with_token
+    super
   end
 end
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -1,8 +1,5 @@
-class Api::V1::ApplicationController < ActionController::Base
+class Api::V1::ApplicationController < BaseController
   protect_from_forgery with: :null_session
-
-  force_ssl if: :ssl_enabled?
-  before_action :strict_transport_security
 
   rescue_from StandardError do |error|
     Raven.capture_exception(error)
@@ -17,14 +14,6 @@ class Api::V1::ApplicationController < ActionController::Base
   end
 
   private
-
-  def ssl_enabled?
-    Rails.env.production?
-  end
-
-  def strict_transport_security
-    response.headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains" if request.ssl?
-  end
 
   def unauthorized
     render json: { status: "unauthorized" }, status: 401
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -1,6 +1,6 @@
 class Api::V1::ApplicationController < BaseController
   protect_from_forgery with: :null_session
-  before_action :authorize
+  before_action :authenticate_or_authorize
 
   rescue_from StandardError do |error|
     Raven.capture_exception(error)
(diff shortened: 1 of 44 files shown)


app/controllers/api/v1/files_controller.rb
```diff
@@ -33,15 +35,27 @@ def file_not_found
     }, status: 404
   end
 
-  def user_id
-    params.require(:user_id)
+  def verify_authentication_token
+    return unauthorized unless authenticate_with_http_token { |token, _options| token == ENV["FILES_ENDPOINT_TOKEN"] }
+  end
+
+  def station_id
```
mdbenjam: Should we do any validation on this stuff?

app/controllers/api/v1/documents_controller.rb
```diff
@@ -16,6 +18,10 @@ def show
 
   private
 
+  def authorize
+    unauthorized unless current_user.can? "Reader"
```
shanear: unfortunately `can?` is different in caseflow-efolder and caseflow. This will cause some discrepancies:

caseflow:
```
def can?(thing)
    return true if admin? && admin_roles.include?(thing)
    roles.include? thing
  end
```

caseflow-efolder:
```
def can?(thing)
    return false if roles.nil?
    roles.include? thing
end
```

config/environments/test.rb
```diff
@@ -48,4 +48,6 @@
 
   ENV["SIDEKIQ_USERNAME"] ||= "caseflow"
   ENV["SIDEKIQ_PASSWORD"] ||= "caseflow"
+
+  ENV["EFOLDER_API_KEY"] ||= "token"
```
shanear: `Rails.env.api_key?` might be more idiomatic than just going straight for the environment variable

spec/requests/api/v1/documents_spec.rb
```diff
@@ -1,5 +1,9 @@
 describe "Documents API v1", type: :request do
   context "Document by document ID" do
+    let!(:current_user) do
```
shanear: These are some very good looking tests. I am so proud 💯

app/controllers/api/v1/files_controller.rb
```diff
@@ -1,5 +1,7 @@
 class Api::V1::FilesController < Api::V1::ApplicationController
-  def show
+  before_action :verify_authentication_token
+
+  def index
```
shanear: we are only calling all of these endpoints from the caseflow back end right?

app/controllers/api/v1/files_controller.rb
```diff
@@ -33,15 +35,27 @@ def file_not_found
     }, status: 404
   end
 
-  def user_id
-    params.require(:user_id)
+  def verify_authentication_token
```
aroltsch: Can we combine two authentication mechanisms. First check if User.from_session exists and if it doesn't exist, check authentication_token.

config/environments/test.rb
mdbenjam: I don't understand what you mean here. How can I use Rails.env.api_key? here? I guess I could use it when I test for it.

shanear: Sorry I mistyped 😬, I meant `Rails.config.api_key`

like we are doing here: https://github.com/department-of-veterans-affairs/caseflow/blob/master/config/environments/production.rb#L79.  We use the `Rails.config.s3_bucket_name` instead of referencing the ENV var directly

app/controllers/api/v1/application_controller.rb
```diff
@@ -20,8 +20,30 @@ def unauthorized
...
+    @authenticate_with_token ||= authenticate_with_http_token do
+      |token, _options| token == Rails.application.config.api_key
+    end
+  end
+
+  def user_has_role
+     current_user && (current_user.can?("Reader") || current_user.can?("System Admin"))
+  end
+
+  def authorize
```
mdbenjam: I moved the authentication to the application controller to check for the token, and then if the user has the role.

app/controllers/api/v1/application_controller.rb
```diff
@@ -20,8 +20,30 @@ def unauthorized
...
-  def verify_reader_api_enabled
-    # TODO: scope this to a current user
-    unauthorized unless FeatureToggle.enabled?(:reader_api)
+  def authenticate_with_token
+    @authenticate_with_token ||= authenticate_with_http_token do
+      |token, _options| token == Rails.application.config.api_key
+    end
+  end
+
+  def user_has_role
```
mdbenjam: Hopefully this avoids the discrepancy between caseflow and efolder can?

config/environments/demo.rb
```diff
@@ -81,4 +81,6 @@
   config.s3_enabled = ENV["AWS_BUCKET_NAME"].present?
   config.s3_bucket_name = ENV["AWS_BUCKET_NAME"]
 
+  config.api_key = "token"
```
mdbenjam: Should demo get this from an env var as well?

spec/features/help_spec.rb
```diff
@@ -3,7 +3,15 @@
     visit("/")
     click_on "first last"
     expect(page).to have_content("Help")
-    click_on "Help"
+
+    # rubocop:disable Lint/HandleExceptions
+    begin
```
mdbenjam: This test started failing on master. Fixing it here, by catching an irrelevant js error.

app/controllers/api/v1/application_controller.rb
```diff
@@ -20,8 +20,30 @@ def unauthorized
...
+     current_user && (current_user.can?("Reader") || current_user.can?("System Admin"))
+  end
+
+  def authorize
+    return unauthorized unless authenticate_with_token || user_has_role
```
shanear: I'd avoid using `unless` with more than 1 expression. It's pretty tough to read. I also think the logic is wrong.

It should be:
`return unauthorized if !authenticate_with_token || !user_has_role`

which isn't the same as:
`return unauthorized if !(authenticate_with_token || user_has_role)`

I'm consistently surprised how useful DeMorgan's Theorum is (https://en.wikipedia.org/wiki/De_Morgan%27s_laws). Probably the most useful thing i learned in college lol.

app/controllers/api/v1/application_controller.rb
shanear: why not just copy the method outright? It's still not the same, because it doesn't have the whole "Admin Roles" thing.

config/environments/demo.rb
shanear: I would say no. It's not worth it.

spec/features/help_spec.rb
shanear: Woah, getting some serious PhantomJS flash backs right now lol

app/controllers/api/v1/application_controller.rb
mdbenjam: I agree it's confusing. I hate unless... Though I'm a big fan of De Morgan's law. That said, I think it is the second one
```
return unauthorized if !(authenticate_with_token || user_has_role)
```
 Per Anya's suggestion we want to allow authentication with either the token or the role.

shanear: but the username, station_id aren't secure. I could spoof any Caseflow user very easily.

app/controllers/api/v1/application_controller.rb
mdbenjam: Does it need to be the same though? It will require a bigger change. And I'm not sure that we need to block people with System Admin from accessing this endpoint.

shanear: Good point. Nope they don't need to be the same, you can be more loose in the API. 👍

app/controllers/api/v1/application_controller.rb
mdbenjam: You either have to pass in the token, or we have to be able to get the user from the session. The session is secure right? You can't spoof a user on the UI right?

app/controllers/api/v1/application_controller.rb
```diff
@@ -20,8 +20,32 @@ def unauthorized
...
+    end
+  end
+
+  def user_has_role
+    current_user && (current_user.can?("Reader") || current_user.can?("System Admin"))
+  end
+
+  def authorize
+    # rubocop:disable Style/NegatedIf
+    return unauthorized if !(authenticate_with_token || user_has_role)
```
aroltsch: You are naming this function `authorize`, yet you are doing authentication here. I think it would be better to take `authenticate_with_token` out of this function.

app/controllers/api/v1/application_controller.rb
```diff
@@ -20,8 +20,32 @@ def unauthorized
...
+    # rubocop:enable Style/NegatedIf
```
aroltsch: If we agree to use negative ifs in the code, we should update `.rubocop.yml`

app/controllers/api/v1/files_controller.rb
```diff
@@ -33,15 +33,11 @@ def file_not_found
     }, status: 404
   end
 
-  def user_id
-    params.require(:user_id)
-  end
-
   def id
-    params[:id]
+    request.headers["HTTP_VETERAN_ID"]
```
aroltsch: should we call this `HTTP_FILE_NUMBER` to be more clear?

aroltsch: I like the idea of moving this out of the URL query string

app/controllers/api/v1/application_controller.rb
shanear: Your not getting the user info from the session. It looks like it's being passed via headers:
https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/539/files#diff-c8e638129123d4606cfda05f4df8bf8bR43

shanear: The session is only in play if you are making a request from the browser to the server. But it seems like this api is only being used by the caseflow backend

Issue #523, opened 2017-07-06
Keep track of file sizes in the documents table


aroltsch:
AC:
One way to check is to:
- SSH into a uat/preprod eFolder server
- access the Ruby console
- Search for a Document using `Document.where` or `Document.last` or even `Document.all.sample` commands
- Verify that there is `size` attribute for that document and that it has a value

The value can be `nil`

astewarttistatech:
**PASSED**

Env: Preprod and UAT

Steps to Recreate:

1.SSH into a uat/preprod eFolder server
2.Access the Ruby console
3.Search for a Document using Document.where or Document.last or even Document.all.sample commands
4.Verify that there is size attribute for that document and that it has a value

Screenshots: 
<img width="565" alt="screen shot 2017-07-24 at 1 42 09 pm" src="https://user-images.githubusercontent.com/23038505/28536752-aa0887dc-7076-11e7-859b-ce53b1328787.png">
<img width="570" alt="screen shot 2017-07-24 at 1 45 56 pm" src="https://user-images.githubusercontent.com/23038505/28536753-aa0b5250-7076-11e7-8f93-e23fdaf38530.png">

PR #541, opened 2017-07-20
Anya/add size to documents

aroltsch:
connects #523 

Add file size to the documents table to track file size for VVA documents.


aroltsch:
Alan suggested it would be a good idea to learn how big VVA files are.

Issue #517, opened 2017-07-03
Coach mark behavior


lakohl:
As a caseworker, I would like to be able to see the coach marks once, and then have the ability to never see them again.

## AC
Rules: 

### First and second pass
1. Display the coach marks when the user lands on the Search page.
1. Allow the user to proceed/perform their work while the coach marks are displayed.
1. Allow the user to close the coach marks by clicking on the X.
1. Display 'Hide tutorial' link at the very top of the page (17px, color: 0071bc, underlined)
1. On click, 'Hide tutorial' eliminates the coach marks on all pages, and changes the 'Hide tutorial' link copy at the top of the page to read 'See what's new!'.
1. On click, the 'See what's new!' link switches to 'Hide Tutorial' and coach marks are displayed throughout the application until the user 'Hides', 'Skips', or clicks 'x' to stop them from displaying.

### Third Pass
1. Coach marks do not display.
1. Display 'See what's new!' link (underlined) at the top of the page.
1. On click, the 'See what's new!' link switches to 'Hide Tutorial' and coach marks are displayed throughout the application until the user 'Hides', 'Skips', or clicks 'x' to stop them from displaying.

## Mocks

First and Second pass: 
<img width="514" alt="screen shot 2017-07-03 at 1 01 51 pm" src="https://user-images.githubusercontent.com/4975959/27802121-d52e3ca0-5fef-11e7-950a-c1414b179019.png">

Third pass:
<img width="512" alt="screen shot 2017-07-03 at 1 01 31 pm" src="https://user-images.githubusercontent.com/4975959/27802133-e4124860-5fef-11e7-89ca-bc7b5dcbab44.png">

mmlumba:
"got it? skip tutorial" is 17px

lakohl:
See what's new! / Hide tutorial link is 17px

lakohl:
My presumption is that one pass = 1 Vet ID successfully retrieved

NickHeiner:
And clicking "hide tutorial" does not advance you to the next pass, right?

lakohl:
No, 'Hide Tutorial' gets rid of the coach mark(s) on that page and on subsequent pages. It's the equivalent of clicking the 'X' on that and all subsequent coach marks. It does not advance the user, they stay on that page.

NickHeiner:
@lakohl let's chat in person about the "second pass" option on Monday. I have some questions that I think are best answered synchronously.

[...]

PR #544, opened 2017-07-21
Handle VVA Coachmarks across multiple user visits

NickHeiner:
Connect #517.

This does not do the "second pass" section of the ticket, but it does some of the first and third. I will talk to @lakohl about the second pass section.

@aroltsch please re-assign as necessary.

### DIFF
--- a/app/controllers/downloads_controller.rb
+++ b/app/controllers/downloads_controller.rb
@@ -83,6 +83,8 @@ class DownloadsController < ApplicationController
 
     current_user.vva_coachmarks_view_count += 1
     current_user.save!
+    
+    binding.pry
 
     render text: ''
   end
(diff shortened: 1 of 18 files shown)


NickHeiner:
The tests are not passing yet. I'll fix that on Monday. But I don't think the overall shape of the code should change, so it should be ready for review now.

app/controllers/downloads_controller.rb
```diff
@@ -73,6 +73,20 @@ def delete
     redirect_to "/"
   end
 
+  helper_method :should_show_vva_coachmarks?
+  def should_show_vva_coachmarks?
```
aroltsch: I think the convention is to place `helper_method` below the method itself.

app/controllers/downloads_controller.rb
```diff
@@ -73,6 +73,20 @@ def delete
...
+    current_user.vva_coachmarks_view_count < 3
+  end
+
+  def increment_vva_coachmarks_status
```
aroltsch: Would it be possible to add a test for this?

Issue #536, opened 2017-07-18
CORS | Enable CORS on eFolder Express to allow the caseflow frontend to hit the document endpoint


mdbenjam:
In order to retrieve documents from eFolder Express via the frontend we need to enable CORS.

# AC
1. Verify we can make requests to the document eFolder endpoint from the Caseflow app.

kierachell:
**PASSED**

<img width="1418" alt="screen shot 2017-07-26 at 11 22 04 am" src="https://user-images.githubusercontent.com/18075411/28631545-97923766-71fb-11e7-914a-8f647e38614a.png">

PR #545, opened 2017-07-24
Enable CORS

mdbenjam:
Connects: #536 

Since caseflow and efolder are under different subdomains we need to enable CORS. We also add real documents to eFolder so that when we dev locally against the API it will return usable material. Finally, we address a recently introduced bug that causes StaleObjectErrors.


app/controllers/api/v1/application_controller.rb
```diff
@@ -38,7 +38,7 @@ def authenticate_or_authorize
     # passed in the header. Otherwise we try to create the current user from the session
     # and authorize based on the presence of the Reader role.
     if authenticate_with_token
-      @current_user = User.find_or_create_by(css_id: css_id, station_id: station_id) if authenticate_with_token
+      @current_user = User.find_or_create_by(css_id: css_id, station_id: station_id)
```
mdbenjam: removed duplicate if

app/jobs/download_manifest_job.rb
```diff
@@ -18,6 +18,7 @@ def perform(download)
       external_documents: external_documents
     )
     download_documents.create_documents
+    download.reload
```
mdbenjam: Now that we are updating existing records, we sometimes get stale object errors when we try to modify the download record. This is because we're accessing it in another job. So we reload before proceeding.

lib/fakes/document_service.rb
```diff
@@ -5,10 +5,11 @@ def self.fetch_documents_for(_download)
     []
   end
 
-  def self.fetch_document_file(_document)
+  def self.fetch_document_file(document)
```
mdbenjam: Return real pdf documents instead of txt documents.

app/jobs/fakes/download_manifest_job.rb
```diff
@@ -55,13 +55,13 @@ def perform(download)
 
     check_and_raise_errors(demo)
 
-    download.update_attributes!(status: :pending_confirmation)
+    download.reload.update_attributes!(status: :pending_confirmation)
   rescue VBMS::ClientError => e
-    capture_error(e, download, :vbms_connection_error)
+    capture_error(e, download.reload, :vbms_connection_error)
```
joofsh: do we need all these `reload`s? We just reloaded on line 58

config/application.rb
```diff
@@ -28,6 +28,13 @@ class Application < Rails::Application
     config.autoload_paths += Dir[Rails.root + 'app/jobs']
     config.autoload_paths << Rails.root.join('lib')
 
+    config.middleware.insert_before 0, "Rack::Cors" do
+      allow do
+        origins (ENV["CORS_URL"] || "http://localhost:3000")
+        resource '*', headers: :any, methods: [:get, :post, :options], credentials: true
```
joofsh: is it possible to do `methods: :any` as well? Or if you need to be explicit, should we include patch/put/head?

app/controllers/api/v1/application_controller.rb
joofsh: haha, good catch 👍

app/jobs/fakes/download_manifest_job.rb
mdbenjam: I think we do, because the rescues can break the flow meaning we may not hit line 58.

config/application.rb
mdbenjam: Good catch--I just copied the example and forgot to change this. We only need get now, but any will let us add to this easily.

app/jobs/fakes/download_manifest_job.rb
joofsh: fair point. This feels a little hacky but I don't have a better suggestion. Onward 👍

Issue #548, opened 2017-07-24
Re-enable Google Analytics


NickHeiner:
We first added Google Analytics in #216, but it no longer works. 

## Acceptance Criteria
For GA account `UA-74789258-2`, metrics show up.

kierachell:
@NickHeiner 
- I don't have/know access to our GA account
- Also, I don't know if UAT/Preprod servers send the events to GA
Shall I get @shanear to validate this?

NickHeiner:
@kierachell you should definitely have access to our GA account. What email should I use? I can add you.

Yes, with this change, UAT and preprod will send events to GA. When we query in GA, we'll set up a filter to only view prod traffic.

kierachell:
**PASSED**
@shanear added me to the account:
<img width="970" alt="screen shot 2017-07-25 at 7 23 10 pm" src="https://user-images.githubusercontent.com/18075411/28597961-c02b7274-716e-11e7-8878-763d946e5d2d.png">

PR #549, opened 2017-07-24
Re-enable GA

NickHeiner:
Connect #548.

To test, log into GA and see metrics showing up. For instance:

image

Our past implementation tried to avoid sending data when running not in production. I suspect that additional complexity is why we accidentally were not sending data for anything, including production. It's dangerous when developers cannot verify the correct behavior locally.

Instead, we log all data. In the GA interface, we can use filtering tools to limit queries to only be the prod domain.

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -47,16 +47,14 @@
 
 -->
 
-  <% if !Rails.configuration.analytics_account.nil? %>
-      <script>
-        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
-              (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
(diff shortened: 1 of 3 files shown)


app/views/layouts/application.html.erb
```diff
@@ -47,16 +47,14 @@
...
-        ga('send', 'pageview', '<%= current_ga_path %>');
-      </script>
-  <% end %>
+  <script>
+    (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+          (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+    })(window,document,'script','//www.google-analytics.com/analytics.js','ga');
+    ga('create', '<%= Rails.configuration.analytics_account %>', 'auto');
+    ga('send', 'pageview', '<%= current_ga_path %>');
```
mogrenAtWork: Is this going to push events from other events and mix them in with prod events?

PR #551, opened 2017-07-25
Fix Files Endpoint

mdbenjam:
Connects: #505 
Artem brought up some good points in validation of the files endpoint. Addressing them here:

1. vbms_document_id should be external_document_id since we also pull docs from VVA.
1. We shouldn't be throwing 404's when we don't retrieve any documents from VBMS or VVA. An empty eFolder is valid.
1. We should handle VBMS and VVA errors properly, by sending that they error-ed in the response.


app/controllers/api/v1/files_controller.rb
```diff
@@ -1,8 +1,6 @@
 class Api::V1::FilesController < Api::V1::ApplicationController
   def index
     render json: json_files
-  rescue ActiveRecord::RecordNotFound
```
mdbenjam: There should be no case where we return a 404. We create a record if we don't already have one.

spec/requests/api/v1/file_spec.rb
```diff
@@ -50,10 +50,77 @@
...
+    end
+
+    context "vva throws a client error" do
+      let(:vva_error) { true }
+
+      before do
+        allow(VVAService).to receive(:fetch_documents_for).and_raise(VVA::ClientError)
+      end
+
+      it "returns existing files, a nil manifest_fetched_at, and vbms_error is true" do
```
shanear: this should be `vva_error`?

mdbenjam: Yeah I actually just updated the tests, and changed this.

Issue #518, opened 2017-07-03
Help | FAQs | Update question/answer with VVA docs available in download


lakohl:
As a Product Owner, I would like the FAQs to reflect the new change of having VVA docs included in the download of the veteran's efolder so that the user knows about this new feature.

## Acceptance Criteria
1. Verify in the "Frequently Asked Questions" section, it contains the newly updated listing of all questions. (Refer to [eFolder Express FAQs](https://github.com/department-of-veterans-affairs/appeals-training/blob/master/eFolder-Express/Rollout-Package/In-App-Help/eFolder.Express_Frequently.Asked.Questions_Sep2017.pdf)) **_Note:  Several questions have changed as well as order._**
1. Verify that in the "FAQ Answers" section, it contains all newly updated questions and their answer.  (Refer to [eFolder Express  FAQs](https://github.com/department-of-veterans-affairs/appeals-training/blob/master/eFolder-Express/Rollout-Package/In-App-Help/eFolder.Express_Frequently.Asked.Questions_Sep2017.pdf))  **_Note:  Several questions have changed as well as order._**
1. Verify that each question in the FAQ section is a hyperlink to its answer. 

## Mocks
**_Current Help Page - update FAQs section_**
![faq_mock](https://user-images.githubusercontent.com/23361723/28533789-6b6f7f44-706c-11e7-9223-330a82f09d96.png)

# Related Story
eFolder Express - VVA | FAQs | Update/Develop questions [#20](https://github.com/department-of-veterans-affairs/appeals-training/issues/20)

astewarttistatech:
**PASSED**

Env: Demo,UAT
Browsers: FF and Chrome 

Steps to Validate: 
1. Access the /help page for eFolder Express
2. Validate that the FAQ was added under the training videos.

Screenshots: 
![efolder expresshelp](https://user-images.githubusercontent.com/23038505/28777580-c235c812-75c9-11e7-847a-1d27e59d5ea1.png)

wturne:
Updated link above with most recent version of eX FAQs.  

[eFolder.Express_Frequently.Asked.Questions_Sep2017.pdf](https://github.com/department-of-veterans-affairs/caseflow-efolder/files/1289967/eFolder.Express_Frequently.Asked.Questions_Sep2017.pdf)

wturne:
Issue reopened by mistake.  Closing issue.

PR #552, opened 2017-07-25
Update Help Page FAQs

Sjones352:
connects #518 
Because most of the content changed in FAQs plus another question was added, I decided to use partials and then rendered them in the help/show file`<% if vva_feature_enabled? %>`

screen shot 2017-07-25 at 5 00 12 pm


app/views/help/show.html.erb
```diff
@@ -55,149 +55,11 @@
...
-  </p>
-
-  <h2 id="need-help">18. What if I still need help?</h2>
-
-  <p>
-     If you still encounter issues or have questions that aren't answered here, reach out to the Caseflow Product Support Team by calling 1-844-876-5548.
-  </p>
+   <% if vva_feature_enabled? %>
+    <%= render "faqs_with_vva" %>
+    <% else %>
```
aroltsch: Indentation is off. `Els`e should be lined up with `if`
https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/views/downloads/_progress.html.erb#L28

PR #558, opened 2017-07-31
Fixing stale document error

mdbenjam:
Trying to stop stale document errors from happening in prod. It's difficult to test locally since we think it's a race condition, and since the fakes demo download manifest job takes a very different path than the real one.


app/services/download_documents.rb
```diff
@@ -54,7 +54,7 @@ def create_documents
         end
       end
 
-      @download.reload.update_attributes!(manifest_fetched_at: Time.zone.now)
+      @download.reload(lock: true).update_attributes!(manifest_fetched_at: Time.zone.now)
```
mogrenAtWork: If you use `lock!` it will reload the object for you

PR #559, opened 2017-07-31
Defaulting CORS config to empty string to allow rake tasks to execute

mogrenAtWork:


config/application.rb
```diff
@@ -30,11 +30,11 @@ class Application < Rails::Application
     # Currently the Caseflow client makes calls to get document content directly
     # from eFolder Express to reduce load on Caseflow. Since Caseflow and eFolder
     # are in different sub-domains, we need to enable CORS.
-    cors_origins = ENV["CORS_URL"]
+    cors_origins = ENV["CORS_URL"] || "" # Default to empty string to allow rake jobs to execute
 
     # Enable localhost CORS for development and test environments and get rid of null values
     # if the environment variable isn't set
-    cors_origins ||= "http://localhost:3000" unless Rails.env.production?
+    cors_origins = "http://localhost:3000" unless Rails.env.production? || !cors_origins.empty?
```
mogrenAtWork: I had to change the logic here since we're defaulting to empty string. If `cors_origins` is an empty string `||=` will not set the value to `localhost`

joofsh: `unless` with multiple conditions makes my head spin. Do you mind changing it to an `if`?

mogrenAtWork:
@kierachell fixed the problem with:
https://github.com/department-of-veterans-affairs/appeals-deployment/issues/725
https://github.com/department-of-veterans-affairs/appeals-deployment/pull/726

Closing this out

PR #561, opened 2017-07-31
Stale Object: Take 2

mdbenjam:
The real cause was updating the veteran info from BGS after the manifest download job had started.


app/jobs/fakes/download_manifest_job.rb
```diff
@@ -55,13 +55,13 @@ def perform(download)
 
     check_and_raise_errors(demo)
 
-    download.reload.update_attributes!(status: :pending_confirmation)
```
mdbenjam: These were unneeded when we attacked the real problem.

spec/requests/api/v1/file_spec.rb
```diff
@@ -182,10 +182,10 @@
               vva_error: false,
               documents: [
                 {
-                  id: download.documents[0].id,
-                  type_id: "124",
-                  received_at: "2017-04-03T00:00:00.000Z",
-                  external_document_id: "2"
+                  id: download.documents[2].id,
```
mdbenjam: For some reason this changed the order of these... We probably should sort on something to make this stable. But for now the order doesn't matter.

PR #567, opened 2017-08-01
API Error Cases

mdbenjam:
Connects: #535

We currently don't handle error cases properly. If a user doesn't pass the correct headers, we get a sentry error. This change updates the API to return 400's if any headers are missing.

### DIFF
--- a/spec/requests/api/v1/file_spec.rb
+++ b/spec/requests/api/v1/file_spec.rb
@@ -133,6 +133,56 @@ describe "File API v1", type: :request do
     end
   end
 
+  context "When headers are missing" do
+    context "missing CSS ID" do
+      let(:headers) do
+        {
+          "HTTP_FILE_NUMBER" => veteran_id,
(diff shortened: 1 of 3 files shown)


spec/requests/api/v1/file_spec.rb
```diff
@@ -133,6 +133,56 @@
...
+          "HTTP_STATION_ID" => user.station_id,
+          "HTTP_CSS_ID" => user.css_id,
+          "HTTP_AUTHORIZATION" => "Token token=#{token}"
+        }
+      end
+
+      it "returns 400" do
+        get "/api/v1/files", nil, headers
+        expect(response.code).to eq("400")
+        expect(response.body).to eq("{\"status\":\"missing header: File Number\"}")
```
aroltsch: you could also parse the JSON, it might be cleaner and it is better to use regex in case error message changes a bit
```
body = JSON.parse(response.body)
expect(body[:status]).to match /missing.+File.+Number/
```

Issue #565, opened 2017-08-01
Make all eX buttons comply with Caseflow Commons styling


lakohl:
Ensure that all eX buttons are sentence case and follow the measurements set out in Caseflow Commons.

## AC
1. Verify that all eX buttons are sentence case.
1. Verify that the height of all the eX buttons (including the new copy Vet ID button) follows Caseflow Commons.
1. Verify that the styling of the padding to the left and right of the button label is in accordance with Caseflow Commons.

Sjones352:
Currently

![buttons](https://user-images.githubusercontent.com/23080951/28954990-ee4a5492-78b0-11e7-85e1-d2c7bba70f38.gif)

lakohl:
nice!

astewarttistatech:
@Sjones352 Which buttons are "eX buttons"?

lakohl:
@astewarttistatech all the buttons in the eX application should follow Caseflow Commons guidelines (they were too big).

astewarttistatech:
**PASSED**

Env: UAT
Browsers: FF and Chrome 

Steps to Validate: 
1. Access eFolder Express
2. Check that all buttons meet the 10x20 padding exception

Screenshots:
<img width="1248" alt="screen shot 2017-08-07 at 11 18 17 am" src="https://user-images.githubusercontent.com/23038505/29033184-31c9043a-7b62-11e7-961c-ae79e2e73a88.png">

PR #571, opened 2017-08-03
eX buttons comply with Caseflow Commons styling

Sjones352:
Connects to https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/565

## AC
1. Verify that all eX buttons are sentence case.
1. Verify that the height of all the eX buttons (including the new copy Vet ID button) follows Caseflow Commons.
1. Verify that the styling of the padding to the left and right of the button label is in accordance with Caseflow Commons.

screen shot 2017-08-03 at 12 07 19 pm

buttons

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,4 +1,5 @@
 source ENV['GEM_SERVER_URL'] || 'https://rubygems.org'
+
 gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", ref: "6e8bed78dcb4e6699f55257ad90d016b71639ec0"
 
 gem "moment_timezone-rails"
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,4 @@
 source ENV['GEM_SERVER_URL'] || 'https://rubygems.org'
-
 gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", ref: "6e8bed78dcb4e6699f55257ad90d016b71639ec0"
 
 gem "moment_timezone-rails"
--- a/app/assets/stylesheets/_main.scss
+++ b/app/assets/stylesheets/_main.scss
@@ -217,15 +217,3 @@ table td {
   padding-right: 20px;
   padding-left: 20px;
 }
-
-.cf-efolder-app,
-  .cf-app-width {
-  margin-left: auto;
-  margin-right: auto;
-  width: 75vw;
-}
-
-
-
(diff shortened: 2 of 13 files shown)


app/assets/stylesheets/_main.scss
```diff
@@ -208,3 +208,12 @@ table td {
   color: $color-gray;
 }
 
+.cf-retrive-button.cf-submit {
```
mmlumba: spelling nit: "retrieve"

app/assets/stylesheets/_main.scss
```diff
@@ -208,3 +208,12 @@ table td {
...
+  padding-right: 20px;
+  padding-left: 20px;
+}
+
+.ee-copy-button {
```
mmlumba: You can change this and the other class to the following:
`padding: 0 20px`
https://developer.mozilla.org/en-US/docs/Web/CSS/padding

mmlumba:
@Sjones352 can you link to the original ticket?

test
```diff
@@ -0,0 +1 @@
+plain text
```
mmlumba: This is a test file that runs after every test.  There's a pending ticket that will resolve to clear this file after every test run but for now delete it.

app/views/downloads/_confirm.html.erb
```diff
@@ -90,11 +89,15 @@
 </div>
 
 <div class="cf-app-segment">
+<div class="cf-push-right">
```
mmlumba: The bottom part still looks like this (might be out of scope for this ticket):
screen shot 2017-08-03 at 2 14 29 pm

Gemfile
```diff
@@ -1,5 +1,4 @@
 source ENV['GEM_SERVER_URL'] || 'https://rubygems.org'
-
```
Sjones352: Fixed

app/assets/stylesheets/_main.scss
Sjones352: Fixed

app/assets/stylesheets/_main.scss
Sjones352: Fixed

test
Sjones352: Fixed

app/views/downloads/_confirm.html.erb
Sjones352: Fixed

PR #575, opened 2017-08-03
User.id bug

mdbenjam:
Connects: #2873

For some reason this bug seems to manifest itself on UAT but not locally. It seems on UAT rails correctly converts user -> user.id when it tries to find the record. But if that record is not found it appears to convert user -> nil when creating the record. I cannot seem to reproduce this locally, but it is tested on UAT and appears to fix the problem.

### DIFF
--- a/app/controllers/api/v1/files_controller.rb
+++ b/app/controllers/api/v1/files_controller.rb
@@ -27,6 +27,6 @@ class Api::V1::FilesController < Api::V1::ApplicationController
   end
 
   def download
-    @download ||= Download.find_or_create_by_user_and_file(current_user, id)
+    @download ||= Download.find_or_create_by_user_and_file(current_user.id, id)
   end
 end
--- a/app/models/serializers/v1/download_serializer.rb
+++ b/app/models/serializers/v1/download_serializer.rb
@@ -12,7 +12,7 @@ class Serializers::V1::DownloadSerializer < ActiveModel::Serializer
   end
 
   attribute :documents do
-    object.documents.map do |document|
+    object.documents.order(:id).map do |document|
       {
         id: document.id,
         type_id: document.type_id,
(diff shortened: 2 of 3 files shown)


app/models/serializers/v1/download_serializer.rb
```diff
@@ -12,7 +12,7 @@ class Serializers::V1::DownloadSerializer < ActiveModel::Serializer
   end
 
   attribute :documents do
-    object.documents.map do |document|
+    object.documents.order(:id).map do |document|
```
mdbenjam: This way our API returns things in a consistent order.

spec/requests/api/v1/file_spec.rb
```diff
@@ -264,6 +264,23 @@
         expect(response.code).to eq("200")
         expect(response.body).to eq(response_body)
       end
+
+      it "doesn't download files added within 3 hours of calling the endpoint" do
```
joofsh: I don't see how this test verifies the above statement. Don't we want to specifically test that `VBMSService.fetch_documents_for` is never called?

app/controllers/api/v1/files_controller.rb
```diff
@@ -27,6 +27,6 @@ def id
   end
 
   def download
-    @download ||= Download.find_or_create_by_user_and_file(current_user, id)
+    @download ||= Download.find_or_create_by_user_and_file(current_user.id, id)
```
joofsh: yay! less rails magic

spec/requests/api/v1/file_spec.rb
mdbenjam: We add a document, but then make sure we verify against the original list. I can add some comments--or do you have a different suggestion about how to test this?

joofsh: A more direct way to test this would be to test that the method does not get called. `expect(VBMService).to receive(: fetch_documents_for).exactly(0).times`

PR #578, opened 2017-08-04
Clean up StaleObjectErrors

mdbenjam:
connects: https://github.com/department-of-veterans-affairs/caseflow/issues/2864

We were getting stale object errors because of a race condition. And we are timing document downloads incorrectly. We should not include S3 in the document download time.

**Test Plan**
- [ ] Verify in a test that we do not write to the DB when the document is fetched from S3.
- [ ] Run two curl commands to the documents endpoint and enable concurrent requests. Ensure we stop receiving StaleObjectErrors.


app/models/fetcher.rb
```diff
@@ -2,22 +2,28 @@ class Fetcher
...
   end
 
   private
 
-  def cache_in_s3
+  def cached_content
+    @cached_content ||= S3Service.fetch_content(document.s3_filename)
+  end
+
+  def cache_in_s3(save_document_metadata: true)
```
shanear: this method name is sort of confusing. It doesn't mention that the document is being downloaded from vbms, which is arguably the most important fact of the method. Caching is secondary.

Additionally, I feel like its probably better programming practice to just have two methods instead of 1 method with a flag that used for more than one logic path.

```
def download_from_service
  external_service.fetch_document_file(document).tap do |result|
    S3Service.store_file(document.s3_filename, result)
  end
end

def download_from_service_and_record
  document.update_attributes!(started_at: Time.zone.now)
  download_from_service.tap do |result|
    document.update_attributes!(
      completed_at: Time.zone.now,
      download_status: :success,
      size: result.try(:bytesize)
    )
  end
end
```

app/models/fetcher.rb
```diff
@@ -2,22 +2,35 @@ class Fetcher
...
-    result = S3Service.fetch_content(document.s3_filename) || cache_in_s3
-    document.update_attributes!(
-      completed_at: Time.zone.now,
-      download_status: :success,
-      size: result.try(:bytesize)
-    )
-    result
+  def content(save_document_metadata: true)
+    return cached_content if cached_content
+    if save_document_metadata
```
mdbenjam: Do you think this conditional is okay?

spec/models/fetcher_spec.rb
```diff
@@ -39,6 +41,15 @@
       it "should update document size" do
         subject
         expect(document.reload).to_not eq nil
+        expect(document.started_at).to_not eq nil
```
shanear: this test name doesn't make sense

Issue #583, opened 2017-08-07
Remove code to look for old S3 name


mdbenjam:
About 2 months ago we changed the file names of documents in S3. We still looked for the old name if we couldn't find the new name so that we wouldn't break functionality when we rolled out. Since the S3 bucket expires documents after 2 weeks, and this change was more than 2 months ago, the old named documents are expired, and we are able to remove this code.

## AC (for dev to check):
- Verify all references to the `old_s3_filename` are gone

## AC (for QA to check):
- Verify documents still get fetched from S3

nickheiner-usds:
> Now the old named documents should be expired, and we should be able to remove this code.

Is there a way that we can get a more rigorous level of certainty than "should" here? :smile: It would be nice to ensure that we're not going to cause any errors.

mdbenjam:
Updated to explain my reasoning.

kierachell:
**PASSED**

- Checked S3 to see if any PDFs remained with old names:
<img width="522" alt="screen shot 2017-10-20 at 9 45 39 am" src="https://user-images.githubusercontent.com/18075411/31823885-8bf6843e-b57b-11e7-86df-03d24d5f898d.png">
<img width="685" alt="screen shot 2017-10-20 at 9 45 43 am" src="https://user-images.githubusercontent.com/18075411/31823886-8c1ab854-b57b-11e7-915b-2d599d8ed587.png">
<img width="681" alt="screen shot 2017-10-20 at 9 45 46 am" src="https://user-images.githubusercontent.com/18075411/31823887-8c38152a-b57b-11e7-82cb-1ac53e9a5c61.png">

Issue #582, opened 2017-08-07
S3 file name should be the external file id


mdbenjam:
Currently we name files in S3 based on the download and local document ids. We should instead name them based on the external id. This way if multiple users try to download a case, the files only need to be added to S3 once.

# AC
1. Verify the S3 filename is the same as the id coming from VBMS or VVA.

kierachell:
**PASSED**
<img width="924" alt="screen shot 2017-08-10 at 8 19 33 pm" src="https://user-images.githubusercontent.com/18075411/29197425-43ad87d0-7e09-11e7-9f56-ff9b67d7b9ac.png">

PR #584, opened 2017-08-07
Change S3 name

mdbenjam:
Connects: #582

We will remove the old name code in #583

### DIFF
--- a/app/models/fetcher.rb
+++ b/app/models/fetcher.rb
@@ -14,7 +14,8 @@ class Fetcher
   private
 
   def cached_content
-    @cached_content ||= S3Service.fetch_content(document.s3_filename)
+    @cached_content ||= (S3Service.fetch_content(document.s3_filename) ||
+      S3Service.fetch_content(document.old_s3_filename))
   end
 
   def download_from_service
--- a/app/models/fetcher.rb
+++ b/app/models/fetcher.rb
@@ -3,7 +3,6 @@ class Fetcher
   attr_accessor :document, :external_service
 
   def content(save_document_metadata: true)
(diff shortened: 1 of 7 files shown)


app/models/fetcher.rb
```diff
@@ -14,10 +13,12 @@ def content(save_document_metadata: true)
   private
 
   def cached_content
-    @cached_content ||= S3Service.fetch_content(document.s3_filename)
+    @cached_content ||= (S3Service.fetch_content(document.s3_filename) ||
+      S3Service.fetch_content(document.old_s3_filename))
   end
 
   def download_from_service
+    return cached_content if cached_content
```
mdbenjam: Moved inside of here since with the new s3 name you could cache this document from another download. But if we never write to the DB, we'll be missing fields like download_status.

spec/services/download_documents_spec.rb
```diff
@@ -17,14 +21,18 @@
...
   end
 
+  let(:first_document_id) { generate_id }
+  let(:second_document_id) { generate_id }
+  let(:third_document_id) { generate_id }
+
   let(:external_documents) do
     [
-      OpenStruct.new(document_id: "1", filename: "filename.pdf", doc_type: "123",
+      OpenStruct.new(document_id: first_document_id, filename: "filename.pdf", doc_type: "123",
```
mdbenjam: We are not clearing the S3 store between tests. We should probably fix that in commons, but at least we need different ids for documents.

app/models/fetcher.rb
shanear: ooh good catch.

Issue #586, opened 2017-08-07
eFolder Integration | API DB Field


mdbenjam:
We need to add a DB field to the Download and documents tables so that we know which downloads and documents are triggered from the API vs from the UI.

# AC
1. Verify there is a new `from_api` field in the downloads table.
1. Verify when a download is triggered from the UI these fields are nil.
1. Verify when a download is triggered from the API these fields are filled.
1. Verify when a download is triggered from the UI, it only updates download or document objects if they were first created with the UI.
1. Verify when a download is triggered from the UI,  it only updates download or document objects if they were first created with the API.

kierachell:
**PASSED**

```
Download.last
  Download Load (1.9ms)  SELECT  "downloads".* FROM "downloads"  ORDER BY "downloads"."id" DESC LIMIT 1
[2017-08-11 14:57:39 -0400]   Download Load (1.9ms)  SELECT  "downloads".* FROM "downloads"  ORDER BY "downloads"."id" DESC LIMIT 1
=> #<Download id: 1166, request_id: nil, file_number: "520488743", status: 1, created_at: "2017-08-11 18:56:51", updated_at: "2017-08-11 18:56:52", lock_version: 2, manifest_fetched_at: nil, started_at: "2017-08-11 18:56:52", completed_at: nil, veteran_last_name: nil, veteran_first_name: nil, veteran_last_four_ssn: nil, user_id: 8, zipfile_size: nil, from_api: true>
irb(main):002:0> Download.first
  Download Load (1.9ms)  SELECT  "downloads".* FROM "downloads"  ORDER BY "downloads"."id" ASC LIMIT 1
[2017-08-11 14:57:44 -0400]   Download Load (1.9ms)  SELECT  "downloads".* FROM "downloads"  ORDER BY "downloads"."id" ASC LIMIT 1
=> #<Download id: 1085, request_id: nil, file_number: "025456121", status: 2, created_at: "2017-08-10 17:11:46", updated_at: "2017-08-10 17:11:51", lock_version: 2, manifest_fetched_at: "2017-08-10 17:11:51", started_at: nil, completed_at: nil, veteran_last_name: nil, veteran_first_name: nil, veteran_last_four_ssn: nil, user_id: 14, zipfile_size: nil, from_api: false>
```

PR #591, opened 2017-08-09
Add new from_api flag to DB

mdbenjam:
Connects: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/586

In order to separate the effects of the new eFolder API from the UI we're adding a flag to the Download table that will mark downloads as coming from the api or not. The API should only update documents that were created via the API and the UI should only update documents that were created via the UI.

**Test Plan**
- [x] Enter DEMO1 into the UI and download it. Then curl the files endpoint with: curl "localhost:4000/api/v1/files" -H "Authorization: Token token=token" -H "css-id: CSFLOW4" -H "station-id: 283" -H "file-number: DEMO1". Ensure there are two DB entries.
- [x] Enter DEMO1 into the UI twice. Ensure only one DB entry is created.
- [x] Enter DEMO1 into the API twice. Ensure only one DB entry is created.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -217,8 +217,9 @@ class Download < ActiveRecord::Base
     end
 
     def find_or_create_by_user_and_file(user_id, file_id)
-      Download.includes(:documents).where(user_id: user_id, file_number: file_id, from_api: true).last ||
-        Download.create(user_id: user_id, file_number: file_id, from_api: true)
+      download_scope = Download.where(user_id: user_id, file_number: file_id, from_api: true)
+
+      download_scope.includes(:documents).last || download_scope.create!
     end
   end
 
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -217,8 +217,8 @@ class Download < ActiveRecord::Base
     end
 
     def find_or_create_by_user_and_file(user_id, file_id)
-      Download.includes(:documents).where(user_id: user_id, file_number: file_id).last ||
-        Download.create(user_id: user_id, file_number: file_id)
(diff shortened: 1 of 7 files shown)


app/models/download.rb
```diff
@@ -217,8 +217,8 @@ def bgs_service
     end
 
     def find_or_create_by_user_and_file(user_id, file_id)
-      Download.includes(:documents).where(user_id: user_id, file_number: file_id).last ||
-        Download.create(user_id: user_id, file_number: file_id)
+      Download.includes(:documents).where(user_id: user_id, file_number: file_id, from_api: true).last ||
```
shanear: you can create a scope and use that to make it a little crisper.

Additionally `create!` might be what you want to use, since if the download isn't saved, that should error.

```
  download_scope = Download..where(user_id: user_id, file_number: file_id, from_api: true)

  download_scope.includes(:documents).last || download_scope.create!
end
```

app/models/search.rb
```diff
@@ -55,7 +55,8 @@ def sanitized_file_number
   def download_scope
     Download.active.where(
       file_number: sanitized_file_number,
-      user: user
+      user: user,
+      from_api: [false, nil]
```
shanear: This is awkard af. we can default `from_api` to false.

db/migrate/20170809174911_add_from_api_to_downloads.rb
```diff
@@ -0,0 +1,5 @@
+class AddFromApiToDownloads < ActiveRecord::Migration
+  def change
+    add_column :downloads, :from_api, :boolean
```
shanear: you can default `from_api` to `false` at the database level here.

Issue #574, opened 2017-08-03
Table header scrolls with the doc manifest


lakohl:
As a user I would like to be able to see the column labels when I scroll through the list of files eX is retrieving, so I can better understand the content of each column. 
Currently the table headings scroll with the doc list :(

**AC** 
1. Ensure the table header contains the following column headers (listed here from left to right): Document Type, Source, Receipt Date.
1. Ensure that table header is always visible on the Start Retrieving eFolder page
1. Ensure the table header is always visible on the Retrieving eFolder page
1. Ensure the table header is always visible on the Download eFolder page

lakohl:
@Sjones352 if you are already working on fixing this, please assign this issue to yourself :)

Sjones352:
Currently. Still a few things to fix 
![header](https://user-images.githubusercontent.com/23080951/29102536-637aab58-7ca9-11e7-9060-d7320f11f258.gif)

lakohl:
Awesome 😀

lakohl:
Can the table headers also retain the grey background?

Sjones352:
@lakohl yes the table headers retain the grey background

![header](https://user-images.githubusercontent.com/23080951/29151627-9de993d8-7d50-11e7-987f-217981c22a5e.gif)

lakohl:
@Sjones352 I noticed that on the 'Start retrieving eFolder' page (the first page after the loading spinner), the far right table header still says 'Upload Date' instead of 'Receipt Date'.

I updated the AC to fix this as well.

kierachell:
### The header scrolls if you...uh... scroll the header... weird:
- Hover over the header and try scrolling
- Weird.
![header](https://user-images.githubusercontent.com/18075411/29197473-9f049b8c-7e09-11e7-80f3-7cdf6426b13c.gif)

Sjones352:
@kierachell Fixed it and assigned you the PR.

kierachell:
**PASSED**

![scroll](https://user-images.githubusercontent.com/18075411/29221508-0c9b14ea-7e8d-11e7-9f1b-2e01a2afa6e9.gif)

PR #596, opened 2017-08-11
Table header scrolls with the doc manifest (#574)

Sjones352:
Table header scrolls with the doc manifest (#574)

Removed the class that was hovering the header and replaced it with a class that leaves header fixed.


kierachell:
scroll

Issue #573, opened 2017-08-03
VVA coach marks - Extra long underline under Hide Tutorial


lakohl:
Hide Tutorial displays across the app (in IE) with an extra long underline that extends to the right all the way to the user name.

amprokop:
Can we add steps for a dev to reproduce? Not sure if we have to turn coach marks on locally or not.

lakohl:
This was observed on more tan one occasion during usability testing with users who use IE as their browser. I was not able to recreate it on my laptop, but it may be a variant of the following artifact which displays on my computer after the link has been clicked once:
<img width="508" alt="screen shot 2017-08-07 at 3 18 24 pm" src="https://user-images.githubusercontent.com/4975959/29042183-e21f8572-7b83-11e7-9e72-6e69bda8073c.png">

astewarttistatech:
**PASSED**

Env: UAT
Browsers: IE

Steps to Validate: 
1. Access eFolder with a css id that has coachmarks enabled. 
2. Validate that the underline doesn't stretch passed the "Hide Tutorial" hyperlink. 

Screenshots: 
![coachmarksie](https://user-images.githubusercontent.com/23038505/29367845-2cb25fa8-826c-11e7-9371-5a258867c514.gif)

PR #598, opened 2017-08-14
change padding to margin to fix underline

mmlumba:
Connects #573


nemodjuric:
screen shot 2017-08-15 at 12 03 15 pm
screen shot 2017-08-15 at 12 00 56 pm

nemodjuric:
@amprokop ^^

amprokop:
@mmlumba @nemodjuric — does that mean that the focus is automatically assigned to the "Hide tutorial" link? in that case, are these coachmarks 508 accessible?

mmlumba:
@amprokop Yes the coachmarks (both the bubble and the link to view them) are accessible.

amprokop:
ok cool!

Issue #613, opened 2017-09-07
Fix Veteran Name in Demo


lakohl:
I would like to see how a Veteran name render in eX.

Currently demo displays 'Test User' where the Veteran name would show. It is therefore not possible to discern the scenarios when the name appear as 'lastname, first name' e.g. on the downloaded eFolder. Please update the Veteran name throughout demo and use Joe Snuffy as the Veteran name.

lakohl:
We are doing a demo for the HI RO on Monday.

Sjones352:
@lakohl and @amprokop I can work on this one if no one has picked it up.

Sjones352:
### Screenshot
currently
<img width="1187" alt="screen shot 2017-09-13 at 12 43 26 pm" src="https://user-images.githubusercontent.com/23080951/30390280-91356ba0-9883-11e7-972b-5a52dec09544.png">

lakohl:
@Sjones352 looks great! We should also make sure the download folder is labeled with the Veteran name and last 4 in this format: 'Snuffy, Joe - 1234'

Sjones352:
### Currently 
<img width="1242" alt="screen shot 2017-09-13 at 2 48 19 pm" src="https://user-images.githubusercontent.com/23080951/30395240-8132e812-9893-11e7-8032-717891e19df5.png">

lakohl:
Thanks @Sjones352 ! Looks good to go.

kierachell:
**PASSED**

Randomized too!??!?! NICE!
<img width="953" alt="screen shot 2017-09-13 at 8 12 25 pm" src="https://user-images.githubusercontent.com/18075411/30406467-ec91b74c-98bf-11e7-9e8c-6ba5cd95c933.png">
<img width="1164" alt="screen shot 2017-09-13 at 8 12 31 pm" src="https://user-images.githubusercontent.com/18075411/30406469-ec947284-98bf-11e7-8dee-1e3778638517.png">
<img width="842" alt="screen shot 2017-09-13 at 8 12 36 pm" src="https://user-images.githubusercontent.com/18075411/30406468-ec93bace-98bf-11e7-9587-4f954aabf85b.png">

PR #620, opened 2017-09-13
Connects to #613 Fix Veteran Name in Demo

Sjones352:
Connects to #613 
Demo using Joe Snuffy and other names  as the Veteran name but not TEST USER.
screen shot 2017-09-13 at 2 48 19 pm

### Screenshot2
screen shot 2017-09-13 at 2 40 22 pm

### DIFF
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -196,6 +196,7 @@ RSpec.feature "Downloads" do
     download = @user_download.last
     expect(download).to_not be_nil
     expect(download.veteran_name).to_not be_nil
+    expect(download.veteran_name).to_not be_empty
 
     expect(page).to have_content "VETERAN ID"
   end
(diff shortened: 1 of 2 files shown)


spec/features/download_spec.rb
```diff
@@ -195,12 +195,9 @@ def assert_coachmark_does_not_exist
 
     download = @user_download.last
     expect(download).to_not be_nil
-    expect(download.veteran_name).to eq("Test User")
-    expect(download.veteran_first_name).to eq("Test")
-    expect(download.veteran_last_name).to eq("User")
-    expect(download.veteran_last_four_ssn).to eq("1224")
+    expect(download.veteran_name).to_not be_nil
```
kierachell: you can also add `be_empty?` or `be_blank?` to make sure it's not an empty string like `" "`

Sjones352: Ok doing it now

Issue #625, opened 2017-09-20
DownloadManifestJob should download as many docs as possible


mdbenjam:
Currently the `DownloadManifestJob` does the following:

1) Try to get a list of documents from VBMS.
1) Try to get a list of documents from VVA.
1) If either of the above fail, abort.

Instead of failing, we should try both document sources at all times. We should also add two fields to the downloads table to mark when each of VVA documents and VBMS documents have been uploaded. On the UI path we should throw an error if either of these were not updated, but the API should work never fail. We should add these timestamps to the API. This way the API can serve whatever documents were available.

# AC
1. Verify the API returns documents from VVA even if VBMS fails (http error only).
1. Verify the API returns documents from VBMS even if VVA fails (http error only).
1. Verify the API returns the timestamp that the VVA and VBMS documents were retrieved.
1. Verify the UI only returns documents if both VVA and VBMS succeed.

kierachell:
### Download manifest is working
<img width="532" alt="screen shot 2017-10-18 at 9 07 45 am" src="https://user-images.githubusercontent.com/18075411/31722643-a7358cd2-b3ea-11e7-8898-63507bb1fe8c.png">
<img width="571" alt="screen shot 2017-10-18 at 9 08 03 am" src="https://user-images.githubusercontent.com/18075411/31722644-a7912592-b3ea-11e7-889a-b46eb8703c83.png">

kierachell:
#### What kind of errors are you expecting to handle?

- 504?
- unreachable?

### The following filenumber has VVA documents, should it still work?

<img width="681" alt="screen shot 2017-10-18 at 11 58 04 am" src="https://user-images.githubusercontent.com/18075411/31728948-ab48375a-b3fb-11e7-8e81-931a4437e4c9.png">

sunil-sadasivan:
Hmm, thanks @kierachell.  Yeah I believe for now we're not handling connection timeouts gracefully, though we should.  

I believe this handles 5XXs from the service, though I could be wrong.  @kierachell, I'll find you tomorrow to discuss.

kierachell:
Check alerts channel. Seems like some VVA errors are not caught.

sunil-sadasivan:
@kierachell, I'm going to create a new ticket for better error handling of network specific errors.  I believe at the moment this will only handle cases where VBMS/VVA library throws a ClientError. 

(new ticket to handle service timeouts and network related issues)
[...]

PR #628, opened 2017-09-29
Refactor Fakes::DownloadManifestJob 

sunil-sadasivan:
In order to tackle #625, I found it ideal to not have a separate Fakes::DownloadManifestJob and instead ensure all fake logic is within Fakes::DocumentService.  

This abstracts and allows us to hit the normal codepath for DownloadDocuments. 

AC to verify this works well:
- [ ] Ensure all local DEMO cases work as expected with localhost:3000 and the local api.


app/services/download_documents.rb
```diff
@@ -38,11 +38,12 @@ def create_documents
     Download.transaction do
       @external_documents.each do |external_document|
         # JRO and SSN are required when searching for a document in VVA
+        type_id = external_document.try(:doc_type) || external_document.type_id
```
mdbenjam: Why is this better than the existing `external_document.doc_type || external_document.type_id`? I think it makes sense to bring it up here, but what does using try do?

app/models/download.rb
```diff
@@ -225,7 +225,7 @@ def find_or_create_by_user_and_file(user_id, file_id)
 
   def force_fetch_manifest_if_expired!
     return if manifest_fetched_at && manifest_fetched_at > 3.hours.ago
-    demo? ? Fakes::DownloadManifestJob.perform_now(self) : DownloadManifestJob.perform_now(self)
+    DownloadManifestJob.perform_now(self)
```
mdbenjam: 🙏

lib/fakes/document_service.rb
```diff
@@ -1,8 +1,61 @@
...
+      num_docs: 0
+    },
+    "DEMODEFAULT" => {
+      manifest_load: 4,
+      num_docs: 10,
+      max_file_load: 3
+    }
+  }.freeze
+
+  def self.fetch_documents_for(download)
```
mdbenjam: This is great here. Now the fakes and real document services line up!

lib/fakes/document_service.rb
```diff
@@ -1,8 +1,61 @@
...
+    demo = DEMOS[download.file_number] || DEMOS["DEMODEFAULT"]
+    sleep_manifest_load(demo[:manifest_load])
+
+    check_and_raise_errors(demo)
+    docs = create_documents(download, demo[:num_docs])
```
mdbenjam: I think we should parallel the VBMS service a little more closely:

```
def self.fetch_documents_for(download)
    @vbms_client ||= init_client

    request = if FeatureToggle.enabled?(:vbms_efolder_service_v1)
                VBMS::Requests::FindDocumentSeriesReference.new(download.file_number)
              else
                VBMS::Requests::ListDocuments.new(download.file_number)
              end
    documents = send_and_log_request(download.file_number, request)
    Rails.logger.info("VBMS Document list length: #{documents.length}")
    documents
  end
```

It only uses the download object for the file_number. Here we're creating the documents and assigning them to the download object. That's duplicating what the real job is doing.

lib/fakes/document_service.rb
```diff
@@ -12,4 +65,30 @@ def self.fetch_document_file(document)
...
+  end
+
+  def self.sleep_manifest_load(wait)
+    sleep(wait || 0)
+  end
+
+  def self.create_documents(download, number)
+    docs = []
+    (number || 0).times do |i|
+      doc = download.documents.create(
```
mdbenjam: Echoing from above, I think this should just use an [open struct](http://ruby-doc.org/stdlib-2.0.0/libdoc/ostruct/rdoc/OpenStruct.html) to create the docs with the same fields.

app/services/download_documents.rb
sunil-sadasivan: Good one Mark.  So I believe .try will not throw an `NoMethodError` exception if the doc_type method doesn't exist.  (instead .try will return nil).  So it's a 

For the local faker, we do not have doc_type as a method, and instead fall back to .type_id.

lib/fakes/document_service.rb
sunil-sadasivan: Good call @mdbenjam

lib/fakes/vva_service.rb
```diff
@@ -0,0 +1,7 @@
+class Fakes::VVAService < Fakes::DocumentService
+  @service_type = "VVA"
+
+  def self.raise_error
+    fail VVA::ClientError
+  end
+end
```
sunil-sadasivan: @mdbenjam, I've created these two service child classes to allow us to better test exceptions from each service.    In the future I can see us expanding on this to better mock how each service can behave.

lib/fakes/document_service.rb
```diff
@@ -50,11 +50,7 @@ class Fakes::DocumentService
   }.freeze
 
   def self.fetch_documents_for(download)
-    demo = DEMOS[download.file_number] || DEMOS["DEMODEFAULT"]
-    sleep_manifest_load(demo[:manifest_load])
-
-    check_and_raise_errors(demo)
-    docs = create_documents(download, demo[:num_docs])
+    docs = list_fake_documents(download.file_number)
```
mdbenjam: nit: you don't need to save this to a variable. Just return it directly.

lib/fakes/document_service.rb
```diff
@@ -66,28 +62,37 @@ def self.fetch_document_file(document)
...
+  # can be overridden by child classes to provide more specific error
+  def self.raise_error
+    fail "Could not obtain docs"
+  end
+
   def self.check_and_raise_errors(demo)
     return unless demo[:error]
-    fail VBMS::ClientError if demo[:error_type] == "VBMS"
-    fail VVA::ClientError if demo[:error_type] == "VVA"
+    self.raise_error if demo[:error_type] == @service_type
```
mdbenjam: nit: I feel like using a `@service_type` is a little opaque. Could we just make this a method instead? I think it's clearer when you have to implement a method, rather than set a variable in a subclass.

lib/fakes/document_service.rb
```diff
@@ -66,28 +62,37 @@ def self.fetch_document_file(document)
...
     docs = []
-    (number || 0).times do |i|
-      doc = download.documents.create(
-        vbms_filename: "happy-thursday-#{SecureRandom.hex}.pdf",
-        type_id: Document::TYPES.keys.sample,
-        document_id: "{#{SecureRandom.hex(4).upcase}-#{SecureRandom.hex(2).upcase}-#{SecureRandom.hex(2).upcase}-#{SecureRandom.hex(2).upcase}-#{SecureRandom.hex(6).upcase}}",
-        mime_type: "application/pdf",
-        received_at: (i * 2).days.ago,
-        downloaded_from: rand(5) == 3 ? "VVA" : "VBMS"
+    (demo[:num_docs] || 0).times do |i|
```
mdbenjam: Super nit: I prefer using map over pushing to an array. You could do something like:

`(0.. (demo[:num_docs] || 0)).to_a.map do |i|`

then just return the open struct in each iteration. Then you don't even have to return docs at the end of this method. The map will return the array for you.

PR #630, opened 2017-10-02
[WIP] Handle service errors gracefully in download manifest job

sunil-sadasivan:
connect #625 

This PR adds support for the file API to return gracefully if VBMS or VVA services fail.  Fetched docs will be returned. 

We now store details of and return with the API when VVA and VBMS manifests are fetched.  `manifest_fetched_at` is still provided in the response for backwards compatibility on clients.  

@mdbenjam there's a lot that's touched here so let me know if you want to pair on this.  I'm also not 100% happy with the approach as ideally we'd refactor the Download model to support running jobs in parallel.


app/jobs/download_manifest_job.rb
```diff
@@ -1,14 +1,49 @@
...
+    external_documents = []
+    has_error = false
+
+    # fetch vbms docs
+    begin
+      external_documents += download_vbms(download)
+    rescue VBMS::ClientError => e
+      has_error = true
+      capture_error(e, download, :vbms_connection_error)
+      return if !graceful
```
sunil-sadasivan: I'm not the proudest of this approach for flow control, however struggled to find better alternatives.  Ideally we should be running multiple jobs rather than taking this approach.  

However for that approach, we'll need to handle coordination and race conditions in the Download model which will require refactoring that doesn't seem necessary to tackle at this point.

app/jobs/download_manifest_job.rb
```diff
@@ -1,14 +1,49 @@
...
+  end
 
+  def download_vva(download)
+    external_documents = []
     if FeatureToggle.enabled?(:vva_service, user: download.user)
-      external_documents += VVAService.fetch_documents_for(download)
+      external_documents = VVAService.fetch_documents_for(download)
+      download.update_attributes!(manifest_vva_fetched_at: Time.zone.now)
     end
+    external_documents
```
mdbenjam: Maybe a little more ruby style would be to not initialize it and return `external_documents || []`

lib/fakes/document_service.rb
```diff
@@ -80,7 +80,9 @@ def self.sleep_manifest_load(wait)
   end
 
   def self.list_fake_documents(file_number)
-    demo = DEMOS[file_number] || DEMOS["DEMODEFAULT"]
+    demo = DEMOS[file_number]
+    return [] if !demo || demo[:num_docs] <= 0
```
mdbenjam: Was there no way to test the no document case before?

app/jobs/download_manifest_job.rb
mdbenjam: Yeah it's definitely ugly... but I think it's the cleanest way to approach this in the current configuration.

db/migrate/20171002160028_add_manifest_vva_fetched_at_to_downloads.rb
```diff
@@ -0,0 +1,5 @@
+class AddManifestVvaFetchedAtToDownloads < ActiveRecord::Migration
```
mdbenjam: Why add these in two different migrations? No real problem with it, just curious.

app/jobs/download_manifest_job.rb
```diff
@@ -1,14 +1,49 @@
 class DownloadManifestJob < ActiveJob::Base
   queue_as :default
 
-  def perform(download)
+  # pass graceful=true if the job should continue to obtain doc manifests from other services after a failure
+  def perform(download, graceful = false)
```
mdbenjam: We probably should add some unit tests for this in the `download_manifest_job_spec.rb` since behavior is changing some.

app/jobs/download_manifest_job.rb
```diff
@@ -1,14 +1,49 @@
...
+    rescue VVA::ClientError => e
+      capture_error(e, download, :vva_connection_error)
+      has_error = true
+      return if !graceful
+    end
+
+    create_documents(download, external_documents, has_error)
+  end
+
+  def download_vbms(download)
```
mdbenjam: Let's make these private since we're not using them outside this class.

app/models/download.rb
```diff
@@ -225,7 +225,7 @@ def find_or_create_by_user_and_file(user_id, file_id)
 
   def force_fetch_manifest_if_expired!
     return if manifest_fetched_at && manifest_fetched_at > 3.hours.ago
```
mdbenjam: Regarding our discussion around what manifest_fetched_at should be if one service fails, I think we can leave this logic how it is, but should probably make a new ticket to move this logic into the job. We should really check each timestamp separately.

lib/fakes/document_service.rb
sunil-sadasivan: Yeahh so to test the no document case, you pass DEMO_NO_DOCUMENT.  However I figured it's important to also test the invalid id case.  Invalid ID previous would assume DEMODEFAULT

app/jobs/download_manifest_job.rb
sunil-sadasivan: Ahh ruby 👍

db/migrate/20171002160028_add_manifest_vva_fetched_at_to_downloads.rb
sunil-sadasivan: Yeah good q @mdbenjam, tbh no strong reason from my side. It was how I added it in my flow (one at a time vs together).  

I generally do have a personal preference for smaller scoped/limited migration changsets.

app/jobs/download_manifest_job.rb
aroltsch: So the idea is to continue fetching a manifest from another service even if one fails?

sunil-sadasivan: Exactly @aroltsch

app/models/download.rb
```diff
@@ -235,7 +235,7 @@ def prepare_files_for_api!(start_download: false)
   end
 
   def start_fetch_manifest
-    DownloadManifestJob.perform_later(self)
+    DownloadManifestJob.perform_later(self, false)
```
aroltsch: You can remove `false` as it is a default value.

app/jobs/download_manifest_job.rb
```diff
@@ -1,14 +1,49 @@
...
+    download.update_attributes!(manifest_vbms_fetched_at: Time.zone.now)
+    external_documents
+  end
 
+  def download_vva(download)
+    external_documents = []
     if FeatureToggle.enabled?(:vva_service, user: download.user)
-      external_documents += VVAService.fetch_documents_for(download)
+      external_documents = VVAService.fetch_documents_for(download)
+      download.update_attributes!(manifest_vva_fetched_at: Time.zone.now)
```
aroltsch: I think it would be cleaner if we separated this job into 2 jobs and called the jobs synchronously. We could have a job helper where we can place shared methods between two jobs.

aroltsch: I think there is a way to run jobs one after the other. However, it might complicate the situation.

db/migrate/20171002160028_add_manifest_vva_fetched_at_to_downloads.rb
aroltsch: Why did we decide to add these two additional fields? Just curious.

app/models/download.rb
sunil-sadasivan: 👍

db/migrate/20171002160028_add_manifest_vva_fetched_at_to_downloads.rb
sunil-sadasivan: Yeah @aroltsch, it seems as if @mdbenjam has a client which may want to display the last fetched at dates from VBMS and VVA for a better user experience.  In order to do that, we'd need to store those dates.

app/models/download.rb
sunil-sadasivan: that's a good idea.  As part of separating this out into multiple services, refetching the specific service when one has failed would be ideal!  I'll create a ticket around that.

sunil-sadasivan: Created ticket #632 for this

sunil-sadasivan:
@mdbenjam @aroltsch, I've pushed up some adjustments from the feedback, including ensuring our test suite is updated with the current changes.  Let me know how things feel.  I hope to move fairly quickly to #632 inspired from the comments here.

app/jobs/download_manifest_job.rb
```diff
@@ -54,6 +39,24 @@ def create_documents(download, external_documents, has_error)
     )
     download_documents.create_documents
     download.update_attributes!(status: :pending_confirmation) if !has_error
+    download_documents
+  end
+
+  private
```
mdbenjam: I think you could even put `create_documents` in here.

spec/jobs/download_manifest_job_spec.rb
```diff
@@ -16,11 +16,20 @@
...
+          expect(download.reload).to be_vbms_connection_error
+        end
+
+        it "does not return documents on error" do
+          expect(DownloadManifestJob.perform_now(download)).to be_nil
+        end
 
-      it "saves download status as :vbms_connection_error" do
-        expect(download.reload).to be_vbms_connection_error
+        it "returns documents on error" do
```
mdbenjam: I would move that comment into the description of the it block for clarity.

spec/models/download_spec.rb
```diff
@@ -342,12 +342,19 @@
...
+        download.update_attributes!(manifest_vva_fetched_at: Time.zone.now - 2.hours)
       end
 
       it "does not start the manifest job" do
+        download.update_attributes!(manifest_vbms_fetched_at: Time.zone.now - 2.hours)
         download.force_fetch_manifest_if_expired!
         expect(DownloadManifestJob).to_not have_received(:perform_now)
       end
+
+      it "with a service error does start the manifest job" do
```
mdbenjam: I would change this title to something like `starts the manifest job when vbms request failed`. Just make service error a little more descriptive.

app/models/download.rb
```diff
@@ -224,7 +224,9 @@ def find_or_create_by_user_and_file(user_id, file_id)
   end
 
   def force_fetch_manifest_if_expired!
-    return if manifest_fetched_at && manifest_fetched_at > 3.hours.ago
+    # if all services have succeeded and has been fetched in the last three hours, don't refetch
+    # TODO (Sunil) - Ideally we should only retry the failed job
+    return if manifest_fetched_at && manifest_vva_fetched_at && manifest_vbms_fetched_at && manifest_fetched_at > 3.hours.ago
```
mdbenjam: I think we should prioritize the TODO, this behavior seems a little strange now. If VVA is down the first time we request documents we'll keep request docs from VBMS until VVA comes back up. Where as if VVA is not down the first time, and then goes down, we won't retry for 3 hours at a time. That's hard behavior to explain to users.

sunil-sadasivan: Yeah hmm @mdbenjam, I'm going to think about that.  Seems like we're quickly running into #632 as being important.   I may refactor the API endpoint and job slightly to allow for retries for specific services.  Will get back to you tonight.

sunil-sadasivan:
Iterating on this PR here: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/630. 

Closing for now

PR #634, opened 2017-10-12
[WIP] Refactor DownloadManifestJobs by service

sunil-sadasivan:
Connect #625 

This creates multiple DownloadManifest jobs by service.  This now allows us to refetch a manifest only the service that has failed if it's not cached... reducing pressure on the other service.


app/models/download.rb
```diff
@@ -1,3 +1,5 @@
+# rubocop:disable Metrics/ClassLength
+# #626 is a task to refactor the Download model
```
sunil-sadasivan: We're encountering linting issues with the class length being too long. Feels like this class is fairly messy and needs some refactoring as part of the conversations we've talked about earlier.

app/jobs/download_manifest_job.rb
```diff
@@ -1,13 +1,32 @@
 class DownloadManifestJob < ActiveJob::Base
   queue_as :default
 
+  # these must be implemented by child classes
+  def service(_download)
+    nil
```
mdbenjam: I think we should throw errors instead of nil here. That way we don't have to check for nil later.

app/models/download.rb
```diff
@@ -223,9 +225,39 @@ def find_or_create_by_user_and_file(user_id, file_id)
...
-    return if manifest_fetched_at && manifest_fetched_at > 3.hours.ago
-    DownloadManifestJob.perform_now(self)
+    vbms_error = fetch_vbms_manifest
+    vva_error = fetch_vva_manifest
+
+    error = vbms_error || vva_error
+    update_attributes!(status: error) if error
+  end
+
+  def capture_error(e)
```
mdbenjam: I'm not sure if you need this. In application_controller we already `rescue_from StandardError`. It's a little different but probably sufficient.

app/jobs/download_all_manifest_job.rb
```diff
@@ -0,0 +1,20 @@
...
+    capture_error(e, download, :vbms_connection_error)
+  rescue VVA::ClientError => e
+    capture_error(e, download, :vva_connection_error)
+  end
+
+  def capture_error(e, download, status)
+    Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+    Raven.capture_exception(e)
+    download.update_attributes!(status: status)
+    nil
```
mdbenjam: What's the nil for?

app/jobs/download_manifest_job.rb
```diff
@@ -1,13 +1,32 @@
...
+    create_documents(download, external_documents)
+  end
 
-    if FeatureToggle.enabled?(:vva_service, user: download.user)
-      external_documents += VVAService.fetch_documents_for(download)
-    end
+  private
 
+  def create_documents(download, external_documents)
+    # only indicate no_documents status if we've successfully completed fetching from services
```
mdbenjam: What if VBMS returns docs, but VVA doesn't. Since we look at external_documents separately, won't we mark the status as no_documents since our second query has no documents?

app/models/download.rb
sunil-sadasivan: Hmm @mdbenjam, so what this is doing is logging the standard error and handling the error gracefully.  I think in the application, we do rescue from `StandardError` but in that case we throw a 500 response, which is slightly different.  

I'm thinking of simply moving the capture_error method to the application controller though.

app/jobs/download_all_manifest_job.rb
```diff
@@ -0,0 +1,28 @@
+class DownloadAllManifestJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(download)
+    # sequentially download documents from all services
+    error, vbms_documents = DownloadVBMSManifestJob.perform_now(download)
+    if error
+      download.update_attributes!(status: error)
+      return
+    end
```
sunil-sadasivan: In this job, we are sequentially downloading manifests from each service.  If an error is returned, simply no longer finish this request.

app/models/download.rb
```diff
@@ -223,9 +225,43 @@ def find_or_create_by_user_and_file(user_id, file_id)
...
+    return true, nil, nil if manifest_vva_fetched_at && manifest_vva_fetched_at > 3.hours.ago
+    error, docs = DownloadVVAManifestJob.perform_now(self)
+    [false, error, docs || []]
+  end
+
   def force_fetch_manifest_if_expired!
-    return if manifest_fetched_at && manifest_fetched_at > 3.hours.ago
-    DownloadManifestJob.perform_now(self)
+    vbms_cached, vbms_error, vbms_docs = fetch_vbms_manifest
+    vva_cached, vva_error, vva_docs = fetch_vva_manifest
```
sunil-sadasivan: Found this to be the best approach to coordinate different logic states that fetch_vbms/vva_manifest can return.  Let me know if you can think of a better way to maintain this.  (Didn't want to include too many raise/rescue patterns)

app/jobs/download_manifest_job.rb
```diff
@@ -1,38 +1,62 @@
...
+    "#{service_name}_connection_error"
+  end
+
+  # obtain docs from a service and save to the document model
+  # retuns <error>, <array of fetched documents>
   def perform(download)
-    external_documents = VBMSService.fetch_documents_for(download)
+    external_documents = download_from_service(download)
+    create_documents(download, external_documents) if !external_documents.empty?
+    return nil, external_documents
```
mdbenjam: I did not know you could return tuples!

app/jobs/download_manifest_job.rb
```diff
@@ -1,38 +1,62 @@
...
-    download.update_attributes!(status: :pending_confirmation)
+  end
 
-  rescue VBMS::ClientError => e
-    capture_error(e, download, :vbms_connection_error)
-  rescue VVA::ClientError => e
-    capture_error(e, download, :vva_connection_error)
+  def download_from_service(download)
+    service = get_service(download)
+    return [] if !service
```
mdbenjam: I think we should remove this line. First, we fail the get_service method if it's not implemented. Second, this could cause us to silently fail getting documents from one of the services. We should just allow it to error on the next line if this is nil.

mdbenjam: Hmmm, okay I see you're returning nil for the VVA service if the feature flag isn't on. It works for now, but I think we should remove this when we turn it on for everyone.

app/models/download.rb
```diff
@@ -223,9 +225,43 @@ def find_or_create_by_user_and_file(user_id, file_id)
...
+    error = vbms_error || vva_error
+    if error
+      update_attributes!(status: error)
+      return
+    end
+
+    # only update download status at this point if we're not using
+    # cached manifests
+    if !vbms_cached && !vva_cached
+      external_documents = vbms_docs + vva_docs
```
mdbenjam: What happens if vbms_docs is cached, and vva_docs is empty but not cached. We'll updated the status to no_documents. I think we might be able to avoid using this cached variable if we instead always return docs. Either what we have cached, or the newly downloaded docs.

app/models/download.rb
```diff
@@ -223,9 +225,43 @@ def find_or_create_by_user_and_file(user_id, file_id)
...
+      if external_documents.empty?
+        update_attributes!(status: :no_documents)
+      else
+        update_attributes!(status: :pending_confirmation)
```
mdbenjam: We probably don't need to use this state in this path. Since this record will never be used in the UI. But maybe that's a bad idea to not use our states consistently, so I'm ambivalent about setting this.

app/jobs/download_manifest_job.rb
sunil-sadasivan: Yes agreed @mdbenjam, it was purely because of the featureflag that I had to include this line.  Do we have a sense when we can turn VVA for everyone?

app/models/download.rb
sunil-sadasivan: Yeah @mdbenjam... feels hacky to not having consistent state, which is why I was thinking about keeping it.  Don't feel too strongly myself. I think it isn't ideal to have a success stuck in the fetching_manifest state though.

app/models/download.rb
```diff
@@ -223,9 +225,45 @@ def find_or_create_by_user_and_file(user_id, file_id)
...
   end
 
+  def get_cached_documents(service)
+    documents.where(downloaded_from: service)
+  end
+
+  # returns <service error>, <docs>
+  def fetch_vbms_manifest
+    # cache manifests for 3 hours
+    return nil, get_cached_documents("VBMS") if manifest_vbms_fetched_at && manifest_vbms_fetched_at > 3.hours.ago
```
sunil-sadasivan: @mdbenjam, we now return the cached document list.

spec/jobs/download_all_manifest_job_spec.rb
```diff
@@ -0,0 +1,76 @@
+describe DownloadAllManifestJob do
```
mdbenjam: Should there be a both succeeds test?

spec/models/download_spec.rb
```diff
@@ -318,35 +319,54 @@
...
+        expect(DownloadVVAManifestJob).to_not have_received(:perform_now)
+      end
+
+      it "does start the VBMS manifest job" do
+        download.update_attributes!(manifest_vbms_fetched_at: Time.zone.now - 4.hours)
+        download.force_fetch_manifest_if_expired!
+        expect(DownloadVBMSManifestJob).to have_received(:perform_now)
+      end
+
+      it "does not start the VBMS manifest job" do
```
mdbenjam: I'd add "when VBMS manifest was fetched less than 3 hours ago" to the end of this it description. I think having two tests in the same context with opposite names is confusing.

sunil-sadasivan: Ah yeah good point!

PR #639, opened 2017-10-16
Stop using old s3 filename convention

lowellnava:
Connects: #583. Know all files with old filenaming convention have expired, will monitor when pushed live.


spec/models/document_spec.rb
```diff
@@ -61,22 +61,12 @@
     subject { document.s3_filename }
 
     let(:document) do
-      Document.new(id: 5, document_id: "{TEST}", mime_type: "application/pdf", download_id: 45)
```
mdbenjam: Why was this removed?

mdbenjam:
I just checked in our S3 bucket and it looks like everything is named with the new convention.

spec/models/document_spec.rb
lowellnava: I removed it because I didn't think it would be used anywhere after old_s3_filename is removed. `grep -r download_id caseflow-efolder | grep -v test.log` looks like it may still exist in db/schema.rb though so maybe I was a little hasty to remove this attribute. (

lowellnava: Just chatted with @mdbenjam and this is some rails magic that I didn't understand. Adding it back in.

PR #644, opened 2017-10-19
Add sensitivity check

mdbenjam:
The goal of this PR is to ensure users can't access cases with sensitivity levels above their permissions via the eFolder API. To accomplish this, we need to call a BGS endpoint from the API path. Unfortunately the code was not really set up for this. BGSService was associated with a class attribute on the Download class. In order to make this new change cleaner, and in order to bring eFolder in line with how we approach caseflow, I've updated our setup so it's the same.

Now we have a `BGSService` service that calls either `Fakes::BGSService` or `ExternalApi::BGSService` depending on if we're in production or not. BGSService no longer takes a user object to initialize, it derives the current user from the request store. To use the service, a client can just create a new BGSService and memoize it for future calls.

We removed some code that would use `Fakes::BGSService` in production if a demo eFolder was being looked up. I don't think anyone looks up demo cases in production, and putting demo data in production seems dangerous.

Finally, we added a staging environment for eFolder. To access it all you need to do is run `rails s -e staging`. You'll also need to create the db `RAILS_ENV=staging rake db:setup`.

**Test Plan**
- [ ] Run eFolder in the new staging environment `RAILS_ENV=staging rake db:setup` then `rails s -e staging -p 4000`.
   - [ ] `curl localhost:4000/api/v1/files  -H "Authorization: Token token=token" -H "css-id: CF_Q_283" -H "station-id: 283" -H "file-number: 111222003"`. Verify you can access it.
   - [ ] `curl localhost:4000/api/v1/files  -H "Authorization: Token token=token" -H "css-id: CF_KHAN_283" -H "station-id: 283" -H "file-number: 111222003"`. Verify you receive a 403 forbidden error.
   - [ ] Log into the UI from the new log in screen to css_id: CF_Q_283 (sensitivity level 9) and station: 283. Try accessing: 111222003. Verify you can access it.
   - [ ] Log into the UI from the new log in screen to css_id: CF_KHAN_283 (sensitivity level 1) and station: 283. Try accessing: 111222003. Verify you cannot access it.


Gemfile
```diff
@@ -67,6 +67,8 @@ gem 'active_model_serializers'
 
 gem 'redis-namespace'
 
+gem 'request_store'
```
mdbenjam: We're adding this just to be in line with what Caseflow usesl

app/controllers/base_controller.rb
```diff
@@ -1,6 +1,7 @@
 class BaseController < ActionController::Base
   force_ssl if: :ssl_enabled?
   before_action :strict_transport_security
+  before_action :current_user
```
mdbenjam: We want to force `current_user` to be defined in the RequestStore.

app/models/download.rb
```diff
@@ -68,8 +72,6 @@ def fetch_veteran_info
     # than a class. Refactor the class method `bgs_service`
     # into an instance one.
     if !@veteran_info_fetched && missing_veteran_info?
-      bgs_service = demo? ? Fakes::BGSService : Download.bgs_service
```
mdbenjam: This is going to break using demo file numbers in production. But I don't think this is a safe pattern.

config/environments/staging.rb
```diff
@@ -0,0 +1,57 @@
+Rails.application.configure do
```
mdbenjam: This is just a copy of development.rb

spec/jobs/save_files_in_s3_spec.rb
```diff
@@ -14,7 +14,7 @@
         veteran_last_name: "Washington"
       )
     end
-    let(:document) do
+    let!(:document) do
```
mdbenjam: Should force creation of the document rather than calling `document` in a before block to make sure it's created.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -16,6 +17,14 @@ def show
 
   private
 
+  def document
+    @document ||= Document.find(params[:id])
+  end
+
+  def can_access?
```
mdbenjam: Do we need this one? Seems safest to add it, but I worry about latency it will cause. Maybe we should just trigger on the files endpoint. That should be sufficient. Thoughts?

app/controllers/base_controller.rb
```diff
@@ -12,14 +13,23 @@ def strict_transport_security
     response.headers["Strict-Transport-Security"] = "max-age=31536000; includeSubDomains" if request.ssl?
   end
 
+  def current_user=(user)
+    RequestStore.store[:current_user] = user
```
mdbenjam: This new current_user setter is here to set the user in the RequestStore so we can access it in our services.

lib/fakes/test_auth_strategy.rb
```diff
@@ -14,16 +14,16 @@ def request_phase
 
   def auth_hash
     hash = super
-    hash.uid = "CSFLOW" + rand(10).to_s
+    hash.uid = hash["info"]["css_id"]
```
mdbenjam: We add to the fake login page. Now you can input a css_id and station_id to test with.

app/controllers/api/v1/documents_controller.rb
mdbenjam: If we decide we do want it, I'll add a test for it.

aroltsch: This check will run on every document? What if we add one more status on the download object with access_denied so that you can check the download instead of making a call to BGS every time.

app/services/bgs_service.rb
```diff
@@ -0,0 +1,5 @@
+class BGSService
+  def self.new
+  	!BaseController.dependencies_faked? ? ExternalApi::BGSService.new : Fakes::BGSService.new
```
aroltsch: Why did you decide to go with this pattern instead of putting it in `initializers` (the same way we do in Caseflow)?

db/schema.rb
```diff
@@ -75,7 +75,6 @@
     t.integer  "user_id"
     t.integer  "zipfile_size",             limit: 8
     t.boolean  "from_api",                           default: false
-    t.datetime "vva_fetched_at"
```
aroltsch: Is there a migration for this?

app/services/external_api/bgs_service.rb
```diff
@@ -39,12 +36,20 @@ def check_sensitivity(file_number)
   private
 
   def init_client
+    # Fetch current_user from global thread
+    current_user = RequestStore[:current_user]
+
+    # This is here to make sure StartCertificationJob
+    # can pass the ip address to the BGS client.
+    # We should find a better way to do this.
+    ip_address = current_user.ip_address || RequestStore[:ip_address]
```
aroltsch: I don't see a use case where you set the ip_address in RequestStore. I do see you are setting it on the user.

app/controllers/base_controller.rb
```diff
@@ -12,14 +13,23 @@ def strict_transport_security
...
+  end
+
   def current_user
-    @current_user ||= User.from_session(session, request)
+    @current_user || self.current_user = User.from_session(session, request)
   end
   helper_method :current_user
 
+  def configure_bgs
+    Thread.current[:bgs_service] = BGSService.new(user: current_user)
```
aroltsch: I thought that you removed `attr_accessor :user` from BGSService?

app/services/external_api/bgs_service.rb
askldjd: This is pretty much unless. Our VPN performs NAT pooling for both ingress and egress. All clients will show up as a single IP.

I have been using 192.168.1.1 for Caseflow monitor. :)

app/controllers/api/v1/documents_controller.rb
askldjd: Ya, calling BGS on every download is probably a bad idea. Should we cache the sensitivity level in Redis?

app/services/external_api/bgs_service.rb
mdbenjam: Yeah I realized the same thing last night. I copied it over from Caseflow (which I think also doesn't add ip_address to RequestStore. I like Alan's idea of hardcoding this!

db/schema.rb
mdbenjam: woops I'll remove this. Changes from another branch.

app/services/bgs_service.rb
mdbenjam: I did that at first, but every time I'd refresh I'd get some error about it being added to the tree too many times. I think it had to do with the rails magic reloading. I think this is actually a better model since changes will take effect on save rather than requiring us to restart the server to see changes in either ExternalApi::BGSService or Fakes::BGSService.

app/controllers/api/v1/documents_controller.rb
mdbenjam: I've updated this. We really need to check if the download this document is associated is owned by this user.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -16,6 +17,14 @@ def show
...
+    forbidden("sensitive record") if document.download.user.css_id != current_user.css_id
```
aroltsch: I think it would better if you moved this check `document.download.user.css_id != current_user.css_id` into a document.rb

app/controllers/base_controller.rb
```diff
@@ -12,14 +13,19 @@ def strict_transport_security
...
   def current_user
-    @current_user ||= User.from_session(session, request)
+    @current_user || self.current_user = User.from_session(session, request)
   end
   helper_method :current_user
 
   class << self
     def dependencies_faked?
-      Rails.env.development? || Rails.env.test? || Rails.env.demo?
+      (Rails.env.development? || Rails.env.test? || Rails.env.demo?) && Rails.env != 'staging'
```
aroltsch: Im confused why you are adding `&& Rails.env != 'staging'`. If it is staging, it should return false for `Rails.env.development? || Rails.env.test? || Rails.env.demo?`

app/services/external_api/bgs_service.rb
```diff
@@ -39,12 +36,16 @@ def check_sensitivity(file_number)
...
+    current_user = RequestStore[:current_user]
+
+    # We hardcode the ip since all clients show up as a single IP anyway.
     BGS::Services.new(
       env: Rails.application.config.bgs_environment,
       application: "CASEFLOW",
-      client_ip: user.ip_address,
-      client_station_id: user.station_id,
-      client_username: user.css_id,
+      client_ip: "192.168.1.1",
```
aroltsch: This IP will never change?

mdbenjam: @askldjd Is this what you did? Or should we get the IP some other way?

Issue #647, opened 2017-10-19
Searching twice for a non-existent SSN causes downloads to hang


nickheiner-usds:
## Repro Steps
1. Search eFolder for a non-existent SSN 
1. See that it gives an error message saying that the SSN is not found
1. Search eFolder for the same SSN again
1. See that the "gathering files now" page hangs indefinitely

@mdbenjam has an explanation for why this occurs.

# AC
1. Verify that in the above flow, the page does not hang, but gives the correct error message saying that SSN is not found.

kierachell:
**PASSED**

- Tried all kinds of wonky IDs... can't get it to hang no more

I can finally remove the logic to handle this from integration tests 👏 

<img width="666" alt="screen shot 2017-11-02 at 8 55 39 pm" src="https://user-images.githubusercontent.com/18075411/32357128-421affda-c010-11e7-84cd-983fa3411f7a.png">
<img width="538" alt="screen shot 2017-11-02 at 8 55 46 pm" src="https://user-images.githubusercontent.com/18075411/32357129-4229789e-c010-11e7-8a6f-c6cd05561e42.png">
<img width="639" alt="screen shot 2017-11-02 at 8 55 56 pm" src="https://user-images.githubusercontent.com/18075411/32357130-42364dbc-c010-11e7-8408-86c86f28c790.png">

PR #648, opened 2017-10-19
Fixing hanging download bug

mdbenjam:
Connects: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/647

When you search for an eFolder via the UI we create a search object. If the download is new, we create a new download record to associate with the search object. We then do a validation on that download object to see if 1) the veteran exists, 2) the user has access to this case file's sensitivity. If either of these fail, the intention is for the download object not to be saved to the DB.

However, since we set the search object's download attribute to `self.download = Download.new()` if we ever save anything from this search record to the DB, we'll inadvertently save this download object as well #railsmagic. In the validate method if either of the failure cases exist: 1) the veteran exists, 2) the user has access to this case file's sensitivity, then we call `update_attributes` to set the status of the search object. This then saves the download object as well.

On a subsequent search, the search model will associate the new search with the old download object. The user is then forwarded to the download record's `show` page, and the user's browser will poll that page forever. Since this veteran doesn't actually exist, the user will never be notified that a download finished, since it never started.

With this change, we only save the download object if we pass the validation.

TL;DR: We should not save download records if the search fails.

### DIFF
--- a/app/models/search.rb
+++ b/app/models/search.rb
@@ -30,13 +30,16 @@ class Search < ActiveRecord::Base
   def perform!
     return true if match_existing_download
 
-    self.download = download_scope.new
+    download = download_scope.new
 
-    return false unless validate!
+    return false unless validate!(download)
 
     transaction do
-      update_attributes!(status: :download_created)
       download.save!
+      update_attributes!(
+        status: :download_created,
+        download: download
+      )
     end
 
     # We fetch the veteran info from BGS before starting the job so that we do any DB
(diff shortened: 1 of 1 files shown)


nickheiner-usds:
LGTM but we should have someone else who is more familiar with eFolder take a look as well.

cmgiven:
Note that we keep stats on bad searches, IIRC. Check the stats module to see the logic being used.

mdbenjam:
@cmgiven this will be more accurate now. If a user keeps searching for not found cases, each search will show up in these stats. Originally it was just the first such search.

Issue #661, opened 2017-10-27
Investigate "Document ID does not exist" errors


amprokop:
https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/529/events/11175/

Look at the text in the body of the error. Sentry has grouped it oddly.

## Questions
- How can this error happen? We fetch our documents according to the manifest received from VBMS.

## Steps
- Are there more instances of this error than the ones associated with the Sentry link above? @kierachell says it has been happening since the summer. Search AWS logs to get a sense of how often this is happening.
- Is this reproducible? Try to call VBMS using the document ids associated with these errors and see if they're really there.

## AC
When VBMS returns a "Document ID does not exist" error, we track the information in the database instead of throwing an error.

nickheiner-usds:
Yeah, this is odd – why does Sentry contain both `WssVerification Exception` and `RecordNotFoundException`? It seems like the one you want me to look in to is the latter.

nickheiner-usds:
I made requests manually, and found the following:

## VMBS says they exist
```
{40052652-0000-CE1C-89B5-070AECDA011E}
{7E0A03DF-02E5-4DAA-8A0C-25371929997F}
{F1C637DB-3904-40B4-AA9B-14E7AF845E5F}
```

## VBMS says they do not exist
```
ExternalApi::VBMSService.fetch_document_file(Document.new(document_id: '{BE836C2D-5162-4ACB-8E21-8CA1D3D9871A}'))
{4D83C104-913B-41E0-96E1-AED635FD3570}
```

The error message I got back was the same reported by Sentry. I retried this request, but it failed a second time as well.

I used the following line of code to test:

```rb
ExternalApi::VBMSService.fetch_document_file(Document.new(document_id: doc_id))
```

amprokop:
Cool. So then potentially we need to handle this error on the application level — if we observe that some documents exist in the VBMS manifest but not in the eFolder, then maybe we should just catch it, log it, and move on, no?

nickheiner-usds:
I made a `FindDocumentSeriesReference` call and confirmed that `{BE836C2D-5162-4ACB-8E21-8CA1D3D9871A}` does indeed exist in a manifest. 

> if we observe that some documents exist in the VBMS manifest but not in the eFolder, then maybe we should just catch it, log it, and move on, no?

[...]

Issue #663, opened 2017-10-30
Improve how we store document errors in the db


nickheiner-usds:
When we have a downstream dependency error, we save the error to the database:

```rb
      rescue VBMS::ClientError => e
        update_document_with_error(document, "VBMS::ClientError::#{e.message}\n#{e.backtrace.join("\n")}")
```

This does not actually produce much useful information. In one case, this looks like:

```
VBMS::ClientError::Unable to parse SOAP message

/opt/efolder-express/src/app/services/vbms_service.rb:31:in `rescue in fetch_document_file'\n/opt/efolder-express/src/app/services/vbms_service.rb:20:in `fetch_document_file'\n/opt/efolder-express/src/app/services/download_documents.rb:35:in `fetch_document'\n/opt/efolder-express/src/app/services/download_documents.rb:52:in `block in download_contents'\n/opt/efolder-express/src/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.7.1/lib/active_record/relation/delegation.rb:132:in `each'\n/opt/efolder-express/src/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.7.1/lib/active_record/relation/delegation.rb:132:in `each_with_index'\n/opt/efolder-express/src/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.7.1/lib/active_record/relation/delegation.rb:132:in `public_send'\n/opt/efolder-express/src/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.7.1/lib/active_record/relation/delegation.rb:132:in `method_missing'\n/opt/efolder-express/src/vendor/bundle/ruby/2.3.0/gems/activerecord-4.2.7.1/lib/active_record/relation/delegation.rb:99:in `method_missing'\n/opt/efolder-express/src/app/services/download_documents.rb:48:in `download_contents'\n/opt/efolder-express/src/app/services/download_documents.rb:151:in `download_and_package'\n/opt/efolder-express/src/app/jobs/get_download_files_job.rb:7:in `perform'
```

**Note:** This may actually be a fix we do in [connect_vbms](https://github.com/department-of-veterans-affairs/connect_vbms/blob/63c456fee8e55e4cb03190b66c809876d5d72746/lib/vbms/client.rb#L167).

If we want to reach out to the VBMS team to report bugs, we need more detail. For instance, we need to see more details about the request we made.

[...]

PR #664, opened 2017-10-30
[#661] Handle document download errors for API

nickheiner-usds:
Connects #661.

## Test
There are two areas to look at:

1. When VBMS / VVA fail, does the UI do the right thing?
    * Currently I am [having trouble](https://dsva.slack.com/archives/C2ZAMLK88/p1509392223000037) verifying this locally.
1. When VBMS / VVA fail, does the API do the right thing?
    * I just manually changed `fetch_content!` to raise various errors, and observed how the API would respond. I don't know if we have or should have a known ID that will raise those errors.
    * To most easily test this, comment out the authentication logic so you don't need to bother with crafting an authenticated request:
```diff
diff --git a/app/controllers/api/v1/application_controller.rb b/app/controllers/api/v1/application_controller.rb
index e76e16f..2f7bc3c 100644
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -41,13 +41,13 @@ class Api::V1::ApplicationController < BaseController
     # authenticates, then we create the current user based on the css_id and station_id
     # passed in the header. Otherwise we try to create the current user from the session
     # and authorize based on the presence of the Reader role.
-    if authenticate_with_token
-      return missing_header("Station ID") unless station_id
-      return missing_header("CSS ID") unless css_id
-      @current_user = User.find_or_create_by(css_id: css_id, station_id: station_id)
-    elsif !user_has_role
-      unauthorized
-    end
+    # if authenticate_with_token
+    #   return missing_header("Station ID") unless station_id
+    #   return missing_header("CSS ID") unless css_id
+    #   @current_user = User.find_or_create_by(css_id: css_id, station_id: station_id)
+    # elsif !user_has_role
+    #   unauthorized
+    # end
   end
 
   def station_id
```


.gitignore
```diff
@@ -22,3 +22,5 @@
 *swp
 
 .idea/*
+
+test
```
nickheiner-usds: Is there a reason this was not `gitignore`d already? I saw this file generated every time I ran `rspec`.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,21 +1,47 @@
 class Api::V1::DocumentsController < Api::V1::ApplicationController
   def show
-    document = Document.find(params[:id])
-    # The line below enables document caching for a month.
+    begin
+      document = Document.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
```
nickheiner-usds: I don't see any need to put almost the entire method within the `begin` clause, since it's only this line that can trigger the `RecordNotFound` that we care about. So, I moved it to be more tightly scoped.

app/models/document.rb
```diff
@@ -29,6 +29,19 @@ def filename
     Zaru.sanitize! "#{cropped_type_name}-#{filename_date}-#{filename_doc_id}.#{preferred_extension}"
   end
 
+  def fetch_content!(save_document_metadata:)
+    return true, fetcher.content(save_document_metadata: save_document_metadata)
+  rescue VBMS::ClientError => e
+    update_with_error "VBMS::ClientError::#{e.message}\n#{e.backtrace.join("\n")}"
```
nickheiner-usds: These messages are not useful, but that's out of scope for this PR: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/663.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,21 +1,47 @@
...
 
-    send_data(
-      document.fetcher.content(save_document_metadata: false),
-      type: document.mime_type,
-      disposition: "attachment",
-      filename: document.filename
-    )
-  rescue ActiveRecord::RecordNotFound
-    document_not_found
+    success, content, error_kind = document.fetch_content!(save_document_metadata: false)
```
nickheiner-usds: @mdbenjam the UI code path calls `@download.touch` and `document.save_locally`. Is there a reason that we don't do that from the API?

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,21 +1,47 @@
...
+    message = "Caseflow eFolder failed to fetch document contents."
+
+    if error_kind == :vva_error || error_kind == :vbms_error
+      status = 502
+      message = "An upstream dependency failed to fetch document contents."
+    end
+
+    render json: {
+      "errors": [
+        "status": "502",
```
nickheiner-usds: Why do we have this field? It feels redundant with the response status itself.

app/controllers/api/v1/documents_controller.rb
aroltsch: The API doesn't need `document.save_locally` because we are not zipping the documents. For the UI, we save docs locally to zip them into a zip file. I am not sure about `@download.touch` @NickHeiner

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,21 +1,47 @@
...
-  rescue ActiveRecord::RecordNotFound
-    document_not_found
+    success, content, error_kind = document.fetch_content!(save_document_metadata: false)
+    if success
+      send_data(
+        content,
+        type: document.mime_type,
+        disposition: "attachment",
+        filename: document.filename
+      ) if success
```
aroltsch: I think you mistakenly left this here.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,21 +1,47 @@
...
+    else
+      document_download_failed(error_kind)
```
aroltsch: I think it is cleaner to do a return earlier:
```
return document_download_failed(error_kind) unless success
send_data(...)
```

app/models/document.rb
```diff
@@ -29,6 +29,19 @@ def filename
...
+    return false, nil, :vbms_error
+  rescue VVA::ClientError => e
+    update_with_error "VVA::ClientError::#{e.message}\n#{e.backtrace.join("\n")}"
+    return false, nil, :vva_error
+  rescue ActiveRecord::StaleObjectError
+    Rails.logger.info "Duplicate download detected. Document ID: #{id}"
+    return false, nil, :caseflow_efolder_error
```
aroltsch: When you have 3 things to return, it is better to return a hash so you don't have to worry about the order and it makes it clearer what you are returning:
```
return { success: false, content: nil, error_type: :caseflow_efolder_error }
```

Also, I don't think you need both success and error_type. You can check if `error_type` is not `nil` then you know it is not successful.

app/models/document.rb
aroltsch: I would not say these messages are not useful. For example, `VBMS::ClientError::Unable to parse SOAP message` - this is what connect_vbms returns (https://github.com/department-of-veterans-affairs/connect_vbms/blob/63c456fee8e55e4cb03190b66c809876d5d72746/lib/vbms/client.rb#L167) so maybe we need to change connect_vbms. The code here would display the entire response.

app/controllers/api/v1/documents_controller.rb
aroltsch: I agree with you!

app/services/download_documents.rb
```diff
@@ -63,18 +63,13 @@ def download_contents(save_locally: true)
...
-        update_document_with_error(document, "VVA::ClientError::#{e.message}\n#{e.backtrace.join("\n")}")
+      success, content, error_kind = document.fetch_content!(save_document_metadata: true)
+      document.save_locally(content, index) if save_locally && success
 
-      rescue ActiveRecord::StaleObjectError
-        Rails.logger.info "Duplicate download detected. Document ID: #{document.id}"
+      if error_kind == :caseflow_efolder_error
         return false
+      else
+        @download.touch
```
aroltsch: Ok, I am trying to remember why we are touching the download here. It would be nice to have a comment lol

spec/requests/api/v1/documents_spec.rb
```diff
@@ -50,6 +50,45 @@
...
+      allow_any_instance_of(Fetcher).to receive(:content).and_raise(ActiveRecord::StaleObjectError.new(nil, nil))
+
+      get "/api/v1/documents/#{document.id}"
+
+      expect(response.code).to eq("500")
+
+      json = JSON.parse(response.body)
+      expect(json["errors"].length).to eq(1)
+      expect(json["errors"].first["title"]).to eq("Document download failed")
+      expect(json["errors"].first["detail"]).to eq("Caseflow eFolder failed to fetch document contents.")
```
aroltsch: I think it is better to use `match` with regex so if you change a message slightly, you test will not fail.
For example, `to match /failed.+contents/`

app/models/document.rb
nickheiner-usds: Yes, perhaps that issue should be filed against `connect_vbms`. I didn't feel like it was useful because I didn't think that `Unable to parse SOAP message` was actually enough to debug the problem.

app/controllers/api/v1/documents_controller.rb
nickheiner-usds: Good point on `document.save_locally`.

app/controllers/api/v1/documents_controller.rb
nickheiner-usds: A few `grep`s does not turn up any usage in Reader, which is the only client I know of, so I'll remove it.

app/controllers/api/v1/documents_controller.rb
nickheiner-usds: image

app/controllers/api/v1/documents_controller.rb
nickheiner-usds: I personally don't like `unless` but I'm happy to do it if you prefer it.

app/controllers/api/v1/documents_controller.rb
mdbenjam: I'm not sure about the touch either, but what Anya said about save_locally is correct.

app/services/download_documents.rb
mdbenjam: Yeah no real idea. Maybe it's to make the download updated_at field equal to when the most recent doc was saved.... Just a guess.

spec/requests/api/v1/documents_spec.rb
nickheiner-usds: I will change this if you really want, but I'm fine with it as is. If the message changes slightly, it's trivial to copy/paste that change into the test. If we try to loosen it up with a regex, we could be letting things past like `There were no failed contents`.

app/controllers/api/v1/documents_controller.rb
mdbenjam: I don't think it's necessary.

spec/requests/api/v1/documents_spec.rb
aroltsch: Ok, I see your point!

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,25 +1,46 @@
...
+      document = Document.find(params[:id])
+    rescue ActiveRecord::RecordNotFound
+      return document_not_found
+    end
+
+    # Enable document caching for a month.
     expires_in 30.days, public: true
 
+    fetch_result = document.fetch_content!(save_document_metadata: false)
+    return document_download_failed(fetch_result[:error_kind]) if fetch_result[:error_kind]
```
aroltsch: Ok, I think I have a idea on how to simplify this. Instead of returning `error_kind`, maybe we should add more statuses to the document: 
```
enum download_status: { pending: 0, success: 1, vbms_failed: 2, vva_failed: 3 }
# method to check if document failed
def failed
  vbms_failed? || vva_failed?
end
```
Then you can check `document.failed?` here instead of `fetch_result[:error_kind]`
It will require some re-factoring but it will so much cleaner. Maybe we can do it in the next ticket. Up to you.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -1,25 +1,46 @@
...
     )
-  rescue ActiveRecord::RecordNotFound
-    document_not_found
   end
 
   private
 
+  def document_download_failed(error_kind)
+    status = 500
+    message = "Caseflow eFolder failed to fetch document contents."
```
aroltsch: Thinking about his more. Do we really need this? If will always be either vva_error or vbms_error, right?

app/models/document.rb
```diff
@@ -29,6 +29,22 @@ def filename
...
+    return {
+      content: fetcher.content(save_document_metadata: save_document_metadata),
+      error_kind: nil
+    }
+  rescue VBMS::ClientError => e
+    update_with_error "VBMS::ClientError::#{e.message}\n#{e.backtrace.join("\n")}"
+    return { content: nil, error_kind: :vbms_error }
+  rescue VVA::ClientError => e
+    update_with_error "VVA::ClientError::#{e.message}\n#{e.backtrace.join("\n")}"
+    return { content: nil, error_kind: :vva_error }
```
aroltsch: If we rely on the document status (comment above), we don't really need to return this hash. It will be cleaner.

app/services/download_documents.rb
```diff
@@ -63,19 +63,11 @@ def download_contents(save_locally: true)
...
-        update_document_with_error(document, "VVA::ClientError::#{e.message}\n#{e.backtrace.join("\n")}")
-
-      rescue ActiveRecord::StaleObjectError
-        Rails.logger.info "Duplicate download detected. Document ID: #{document.id}"
-        return false
-      end
+      fetch_result = document.fetch_content!(save_document_metadata: true)
+      document.save_locally(fetch_result[:content], index) if save_locally && !fetch_result[:error_kind]
+
+      return false if fetch_result[:error_kind] == :caseflow_efolder_error
```
aroltsch: `caseflow_efolder_error` could be another document status

app/services/download_documents.rb
```diff
@@ -63,19 +63,11 @@ def download_contents(save_locally: true)
     @download.update_attributes!(started_at: Time.zone.now)
     @download.documents.where(download_status: 0).each_with_index do |document, index|
       before_document_download(document)
-      begin
-        content = document.fetcher.content
-        document.save_locally(content, index) if save_locally
-        @download.touch
-      rescue VBMS::ClientError => e
-        update_document_with_error(document, "VBMS::ClientError::#{e.message}\n#{e.backtrace.join("\n")}")
```
aroltsch: If we add vbms_failed or vva_failed, we don't have to specify `VBMS::ClientError` in the error_message.

nickheiner-usds:
I really like that idea for a refactor. But I also think that it's a big enough refactor that it could be done in its own PR. `grep`ing around for `download_status`, I see a fair amount of usage, and I'd like to be able to focus on those changes exclusively so we don't introduce errors. Are you ok with that?

aroltsch:
I am okay with that. Would you be able to create a ticket for the refactor please.

app/services/download_documents.rb
```diff
@@ -63,19 +63,11 @@ def download_contents(save_locally: true)
...
-
-      rescue ActiveRecord::StaleObjectError
-        Rails.logger.info "Duplicate download detected. Document ID: #{document.id}"
-        return false
-      end
+      fetch_result = document.fetch_content!(save_document_metadata: true)
+      document.save_locally(fetch_result[:content], index) if save_locally && !fetch_result[:error_kind]
+
+      return false if fetch_result[:error_kind] == :caseflow_efolder_error
+      @download.touch
```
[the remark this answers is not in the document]
nickheiner-usds: I wonder that too! :smile:

nickheiner-usds:
> I am okay with that. Would you be able to create a ticket for the refactor please. 

I've updated #663 to encompass this refactor.

Issue #656, opened 2017-10-25
Optimistic locking causes error to be thrown


amprokop:
https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/?query=StaleObjectError

Sentry captures StaleObjectErrors from eFolder Express at a steady clip. These errors are caused by optimistic locking. If the same object is loaded in two different contexts, and it's saved in one context, then the other, then Rails will throw an error.

There are a number of lines of code where we observe this error:

```
app/jobs/download_manifest_job.rb in capture_error at line 36
app/jobs/download_manifest_job.rb in download_from_service at line 55
app/services/download_documents.rb in block in create_documents at line 58
app/services/download_documents.rb in block in create_documents at line 57
app/services/download_documents.rb in block (3 levels) in create_documents at line 54
app/services/download_documents.rb in block (3 levels) in create_documents at line 53
 app/controllers/downloads_controller.rb in start at line 22
```
## AC
- Errors should not be thrown by the eFolder API if a user makes multiple near-simultaneous requests to the eFolder API

aroltsch:
@amprokop can I close this issue in favor of https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/650

amprokop:
@aroltsch — ya you can close it, this is complete

PR #665, opened 2017-10-30
Catch StaleObjectError and retry

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/656

There are some deep refactors we're going to be doing to eFolder in the coming weeks, but we're currently throwing StaleObjectErrors many times a day. The refactors will fix the underlying cause of these errors, but if we catch the error and wait for the race condition to be resolved here, it should keep most users from being affected by the errors for now. This is definitely a temporary fix.

Testing: in the rails console, i manually tested that the retry loop works as expected. Otherwise, unit tests.


app/controllers/api/v1/files_controller.rb
```diff
@@ -13,6 +15,19 @@ def json_files
       download,
       each_serializer: Serializers::V1::DownloadSerializer
     ).as_json
+
+  rescue ActiveRecord::StaleObjectError
```
mdbenjam: Shouldn't we just wrap this entire function in the retry when rather than catching it first, then repeating all the code? We have it repeating four times now.

amprokop: Yep. I wanted to add the `reload` method *only* on the retry, that's why.

app/controllers/api/v1/files_controller.rb
```diff
@@ -13,6 +15,19 @@ def json_files
...
+    retry_when ActiveRecord::StaleObjectError, limit: 3 do
```
shanear: This seems like it will cause a good amount of thrashing, since the `StaleObjectError`s will keep happening when two downloads are going at the same time.

In the situation of a `StaleObjectError`, why are we trying to continue to download? It's already downloading on another thread, we just need to wait for it to finish and then return it. Something like this:

```
sleep 1 until download.reload.complete?

ActiveModelSerializers::SerializableResource.new(
  download,
  each_serializer: Serializers::V1::DownloadSerializer
).as_json
```

Seems like we should also verify that we have a timeout incase some weird situation where the download wasn't actually running in the other thread.

amprokop: or when it errors

amprokop: cool will do

app/controllers/api/v1/files_controller.rb
```diff
@@ -8,13 +8,35 @@ def index
...
+    tries = 1
+    until download.reload.all_manifests_current? || tries >= TRIES_TO_TIMEOUT do
+      sleep 2
+      tries += 1
+    end
+
+    # After we've waited the allotted number of times, let's send back
+    # what we currently have even if the manifest hasn't finished refreshing.
+    # In some cases, this will not be the refreshed list of documents,
+    # but the caller can also call the API again later.
```
shanear: awesome comments

app/controllers/api/v1/files_controller.rb
```diff
@@ -8,13 +8,35 @@ def index
...
       download,
       each_serializer: Serializers::V1::DownloadSerializer
     ).as_json
+
+  rescue ActiveRecord::StaleObjectError
+    # We expect StaleObjectErrors when a user is trying to fetch
+    # the documents list more than once simultaneously. Until we solve our underlying
+    # problem with a refactor, let's stop the API caller from receiving the
+    # error by waiting and checking if the other manifest fetch has finished.
+    tries = 1
```
shanear: very C++ 😄 , but not sure a better way

app/controllers/api/v1/files_controller.rb
```diff
@@ -8,13 +8,35 @@ def index
 
   private
 
+  TRIES_TO_TIMEOUT = 5
+
   def json_files
     download.prepare_files_for_api!(start_download: download?)
```
shanear: if `download?` resolves to `true`, then the rescue doesn't work. We should either hard code this to false, or handle both cases.

app/controllers/api/v1/files_controller.rb
shanear: i'm confused, it seems like the rescue really only applies to `download.prepare_files_for_api!(start_download: download?)`, the serializer stuff seems like it's being duplicated because the rescue block is too broad.

why not do this?
```
def json_files
  begin do
    download.prepare_files_for_api!(start_download: download?)
  rescue ActiveRecord::StaleObjectError
    # retry stuff
  end

   ActiveModelSerializers::SerializableResource.new(
      download,
      each_serializer: Serializers::V1::DownloadSerializer
    ).as_json
end
```

Issue #666, opened 2017-10-30
VBMS docs don't save yet manifest says it's fetched


mdbenjam:
On 10/30 a user reported that only VVA docs were present in a case. After examining the download we discovered that 430 docs were pulled from VBMS, the `vbms_manifest_fetched_at` was set, then the docs were added to the download in `create_documents` in `download_documents.rb`. The last line in `create_documents`: `@download.update_attributes!(manifest_fetched_at: Time.zone.now)` failed with a stale document error. Since creating the documents was in the same transaction block, we rolled all of it back. Yet since `vbms_manifest_fetched_at` was set earlier, it was still set, but the documents were not. Since we don't re-download the documents for 3 hours after the manifest was set, this user would not be able to just refresh the page to get the docs.

# AC
1. Verify that `vbms_manifest_fetched_at` and `vva_manifest_fetched_at` are only set if the corresponding document records have been created.

sunil-sadasivan:
Oh shoot this is a good one Mark!  Great catch here. Let's chat about this tomorrow. 

Truthfully I believe `manifest_fetched_at` is only used for backwards compatibility so clients don't break, I think we can remove this if we feel comfortable clients don't use this any more.

astewarttistatech:
@mdbenjam How would this be validated?

mdbenjam:
We don't have a good way of reproducing what we think caused this. So we're going to close it on the assumption that the automated test is a sufficient test case.

PR #673, opened 2017-11-01
Download object Manifest timestamps should update after documents are created

sunil-sadasivan:
This helps reduce the occurrence of a StaleObject error for the download object. 

#666


sunil-sadasivan:
🎊 thanks Mark!

PR #674, opened 2017-11-02
Remove gemserver env var

nickheiner-usds:


nickheiner-usds:
This is not necessary for our fix.

PR #680, opened 2017-11-06
Add NewRelic

nickheiner-usds:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/3842.

As noted in https://github.com/department-of-veterans-affairs/caseflow/pull/3843, we have verified that we are not sending PII or env vars. I used the same HTTP header filtering here to ensure that we do not seen PII in headers.

image

### DIFF
--- a/newrelic.yml
+++ b/newrelic.yml
@@ -1,0 +1,52 @@
+#
+# Generated November 03, 2017
+#
+# For full documentation of agent configuration options, please refer to
+# https://docs.newrelic.com/docs/agents/ruby-agent/installation-configuration/ruby-agent-configuration
+common: &default_settings
+  app_name: Caseflow eFolder
+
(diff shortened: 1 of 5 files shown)


newrelic.yml
```diff
@@ -0,0 +1,52 @@
...
+  # To avoid sending PII, we explicitly deny all headers and parameters. We then whitelist known safe attributes.
+  attributes.exclude: [response.headers.*, request.headers.*, request.parameters.*]
+  attributes.include: [
+    response.headers.contentType,
+    response.headers.contentLength,
+    request.headers.userAgent,
+    request.headers.accept,
+    request.headers.host,
+    request.headers.contentType,
+  ]
```
sunil-sadasivan: 👍

PR #692, opened 2017-11-08
Converting tiff files to pdfs

mdbenjam:
Connects: https://github.com/department-of-veterans-affairs/caseflow/issues/3782
Tech Spec: https://github.com/department-of-veterans-affairs/appeals-pm/pull/1621

Some files in VVA are TIFFs not PDFs. This change adds a check for TIFF files and converts them to PDFs before sending them to the user. Note we only upload PDFs to S3, TIFFs never get uploaded.

**Test Plan**
- [ ] Run app and download a DEMO case. 1 out of 2 files are randomly served as TIFFs, but the result should be all PDFs.


Gemfile
```diff
@@ -97,6 +99,7 @@ group :development, :test do
   gem 'bundler-audit'
 
   gem 'pry'
+  gem 'pry-byebug'
```
mdbenjam: Gives the niceness of binding.pry with debugging capabilities!

app/services/prometheus_service.rb
```diff
@@ -5,6 +5,26 @@
 class PrometheusService
   # :nocov:
   class << self
+    def image_magick_request_attempt_counter
```
mdbenjam: This is so we can get some sense of how long these files take to convert.

app/models/fetcher.rb
```diff
@@ -16,11 +16,37 @@ def cached_content
     @cached_content ||= S3Service.fetch_content(document.s3_filename)
   end
 
+  def convert_from_tiff(result)
+    MetricsService.record("Image Magick: Convert tiff to pdf",
+                          service: :image_magick,
+                          name: "convert") do
```
aroltsch: I think it is better to give more descriptive names otherwise it will just say `convert` in Grafana.

app/models/fetcher.rb
```diff
@@ -16,11 +16,37 @@ def cached_content
     @cached_content ||= S3Service.fetch_content(document.s3_filename)
   end
 
+  def convert_from_tiff(result)
```
aroltsch: I think we should put this method into a helper `Converter` to separate logic into a more reusable way.

app/models/document.rb
```diff
@@ -126,7 +126,7 @@ def self.historical_average_download_rate
   end
 
   def preferred_extension
-    mime = MIME::Types[mime_type].first
+    mime = MIME::Types[converted_mime_type || mime_type].first
```
aroltsch: Why do we need this new attribute?

app/models/fetcher.rb
```diff
@@ -16,11 +16,42 @@ def cached_content
...
+  end
+
   def download_from_service
     return cached_content if cached_content
-    external_service.fetch_document_file(document).tap do |result|
-      S3Service.store_file(document.s3_filename, result)
-    end
+
+    result = external_service.fetch_document_file(document)
+    result = convert_from_tiff(result) if document.mime_type == "image/tiff" && tiff?(result)
```
aroltsch: I would imagine something like that if we had a Converter class:

```
result = Converter.new(result, document.mime_type).process!
```

and then in the `Converter` class, the `process` method can determine if anything needs to be converted.

app/models/document.rb
mdbenjam: Without this new attribute we have the following problem.

Option 1) We leave mime_type as 'image/tiff'. This means even after converting it, we send the mime_type 'image/tiff' to the client when it's really a PDF.
Option 2) We rewrite mime_type as 'application/pdf'. This means we send the correct mime_type to the client, but once the S3 cache expires this document and we re-download it, we have no recollection that the type is really 'image/tiff' and we don't reconvert the new download.

There are probably other ways to solve this. We could maintain a list of types we convert, and never send that mime_type to the user, always map those types to pdf. I don't like this attribute either, I'm definitely open to thoughts?

aroltsch: I see. I think what you currently have is okay. However, this brings up another point: we should never expire individual documents in s3 because we know the content never changes. Certain metadata can change but it is okay, we can update that.

app/services/image_converter_service.rb
```diff
@@ -0,0 +1,55 @@
...
+  end
+
+  private
+
+  # Adding a magic number check based on this recommendation: https://imagetragick.com/
+  def tiff?
+    "MM\u0000*" == @image[0..3] || "II*\u0000" == @image[0..3]
+  end
+
+  def convert_from_tiff
```
aroltsch: could you please name it `convert_tiff_to_pdf`

app/services/image_converter_service.rb
```diff
@@ -0,0 +1,55 @@
+class ImageConverterService
+  # Converts images to PDFs
+  def initialize(image, mime_type)
```
aroltsch: If you include `include ActiveModel::Model`, you don't have to implement `initialize`. 
I think it makes it a bit cleaner. You can take a look at fetcher.rb

spec/services/image_converter_service_spec.rb
```diff
@@ -0,0 +1,41 @@
+describe ImageConverterService do
```
aroltsch: This is awesome.

app/models/document.rb
mdbenjam: Yeah it's a balance, I don't want to recreate VBMS in S3. We should store docs for as long as we reasonably think they should be used, but probably not forever. I think the new way of doing this should make it cleaner without needing this field.

nickheiner-usds:
After this PR, when I run the tests, I get the following error:

```
Failures:

  1) ImageConverterService#process when image is a tiff converts it
     Failure/Error:
       MiniMagick::Tool::Convert.new do |convert|
         convert << tiff_name
         convert << pdf_name
       end
     
     MiniMagick::Error:
       ImageMagick/GraphicsMagick is not installed
     # ./app/services/image_converter_service.rb:45:in `block in convert_tiff_to_pdf'
     # ./app/services/metrics_service.rb:12:in `block in record'
     # ./app/services/metrics_service.rb:11:in `record'
     # ./app/services/image_converter_service.rb:29:in `convert_tiff_to_pdf'
     # ./app/services/image_converter_service.rb:10:in `process'
     # ./spec/services/image_converter_service_spec.rb:18:in `block (4 levels) in <top (required)>'
```

It looks like the gem we're using requires ImageMagick or GraphicsMagick to be installed as an external dep. I'm guessing that we want the former. The "First Time Development Setup" in the README should have been amended to include this new dependency.

Issue #670, opened 2017-11-01
Moving away from the Downloads table


amprokop:
This was created as an alternate option to the refactor proposed in https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/626. It's definitely not complete or done. Thoughts and quibbles welcome. Feel free to edit.

## Context

Same as https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/626. To summarize: 

- The Downloads table is overloaded
- Race conditions related to it cause StaleObjectErrors
- UI code paths and API code paths are in a kind of Gordian knot
- We've potentially drifted away from the time when the name "Download" was good — it made sense when we were only serving single ZIP files containing lots of documents

## Overview

The refactor proposes in https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/626 gets us to a desirable place in the end, but getting there will result in a number of painful and incremental steps. We have the advantage that the data in the Downloads table has a very short lifespan, and we only need to preserve it for auditing purposes. We should consider instead creating new tables, referring to them in v2 of our API, and only backfilling data into them where necessary.

## Moving away from the Downloads table

A row in the Downloads table currently represents something like "The status of the current attempts to talk to VBMS and VVA and obtain lists of the documents in both services, as well as some accompanying denormalized data about the Veteran".  

### DocumentList

Records in this table are mutable, and represent the most recent state of our document list. Each document has a pointer to a DocumentList. The eFolder ZIP file also will have a pointer to a document list.

```
  create_table "document_lists" |t|
    t.string   "file_number"
    t.string   "source"      # (VBMS or VVA)
    t.datetime "created_at",                                      null: false
    t.datetime "updated_at",                                      null: false
    t.string   "veteran_last_name"
    t.string   "veteran_first_name"
    t.string   "veteran_last_four_ssn"
    t.integer  "zipfile_size",          limit: 8
  end 
```

### DocumentListLogs

```
  create_table "document_list_logs" |t|
    t.integer  "document_list_id"
    t.datetime "started_at"
    t.datetime "succeeded_at"
    t.datetime "errored_at"
    t.integer  "user_id"
    t.datetime "created_at",                                      null: false
  end
```

[...]

Issue #626, opened 2017-09-20
Migrate eFolder DB to split up Documents and Downloads


mdbenjam:
# Context

When working on the eFolder Express integration we uncovered several architectural decisions that made future extensibility difficult. The main obstacle was that the Download and Document tables both contained data from VBMS/VVA as well as maintained values that were specific to each instance of a user interacting with the system. When a user searches for a veteran id we look for a download record from the last three days. If such a record doesn't exist, then a new one is created. Each download object has its own document objects associated with it. This means that downloads done outside of the three days are recreated in the DB, and downloads of the same veteran id but by different users are held in different records. This means that all of the data about the download and documents from VBMS are duplicated as we try to keep track of all the fields that are specific to an individual interaction with the app such as the time it took to download and whether or not those downloads were successful.

# Overview

We propose separating these two functions in the database. The Download table will contain exactly one entry for each veteran id. The Document table will contain exactly one entry for each document in the veteran's file. Then separate log tables will maintain the specific information about every user interaction with the system.

We think this structure has three main advantages:

1) StaleObjectErrors: eFolder has had a number of StaleObjectErrors in Sentry. This is partly due to having a complex architecture with several threads all trying to access DB entries at once. However, if we separate the actual data of the entry from the audit functions, we should be able to reduce these. It's the audit side of the tables that are usually in conflict, but if we aren't limited to one Download row and instead have several audit rows then we can create new audit entries for each user's download and avoid trying to write to them at the same time.

1) Store Less Data: We are needlessly duplicating data in our DB. It seems that where possible we should reduce the number of rows in our tables.

1) Better Caching: Now that we have implemented the API endpoints we've increased the instances when two people are going to be looking at the same Download around the same time. If attorney A and judge B want to both view the same case it would be ideal for them to be able to share the same download object so that fetching the data can avoid a call to VBMS.

1) Easier Reasoning About Data: With this model we can impose uniqueness constraints on our tables. No two downloads can have the same veteran id. No two documents can have the same document id. These constraints allow us to replace fuzzier `wheres` with concrete `finds`.

# Transitioning the Downloads Table

Transitioning the Downloads table will take several steps. Consider the current schema below:
```
  create_table "downloads", force: :cascade do |t|
    t.string   "request_id"
    t.string   "file_number"
    t.integer  "status",                          default: 0
    t.datetime "created_at",                                      null: false
    t.datetime "updated_at",                                      null: false
    t.integer  "lock_version"
[...]

PR #694, opened 2017-11-09
Anya/database changes tech spec

aroltsch:
connects #696 
connects #670 
connects #626


docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+
+Tables:
+
+```
+  create_table "manifests" |t|
+    t.string   "file_number"
+    t.string   "veteran_last_name"
+    t.string   "veteran_first_name"
+    t.string   "veteran_last_four_ssn"
+    t.zipfile_size,                 # will be updated based on the changes to the documents
```
mdbenjam: What do we use this for? Do we have to save it?

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+```
+ create_table "records" do |t|
+    t.integer  "manifest_id"
+    t.integer  "status",  default: 0
+    t.string   "external_document_id"
+    t.string   "mime_type"
+    t.datetime "received_at"      # VBMS and VVA timestamp
+    t.string   "type_description"
+    t.string   "type_id"          # will be deprecated when we move to the VBMS new eFolder API
+    t.string   "vbms_filename"    # deprecate
```
mdbenjam: Yeah do we use this field? Seems not useful.

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+    t.string   "filepath"         # deprecate??
```
mdbenjam: I don't like this field. Seems like we can calculate this value.

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+    t.string   "jro"              # deprecate??
```
mdbenjam: Do you know what this is for?

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+    t.datetime "updated_at",
+  end
+```
+
+Open questions:
+
+1. Based on the discussion with multiple people, we don't need to expire `manifests`.
+   We will have `user_manifests` and `records` tables that will have all the necessary
+   information to answer questions such as "this user downloaded XYZ documents on XXX date".
+2. Do we need `searches` table?
```
mdbenjam: I don't think we need this table. In effect the `user_manifests` table is the new Search table.

docs/v2_data_model.md
aroltsch: We do need this for streaming purposes: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/downloads_controller.rb#L95

docs/v2_data_model.md
aroltsch: Yeah, we can get rid of it.

docs/v2_data_model.md
aroltsch: Exactly.

docs/v2_data_model.md
aroltsch: `jro`, `ssn` and `source` are used when getting content from VVA: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/services/external_api/vva_service.rb#L21. I know these fields are required. However, I'd like to check if if they actually check the values for these fields.

docs/v2_data_model.md
aroltsch: I agree we can get rid of it.

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+    t.string   "external_document_id"
+    t.string   "mime_type"
+    t.datetime "received_at"      # VBMS and VVA timestamp
+    t.string   "type_description"
+    t.string   "type_id"          # will be deprecated when we move to the VBMS new eFolder API
+    t.string   "vbms_filename"    # deprecate
+    t.string   "filepath"         # deprecate??
+    t.string   "jro"              # deprecate??
+    t.string   "ssn"              # deprecate??
+    t.integer  "size"             # deprecate??
```
aroltsch: I need to check why we need `size`

aroltsch: We can probably keep size to keep track of document sizes.

docs/v2_data_model.md
aroltsch: Ok, I checked, 
1. It seems as `jro` can be any value. 
2. `ssn` is the same as a file number so we can remove it
3. `source` can be one of these two values: `[P8, IS]`

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+  end
+```
+
+Open questions:
+
+1. Based on the discussion with multiple people, we don't need to expire `manifests`.
+   We will have `user_manifests` and `records` tables that will have all the necessary
+   information to answer questions such as "this user downloaded XYZ documents on XXX date".
+2. Do we need `searches` table?
+3. Using this data model, how can we migrate the existing data?
```
aroltsch: @sunil-sadasivan @mdbenjam ideas:
1) Create our new DB structure with associated models
2) Point the API to this new code
3) Eventually deprecate the UI and point whatever rises from the ashes to the API
4) Leave the old DB tables, but don’t try to migrate them

mdbenjam: That sounds like a good plan to me.

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
+Things to consider:
```
mdbenjam: I think we should add more context around this tech spec at the beginning. Why are we doing this, what do we hope to achieve. I think a lot of this is stated in the previous tech spec iterations. Feel free to copy it, but it will give someone more context if they stumble across this. I also think it makes sure we're aligned on goals.

docs/v2_data_model.md
```diff
@@ -0,0 +1,70 @@
...
+1. We will start with the simplest solution possible
+2. We will probably deprecate eX stats page so we will not worry about it at this point.
```
mdbenjam: nit: I would change the language here to be more direct. Instead of `we will probably`. Just say `we will`. And note that we'll work with Chris and others to figure out the right medium for the stats page in a different spec.

docs/v2_data_model.md
mdbenjam: I would just add the above into the spec.

aroltsch: Yup!

docs/v2_data_model.md
aroltsch: Good suggestion!

docs/v2_data_model.md
aroltsch: Ok!

Issue #697, opened 2017-11-13
[v2] Create tables and associated models


aroltsch:
Reference: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/694/files

aroltsch:
This has been merged, nothing to validate from the QA.

PR #703, opened 2017-11-14
Anya/697 v2 tables and models

aroltsch:
connects #697


db/migrate/20171113203025_create_manifests.rb
```diff
@@ -0,0 +1,13 @@
+class CreateManifests < ActiveRecord::Migration
+  def change
+    create_table :manifests do |t|
+      t.string   "file_number"
+      t.string   "veteran_last_name"
+      t.string   "veteran_first_name"
+      t.string   "veteran_last_four_ssn"
+      t.integer  "zipfile_size",  limit: 8  # The value will be updated based on the changes to the documents
```
mdbenjam: I'm wondering if we should leave this out for now. We don't use it in the API path, and I'm curious in still exploring whether there is a way to avoid putting this in the DB. This is the only thing that will need to be updated in this record.

aroltsch: I think it is fine to leave it here for now. I am pretty sure we will need to save it somewhere.

aroltsch:
Yup, I was going to add indexes in the subsequent PRs but yes it makes sense to add them as part of this PR.

PR #707, opened 2017-11-15
Add index [download_id, document_id] to documents

aroltsch:
connects https://github.com/department-of-veterans-affairs/caseflow/issues/3921#issuecomment-344699028

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -57,7 +57,6 @@ ActiveRecord::Schema.define(version: 20171115193142) do
 
   add_index "documents", ["completed_at"], name: "index_documents_on_completed_at", using: :btree
   add_index "documents", ["download_id", "document_id"], name: "index_documents_on_download_id_and_document_id", using: :btree
-  add_index "documents", ["download_id"], name: "index_documents_on_download_id", using: :btree
   add_index "documents", ["download_status"], name: "index_documents_on_download_status", using: :btree
 
   create_table "downloads", force: :cascade do |t|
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -56,6 +56,7 @@
   end
 
   add_index "documents", ["completed_at"], name: "index_documents_on_completed_at", using: :btree
+  add_index "documents", ["download_id", "document_id"], name: "index_documents_on_download_id_and_document_id", using: :btree
   add_index "documents", ["download_id"], name: "index_documents_on_download_id", using: :btree
```
shanear: now that you have a index on download_id, document_id, you could remove this index

aroltsch: I can remove the index on download_id? Even if we have queries that search by just a download_id?

amprokop:
came here to ask what @shanear did. potentially this has to be an off-hours one. this table is huuuuugee

db/schema.rb
shanear: yeah, because it's a binary tree index on the download_id at the top level, so it just wont use the secondary b-tree index on document_id.

You can see for yourself by removing the download_id index and tracing the lookup

aroltsch:
We will deploy off hours.

db/schema.rb
aroltsch: Ah okay! Makes sense!

kierachell:
Does out Rails version support: https://github.com/rails/rails/commit/2d33796457b139a58539c890624591c97354d334
? to avoid locks

PR #714, opened 2017-11-16
Modify migration to add index on documents using concurrently

aroltsch:


aroltsch:
The problem is that migration will be re-run on UAT/demo because I am modifying existing migration.

sunil-sadasivan:
Yeah we should monitor the CPU/memory as this executes at night.  This looks good to me.

Issue #669, opened 2017-11-01
EOFErrorSidekiq 


nemodjuric:
Sentry link: 
https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/535/

## AC
This error no longer appears in production.

nickheiner-usds:
Here are the relevant log entries: https://console.amazonaws-us-gov.com/cloudwatch/home?region=us-gov-west-1#logEventViewer:group=dsva-appeals-efolder_worker-prod/opt/efolder-express/src/log/efolder-express-worker.out;stream=i-0c1eb60192423571f-172.30.82.13;filter=07f018e5edf0917663b8b1d3.

nickheiner-usds:
This was the only place I could find in our gems that includes the error message:

`gems/httpclient-2.8.3/lib/httpclient/jruby_ssl_socket.rb`:
```rb
    def readpartial(size, buf = nil)
      while @bufstr.size == 0
        if fill() == -1
          raise EOFError.new('end of file reached')
        end
      end
      str = @bufstr.slice!(0, size)
      debug(str)
      if buf
        buf.replace(str)
      else
        str
      end
    end
```

nickheiner-usds:
> Artem Vovk [7 days ago] 
> It happens sometimes when their service just…drops…
>
>
> Mark Benjamin [7 days ago] 
> there's like nothing in the ticket
>
>
> Artem Vovk [7 days ago] 
> like: I'll be listing documents or something and then packets just stop
```

**Comment by nickheiner-usds** [comment]
One of the failed jobs has the following parameter:

```
"_aj_globalid": "gid://caseflow-efolder/Download/64041"
```

If I look for a download with ID `64041` in the prod database, I'll see that it was indeed downloaded at the time of the failure for that job.

**Comment by nickheiner-usds** [comment]
```
irb(main):009:0> Download.find(64041).errors?
  Download Load (0.6ms)  SELECT  "downloads".* FROM "downloads" WHERE "downloads"."id" = $1 LIMIT 1  [["id", 64041]]
[2017-11-16 15:47:03 -0500]   Download Load (0.6ms)  SELECT  "downloads".* FROM "downloads" WHERE "downloads"."id" = $1 LIMIT 1  [["id", 64041]]
[...]

PR #715, opened 2017-11-16
Do not log EOFError to Sentry

nickheiner-usds:
Connect #669.

## Test
I used a Sentry processor to trigger a breakpoint when Sentry was going to log an exception. I applied the following diff:

```diff
diff --git a/app/jobs/fakes/download_files_job.rb b/app/jobs/fakes/download_files_job.rb
index 5f51068..2d5fc5f 100644
--- a/app/jobs/fakes/download_files_job.rb
+++ b/app/jobs/fakes/download_files_job.rb
@@ -2,6 +2,8 @@ class Fakes::DownloadFilesJob < ActiveJob::Base
   queue_as :default
 
   def perform(download)
+    puts "~~~~~ ++++++== ================ raising error fak ============================="
+    raise 'fak'
     demo = Fakes::DocumentService::DEMOS[download.file_number] || Fakes::DocumentService::DEMOS["DEMODEFAULT"]
 
     Fakes::DocumentService.errors = demo[:error]
diff --git a/config/initializers/sentry.rb b/config/initializers/sentry.rb
index 66f61c5..3e113f9 100644
--- a/config/initializers/sentry.rb
+++ b/config/initializers/sentry.rb
@@ -1,9 +1,17 @@
+class MyJobProcessor < Raven::Processor
+  def process(data)
+    binding.pry
+  end
+end
+
 if ENV['SENTRY_DSN']
   Raven.configure do |config|
     config.dsn = ENV['SENTRY_DSN']
-
+    
     # Sometimes, a VA backend will randomly fail. When we retry, it generally works.
     # We do not need to log this to Sentry, because it's not actionable for us.
     config.excluded_exceptions += ['EOFError']
+
+    config.processors += [MyJobProcessor]
   end
 end

```

I ran the following command:

```
$ SENTRY_SECRET_KEY=fake SENTRY_DSN=the_actual_value bundle exec sidekiq -v
```

By modifying the diff above, I was able to validate:

1. Without this PR, `EOFError`s are sent to Sentry.
1. With this PR, `EOFError`s are no longer sent to Sentry.
1. With this PR, Other errors are still sent to Sentry.

### DIFF
--- a/app/jobs/download_files_job.rb
+++ b/app/jobs/download_files_job.rb
@@ -1,6 +1,14 @@
 class DownloadFilesJob < ActiveJob::Base
   queue_as :default
 
+  # Sometimes, a VA backend will randomly fail. When we retry, it generally works.
+  # We will explicitly rescue and retry here, because otherwise the first
+  # failure will be logged to Sentry. For this error, we only want the last one
+  # to be logged to Sentry.
+  rescue_from EOFError do
+    retry_job
+  end
+
   def perform(download)
     download_documents = DownloadDocuments.new(download: download)
 
--- a/app/jobs/download_files_job.rb
+++ b/app/jobs/download_files_job.rb
@@ -1,14 +1,6 @@
 class DownloadFilesJob < ActiveJob::Base
   queue_as :default
 
-  # Sometimes, a VA backend will randomly fail. When we retry, it generally works.
-  # We will explicitly rescue and retry here, because otherwise the first
-  # failure will be logged to Sentry. For this error, we only want the last one
-  # to be logged to Sentry.
-  rescue_from EOFError do
-    retry_job
-  end
-
   def perform(download)
     download_documents = DownloadDocuments.new(download: download)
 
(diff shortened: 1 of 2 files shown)


app/jobs/download_files_job.rb
```diff
@@ -1,6 +1,14 @@
 class DownloadFilesJob < ActiveJob::Base
   queue_as :default
 
+  # Sometimes, a VA backend will randomly fail. When we retry, it generally works.
+  # We will explicitly rescue and retry here, because otherwise the first
+  # failure will be logged to Sentry. For this error, we only want the last one
+  # to be logged to Sentry.
```
shanear: > For this error, we only want the last one to be logged to Sentry.

Does this code do this?

nickheiner-usds: Yeah, I was just coming here to leave that comment. As I understand it, this code will actually retry indefinitely when we're getting `EOFError`, which is not desirable. I was not able to find something like `attempt_number` earlier, so I am not sure how we can only retry for the first `n` times.

Issue #569, opened 2017-08-02
Issues with email address and CSS ID


aroltsch:
1. Email address have spaces in front of them. E.g, email: "  Tecola.Plowden@va.gov"
2. We will not worry about converting css_ids to upper case in our DB. We need to think about CSS IDs more. Previously, all CSS IDs were upper case then at some point in time some users CSS IDs were changed to include a combination of upper/lower case. Was it a one time change for the existing users? If yes, once we clean up the duplicates, we will not run into the same problem again.

AC:
1. Ensure emails don't have spaces. Create a migration to clean up existing emails
2. Write a migration to clean up duplicate user records, use the latest css_id for user
3. Test both migrations on the DB prod copy

amprokop:
@aroltsch — would you mind giving some context on this issue? I think you said that FeatureToggles didn't work properly because of these, are there other problems that were created?

aroltsch:
@amprokop It causes a few issues:
1. Emails have empty spaces - makes it more difficult to search for a user by email
2. We have duplicate User records for the same user due to CSS ID discrepancies: we might have up to 3 user records with the same css id, the only difference is upper/lower case letters. 
3. Problems with a feature toggle due to upper/lower case letter combinations in CSS IDs: https://github.com/department-of-veterans-affairs/caseflow-commons/issues/111

lowellnava:
@aroltsch two questions:

1. Do we want to remove all spaces from an email address, including spaces in the username/local part (i.e. "king zombo@zombo.com" would become "kingzombo@zombo.com"), or do we want to remove only leading and trailing whitespace from the email address ("       king zombo@zombo.com " would become "king zombo@zombo.com")?

2. When we store the CSS IDs in the database do we want to convert all characters to uppercase (or lowercase)? Or do we want to retain the casing of the CSS ID and ensure the CSS ID is unique regardless of case (perhaps by using [validates_uniqueness_of()](http://api.rubyonrails.org/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of) with :case_sensitive set to false)?

lowellnava:
@aroltsch Another question, should Users' CSS_IDs be unique by themselves, or should only the combination of CSS_ID and station_ID be unique?

aroltsch:
@lowellnava 
1. I am pretty sure VA email addresses don't have spaces in the middle so yes only leading and trailing whitespace.
[...]

PR #719, opened 2017-11-17
Normalize user email address and css_id

lowellrex:
Connects #569.

Trims leading and trailing whitespace from email addresses when creating a new User object. Combining existing records will be left for a future ticket.

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -60,7 +60,7 @@ class Api::V1::ApplicationController < BaseController
   end
 
   def css_id
-    request.headers["HTTP_CSS_ID"].try(:upcase)
+    request.headers["HTTP_CSS_ID"]
   end
 
   def capture_error(e)
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -60,7 +60,7 @@ class Api::V1::ApplicationController < BaseController
   end
 
   def css_id
-    request.headers["HTTP_CSS_ID"]
+    request.headers["HTTP_CSS_ID"].try(:upcase!)
   end
 
   def capture_error(e)
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -60,7 +60,7 @@ class Api::V1::ApplicationController < BaseController
   end
 
   def css_id
-    request.headers["HTTP_CSS_ID"].try(:upcase!)
+    request.headers["HTTP_CSS_ID"].try(:upcase)
   end
 
   def capture_error(e)
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -48,7 +48,7 @@ class User < ActiveRecord::Base
     def from_session(session, request)
       return nil unless session["user"]
       user = session["user"]
-      find_or_create_by(css_id: user["css_id"], station_id: user["station_id"]).tap do |u|
+      find_or_create_by(css_id: user["css_id"].try(:upcase), station_id: user["station_id"]).tap do |u|
         u.name = user["name"]
         u.email = user["email"]
         u.roles = user["roles"]
@@ -64,7 +64,7 @@ class User < ActiveRecord::Base
 
       {
         id: auth_hash.uid,
-        css_id: auth_hash.uid,
+        css_id: auth_hash.uid.try(:upcase),
         email: raw_css_response["http://vba.va.gov/css/common/emailAddress"],
         name: "#{first_name} #{last_name}",
         roles: raw_css_response.attributes["http://vba.va.gov/css/caseflow/role"],
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -73,8 +73,8 @@ class User < ActiveRecord::Base
     end
 
     def find_or_create_by(args)
-      args[:css_id].try(:upcase!)
-      args[:email].try(:strip!)
+      args[:css_id] = args[:css_id].try(:upcase)
+      args[:email] = args[:email].try(:strip)
       super
     end
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -7,8 +7,8 @@ class User < ActiveRecord::Base
   has_many :manifests, through: :user_manifests
 
   after_initialize do |u|
-    u.email.try(:strip!)
-    u.css_id.try(:upcase!)
+    u.email.try(:strip!) if new_record?
+    u.css_id.try(:upcase!) if new_record?
   end
 
   NO_EMAIL = "No Email Recorded".freeze
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,6 +6,11 @@ class User < ActiveRecord::Base
   has_many :user_manifests
   has_many :manifests, through: :user_manifests
 
+  after_initialize do |u|
+    u.email.try(:strip!)
+    u.css_id.try(:upcase!)
+  end
+
   NO_EMAIL = "No Email Recorded".freeze
 
   attr_accessor :name, :roles, :ip_address
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,10 +6,10 @@ class User < ActiveRecord::Base
   has_many :user_manifests
   has_many :manifests, through: :user_manifests
 
-  after_initialize do |u|
-    u.email.try(:strip!) if new_record?
-    u.css_id.try(:upcase!) if new_record?
-  end
+  validates :css_id, format: {
+    with: /\b[^a-z]+\b/,
+    message: "CSS_IDs cannot contain lowercase letters to avoid duplicate User records with differently cased CSS_IDs"
+  }
 
   NO_EMAIL = "No Email Recorded".freeze
 
@@ -48,7 +48,7 @@ class User < ActiveRecord::Base
     def from_session(session, request)
       return nil unless session["user"]
       user = session["user"]
-      find_or_create_by(css_id: user["css_id"].try(:upcase), station_id: user["station_id"]).tap do |u|
+      find_or_create_by(css_id: user["css_id"], station_id: user["station_id"]).tap do |u|
         u.name = user["name"]
         u.email = user["email"]
         u.roles = user["roles"]
@@ -64,12 +64,18 @@ class User < ActiveRecord::Base
 
       {
         id: auth_hash.uid,
-        css_id: auth_hash.uid.try(:upcase),
+        css_id: auth_hash.uid,
         email: raw_css_response["http://vba.va.gov/css/common/emailAddress"],
         name: "#{first_name} #{last_name}",
         roles: raw_css_response.attributes["http://vba.va.gov/css/caseflow/role"],
         station_id: raw_css_response["http://vba.va.gov/css/common/stationId"]
       }
+    end
+
+    def find_or_create_by(args)
+      args[:css_id].try(:upcase!)
+      args[:email].try(:strip!)
+      super
     end
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,10 +6,9 @@ class User < ActiveRecord::Base
   has_many :user_manifests
   has_many :manifests, through: :user_manifests
 
-  validates :css_id, format: {
-    with: /\b[^a-z]+\b/,
-    message: "CSS_IDs cannot contain lowercase letters to avoid duplicate User records with differently cased CSS_IDs"
-  }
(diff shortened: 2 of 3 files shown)


spec/models/user_spec.rb
```diff
@@ -40,6 +40,20 @@
     end
   end
 
+  context "validation on object instantiation" do
+    let(:email) { "   email_address with leading_and_trailing_whitespace@zombo.com  " }
+    let(:css_id) { "lowercase_string" }
+    subject { User.new(email: email, css_id: css_id) }
+
+    it "trims leading and trailing whitespace from User's email address" do
+      expect(subject.email).to eq email.strip
```
aroltsch: I think it is better to test against the hard-coded value so we know exactly what we expect. For example, 
```
expect(subject.email).to eq "email_address with leading_and_trailing_whitespace@zombo.com"
```

app/models/user.rb
```diff
@@ -43,7 +48,7 @@ class << self
     def from_session(session, request)
       return nil unless session["user"]
       user = session["user"]
-      find_or_create_by(css_id: user["css_id"], station_id: user["station_id"]).tap do |u|
+      find_or_create_by(css_id: user["css_id"].try(:upcase), station_id: user["station_id"]).tap do |u|
```
shanear: Since we're upcasing at the source in `from_css_auth_hash` we shouldn't need this for long (except maybe for a couple days while peoples sessions are residual from the old way).

Want to add a note or issue to take it out?

app/models/user.rb
```diff
@@ -6,6 +6,11 @@ class User < ActiveRecord::Base
   has_many :user_manifests
   has_many :manifests, through: :user_manifests
 
+  after_initialize do |u|
+    u.email.try(:strip!) if new_record?
```
shanear: seems like this would be more appropriate in `from_css_auth_hash`

app/models/user.rb
```diff
@@ -6,6 +6,11 @@ class User < ActiveRecord::Base
...
+    u.css_id.try(:upcase!) if new_record?
```
shanear: we already basically achieve this in `from_css_auth_hash`

app/controllers/api/v1/application_controller.rb
```diff
@@ -60,7 +60,7 @@ def station_id
   end
 
   def css_id
-    request.headers["HTTP_CSS_ID"]
+    request.headers["HTTP_CSS_ID"].try(:upcase)
```
shanear: where is this coming from? I feel like we shouldn't be slamming upcase everywhere we reference `css_id` 😄

spec/models/user_spec.rb
lowellnava: I took a little different tack on the approach (added the whitespace to an existing variable), but I think I achieved what you were looking for in order to make the test more explicit. Let me know how it looks!

app/controllers/api/v1/application_controller.rb
lowellnava: I think we need to uppercase the css_id here because on [line 52](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/719/files/96a4607baf5fffb11ac8fc21740a1076586a206d#diff-c8e638129123d4606cfda05f4df8bf8bR52) we use the css_id to instantiate a new User object through `find_or_create_by`.

app/models/user.rb
lowellnava: I added this logic in the model itself because we currently instantiate User objects in three different places (application_controller, base_controller, and sessions_controller) in three different ways (`find_or_create_by`, `from_session`, and `from_css_auth_hash`) and I can foresee us adding additional ways to instantiate users, and I want to protect us from the situation where we forget to uppercase the css_id at that point in time. I'm open to ideas about other ways we could consolidate this logic though, since I agree that this implementation is far from ideal.

shanear: overwriting the setter seems good? Then take out the upcase everywhere else?

```
def css_id=(css_id)
  write_attribute(:css_id, value && value.upcase!)
end
```

app/models/user.rb
lowellnava: The base_controller calls `from_session()` directly in `current_user()`. Is the `current_user()` function currently not used (is the `current_user` variable always set by the application_controller's `authenticate_or_authorize()` function?), or will it soon be replaced by something else?

aroltsch: Would it make sense to override `find_or_create_by` in User.rb?

app/models/user.rb
lowellnava: I don't think that overwriting the setter will do the trick. Quick test with the code above:
```
# Insert a user with a mixed-case css_id in the database:
psql> insert into users (css_id, station_id, created_at, updated_at) values ('someString', 999, current_timestamp, current_timestamp);

# Overwriting the setter does not take place before the search, as shown
# by searching for a User with a mixed-case css_id:
rails c> u = User.find_or_create_by(css_id: 'someString', station_id: 999)
  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."css_id" = $1 AND "users"."station_id" = $2 LIMIT 1  [["css_id", "someString"], ["station_id", "999"]]
[2017-11-20 14:17:41 -0500]   User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."css_id" = $1 AND "users"."station_id" = $2 LIMIT 1  [["css_id", "someString"], ["station_id", "999"]]
=> #<User id: 13, css_id: "someString", station_id: "999", email: nil, created_at: "2017-11-20 14:16:57", updated_at: "2017-11-20 14:16:57", vva_coachmarks_view_count: 0>

# However, it does take place before the insert as shown by searching
# for a User that does not exist (css_id does not strictly match) with a
# mixed-case css_id that will be transformed into all uppercase.
rails c> v = User.find_or_create_by(css_id: 'SOMEString', station_id: 999)
  User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."css_id" = $1 AND "users"."station_id" = $2 LIMIT 1  [["css_id", "SOMEString"], ["station_id", "999"]]
[2017-11-20 14:19:00 -0500]   User Load (0.3ms)  SELECT  "users".* FROM "users" WHERE "users"."css_id" = $1 AND "users"."station_id" = $2 LIMIT 1  [["css_id", "SOMEString"], ["station_id", "999"]]
   (0.1ms)  BEGIN
[2017-11-20 14:19:00 -0500]    (0.1ms)  BEGIN
  SQL (0.3ms)  INSERT INTO "users" ("css_id", "station_id", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["css_id", "SOMESTRING"], ["station_id", "999"], ["created_at", "2017-11-20 19:19:00.437298"], ["updated_at", "2017-11-20 19:19:00.437298"]]
[2017-11-20 14:19:00 -0500]   SQL (0.3ms)  INSERT INTO "users" ("css_id", "station_id", "created_at", "updated_at") VALUES ($1, $2, $3, $4) RETURNING "id"  [["css_id", "SOMESTRING"], ["station_id", "999"], ["created_at", "2017-11-20 19:19:00.437298"], ["updated_at", "2017-11-20 19:19:00.437298"]]
   (2.0ms)  COMMIT
[2017-11-20 14:19:00 -0500]    (2.0ms)  COMMIT
=> #<User id: 14, css_id: "SOMESTRING", station_id: "999", email: nil, created_at: "2017-11-20 19:19:00", updated_at: "2017-11-20 19:19:00", vva_coachmarks_view_count: 0>

File: app/models/user.rb (hunk 2 above)
Comment (lowellnava) [reply app/models/user.rb]: If we move the uppercasing logic into `find_or_create_by` we would still need to have some logic to uppercase `css_id` in `from_css_auth_hash` because we instantiate a new User object without creating the associated record in the database (and triggering all the associated ActiveRecord callbacks along the way).

This actually might be fine because as far as I can tell (might be totally wrong about this), the only place we use the "user" element of the session is back in the base_controller where we use the session to create a record in the database.

Would it be possible to just create the database record in `from_css_auth_hash()`? This would allow us to get rid of `from_session()` entirely (in reality that logic would move to `from_css_auth_hash()`) and have both ways we instantiate User objects would then use `find_or_create_by()`?

Comment (lowellnava) [reply app/models/user.rb]: Chatted with @aroltsch and it seems like creating the user record in `from_css_auth_hash()` is not the way we want to go. I like Anya's suggestion of overriding `find_or_create_by()` for the user (and adding some validation to the model to protect ourselves if we forget about this change in the future). Will strip out the excessive `upcase()` calls and whip up a new PR.

File: app/models/user.rb [hunk 6]
```diff
@@ -66,5 +71,11 @@ def from_css_auth_hash(auth_hash)
         station_id: raw_css_response["http://vba.va.gov/css/common/stationId"]
       }
     end
+
+    def find_or_create_by(args)
```
aroltsch: @lowellrex  thinking about it more, I am not sure I am 100% for this solution. I think it makes more sense to override the setter instead or implement `before_save :uppercase_css_id`. Once we merge the PR, we might create duplicates, but I think it is okay because we have a migration to clean this up.

lowellrex: My concern with going that route is that we would create duplicates after this is merged as well. Consider the example where the User record with css_id "USER_A" is input and somebody attempts to log in with css_id "user_a" (and the same station_id). The find stage of `find_or_create_by()` will fail to find a matching record so we will attempt to create the record. Overwriting the setter or implementing a `before_save` callback will result in the css_id being uppercased and we will end up with two rows in the user table with the same css_id and station_id.

As far as I can tell there are two ways to avoid ending up in this state. First, we could make the combination of css_id and station_id into a unique key (it is currently an index but not a unique key...) or add validation to enforce uniqueness in the model. However, we would need to add some error handling around our calls to `find_or_create_by()` because the create stage of that function will raise an error when the duplicate record fails validation/ database uniqueness constraint.

The second way to avoid this is to uppercase the css_id before we search for existing User records (we can also add the uniqueness validation/constraint with this approach). One way to do that would be to overwrite `find_or_create_by()` (the current approach). Another way to do that would be to uppercase the css_id before we call `find_or_create_by()` (the previous approach that resulting in `upcase()` being peppered throughout the code).

Are there other approaches to uppercasing the css_id before we search for existing records in the database? Are there other ways to prevent duplicates with the same correctly capitalized css_id from ending up in the database (we could have a cronjob run the series of sql statements every hour or something to clean up after ourselves)?

app/models/user.rb
```diff
@@ -66,5 +70,36 @@ def from_css_auth_hash(auth_hash)
...
+    # abstract class that ActiveRecord::Base implements, but then I still do not fully understand how rails does all
+    # of its magic.
+    #
+    # Link to ActiveRecord::Relation's find_or_create_by() implementation:
+    # https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by
+    #
+    # One other thought is that our reliance on hijacking a library's internal implementation details to accomplish
+    # our objective (that is, this fix relies on the ActiveRecord::Relation.find_or_create_by() calling find_by() and
+    # will fail should that internal implementation detail change this fix will fail to function) may indicate that
+    # we are solving this in a less-than-ideal fashion and we should consider better/easier/more direct solutions.
```
lowellrex: (Duplicating this note in a comment for ease of reading)

Re-implement [ActiveRecord's find_or_create_by()](https://apidock.com/rails/v4.0.2/ActiveRecord/Relation/find_or_create_by) method in order to properly call find_by() from within find_or_create_by() since ActiveRecord::Relation.find_or_create_by() (by way of ActiveRecord::Base?) is not calling User.find_by() as I expected it would. Perhaps this is because of some way ruby/ActiveRecord handles inheritance and child class method overriding? It does not appear as if ActiveRecord::Relation is an abstract class that ActiveRecord::Base implements, but then I still do not fully understand how rails does all of its magic.

One other thought is that our reliance on hijacking a library's internal implementation details to accomplish our objective (that is, this fix relies on the ActiveRecord::Relation.find_or_create_by() calling find_by() and will fail should that internal implementation detail change this fix will fail to function) may indicate that we are solving this in a less-than-ideal fashion and we should consider better/easier/more direct solutions.

lowellrex:
There is a consensus that we are unhappy with all of the solutions we explored in this PR. The issue that generated this PR (#569) is minor enough that it does not warrant such significant changes to the codebase. We will accomplish the data transformations that the issue asks for in a different PR and that should be a good enough solution for the time being.

Additionally, should we ever revisit this idea (to do this protection in the code itself), I recommend using unique, case-insensitive indexes on the database itself (though I guess this could be done through rails too?):
```
create unique index on users (upper(css_id), station_id);
```

PR #722, opened 2017-11-18
Update database.yml to allow env params to configure db connection

sunil-sadasivan:
I use postgres and redis in docker containers locally.  We should have a way for devs to override the default local postgres server. 

This adds a way to override the rails db based on environment variables set.


config/database.yml
```diff
@@ -12,7 +12,11 @@ default: &default
 
 development:
   <<: *default
-  username: <%= `whoami` %>
+  host: <%= ENV['POSTGRES_HOST'] %>
+  username: <%= ENV['POSTGRES_USER'] || `whoami` %>
+  <% if ENV['POSTGRES_PASSWORD'] %>
```
aroltsch: I am not sure we need this `if` statement. If `ENV['POSTGRES_PASSWORD']` is nil, it would just set `nil` for password.

config/database.yml
```diff
@@ -12,7 +12,11 @@ default: &default
...
+  password: <%= ENV['POSTGRES_PASSWORD'] %>
+  <% end %>
   database: caseflow_efolder_development
 
 staging:
```
aroltsch: Would we need the same for staging?

sunil-sadasivan:
Hey @aroltsch, apologies, this got off my radar for a while!  

I just made an adjustment to remove the if statements.  

I am unsure where we use the `staging` environment.  @kierachell mentions that we do not quite use the staging environment.  I'm a bit hesitant to adjust the staging env for consistency if I'm unable to test this path out.

sunil-sadasivan:
I've also simplifed this to ensure local environments have shared local default config.

PR #724, opened 2017-11-19
Adds toggle for forward proxy

amprokop:


app/services/external_api/vbms_service.rb
```diff
@@ -42,31 +42,10 @@ def self.fetch_document_file(document)
...
-
-    config
-  end
-
   def self.init_client
-    return VBMS::Client.from_env_vars(logger: RailsVBMSLogger.new,
-                                      env_name: ENV["CONNECT_VBMS_ENV"]
-                                     ) if Rails.application.secrets.vbms["env"]
-
-    VBMS::Client.new(
```
amprokop: Looks to me like we transitioned to creating the vbms client using env vars so this can be removed. Feel free to revert if not.

Gemfile
```diff
@@ -76,7 +76,7 @@ gem "mini_magick"
 gem 'connect_vbms', git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "2e401bf7749f0780280951cbde777b7d67765554"
 gem 'connect_vva', git: "https://github.com/department-of-veterans-affairs/connect_vva.git", ref: "55ae8fa0a526a879818ce5835ae5384c55d3c64c"
 gem 'bgs', git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git", :branch => 'master'
```
amprokop: we probably shouldn't be pointing directly to master here. when we merge https://github.com/department-of-veterans-affairs/ruby-bgs/pull/44 update this line

app/services/external_api/vbms_service.rb
kierachell: 🗑 cleaning up!

PR #725, opened 2017-11-20
Add migration to deduplicate existing user data

lowellrex:
Connects #569. 

See [this comment](https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/569#issuecomment-345375868) and the ones following it on the issue ticket for additional background.

### DIFF
--- a/db/migrate/20171120161557_deduplicate_user_records.rb
+++ b/db/migrate/20171120161557_deduplicate_user_records.rb
@@ -1,11 +1,6 @@
 class DeduplicateUserRecords < ActiveRecord::Migration
   # rubocop:disable Metrics/MethodLength
   def up
-    # Run these statements as part of a single transaction for safety and thoroughness (we could potentially add
-    # a new user after the delete operation took place and capitalizing each letter would then cause a primary
-    # key conflict). Also, this transaction will probably be very speedy so any locking effects will be small.
-    execute "begin;"
-
     # Update css_id field in downloads table to reference older version of duplicate user.
     execute <<-SQL
       with t as (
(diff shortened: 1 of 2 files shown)


db/migrate/20171120161557_deduplicate_user_records.rb
```diff
@@ -0,0 +1,73 @@
...
+        from users a
+        join users b
+          on upper(a.css_id) = upper(b.css_id)
+          and a.station_id = b.station_id
+          and a.created_at < b.created_at
+      )
+      update downloads
+      set user_id = t.original_id
+      from t
+      where user_id = t.duplicate_id;
```
shanear: i don't think you need this. but it doesn't really matter that it's there either 😄

db/migrate/20171120161557_deduplicate_user_records.rb
```diff
@@ -0,0 +1,73 @@
...
+          and a.station_id = b.station_id
+          and a.created_at < b.created_at
+      )
+      update searches
+      set user_id = t.original_id
+      from t
+      where user_id = t.duplicate_id;
+    SQL
+
+    # Delete newer duplicate user records.
```
shanear: It'd be nice to validate that there are in fact no searches or downloads referencing the users we are deleting. Just to be safe.

kierachell: Something like:
```

with t as (
select
  a.id original_id,
  b.id duplicate_id,
  a.css_id css_id1,
  b.css_id css_id2
from users a
join users b
on upper(a.css_id) = upper(b.css_id)
  and a.station_id = b.station_id
  and a.created_at < b.created_at)
select searches.id, file_number, duplicate_id from searches
join t on searches.user_id = t.original_id
union
select searches.id, file_number, original_id from searches
join t on searches.user_id = t.duplicate_id
limit 100;
``` 
??

PR #731, opened 2017-11-22
Trim whitespace from around user email addresses

lowellrex:
Connects #569.


aroltsch:
@lowellrex would you also have another PR to trim white spaces on the `user` save?

lowellrex:
@aroltsch Just an oversight, added to this PR

PR #732, opened 2017-11-22
Records need to be associated with the manifest source table

aroltsch:
Update to the v2 data model. We need to connect `records` to `manifest_sources` so we know where the document came from.

Reference: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/database_structure.md

### DIFF
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -1,7 +1,7 @@
 class Manifest < ActiveRecord::Base
-  has_many :manifest_statuses
+  has_many :sources, class_name: "ManifestSource"
   has_many :user_manifests
-  has_many :records
+  has_many :records, through: :sources
 
   validates :file_number, presence: true, uniqueness: true
 end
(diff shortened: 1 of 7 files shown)


app/models/manifest.rb
```diff
@@ -1,7 +1,7 @@
 class Manifest < ActiveRecord::Base
-  has_many :manifest_statuses
+  has_many :sources, class_name: "ManifestSource"
   has_many :user_manifests
-  has_many :records
+  has_many :records, through: :sources
```
tejans24: would this still work if ManifestSource doesn't have any association to Records?

app/models/record.rb
```diff
@@ -1,6 +1,6 @@
 class Record < ActiveRecord::Base
-  belongs_to :manifest
+  belongs_to :manifest_source
```
tejans24: ManifestSource does need a association to record then?

aroltsch: Yes!

app/models/manifest.rb
aroltsch: It will not work!

Issue #701, opened 2017-11-14
[v2] Connect API ManifestsController to the v2 data model


aroltsch:
tech notes: 
1. Create a new file called `controllers/api/v2/manifests_controller.rb`
2. Connect API::V2::ManifestsController to the new data model: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/database_structure.md

# AC:
Ensure the new API endpoint `api/v2/manifests` works the same as `/api/v1/files`

# Steps to Validate:
1.
```Curl the API with curl "http://localhost:3000/api/v2/manifests" -H "Authorization: Token token=token" -H "css_id: CF_Q_283" -H "station_id: 283" -H "file_number: 1234"```
2. Ensure statuses are correct
3. Ensure the record list is accurate

**Comment by kierachell** [comment]
### you mean `file-number` header, not `veteran-id`
### Why does this 500 result in a `forbidden record` response?
```
SOAP response (status 500)                       dsva-appeals-efolder-uat/opt/efolder-express/src/log/efolder-express.out i-08adbc13a11592d98-172.30.66.34 D, [2017-12-03T17:20:03.191427 #2902] DEBUG -- : <?xml version='1.0' encoding='UTF-8'?><S:Envelope
 xmlns:S="http://schemas.xmlsoap.org/soap/envelope/"><S:Header><work:WorkContext xmlns:work="http://oracle.com/weblogic/soap/workarea/">rO0ABXdUAB13ZWJsb2dpYy5hcHAuQ29ycG9yYXRlRGF0YUVBUgAAANYAAAAjd2VibG9naWMud29ya2FyZWEuU3RyaW5nV29ya0NvbnRleHQABjEwLjIuMAAA</work:WorkContext></S:Header><S:Body><S:Fault xmlns:ns4="http://www.w3.org/2003/05/soap-envelope"><faultcode>S:Server</faultcode><faultstring>findFlash
es: Invalid Input</faultstring><detail><ns2:ShareException xmlns:ns2="http://services.share.benefits.vba.va.gov/"><message>findFlashes: FileNumber is empty</message></ns2:ShareException></detail></S:Fault></S:Body></S:Envelope>
[...]

PR #737, opened 2017-11-27
Anya/701 manifests controller v2

aroltsch:
connects #701 

This is a working version of `/api/v2/manifests` connected to the v2 data model.
In the next PR:
1. Write automated tests
2. Implement or create tickets for all TODO comments

Reference: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/database_structure.md

### DIFF
--- a/app/jobs/v2/download_manifest_job.rb
+++ b/app/jobs/v2/download_manifest_job.rb
@@ -1,0 +1,27 @@
+class V2::DownloadManifestJob < ActiveJob::Base
+  queue_as :default
+
+  around_perform :set_vbms_version
+
+  def perform(service, manifest_source)
+    documents = service.fetch_documents_for(manifest_source.manifest)
+    DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
+  # TODO: make a parent class for these two errors
+  rescue VBMS::ClientError, VVA::ClientError => e
+    Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+    Raven.capture_exception(e)
+  end
+
+  private
+
+  # Use VBMS efolder new API
+  def set_vbms_version
+    FeatureToggle.enable!(:vbms_efolder_service_v1)
+    yield
+    FeatureToggle.disable!(:vbms_efolder_service_v1)
+  end
+
+  def max_attempts
+    1
+  end
+end
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -4,4 +4,49 @@ class Manifest < ActiveRecord::Base
   has_many :records, through: :sources
 
   validates :file_number, presence: true, uniqueness: true
+
+  def start!
+    # TODO: create UserManifest object
+    # TODO: can we do it in parallel
+    vbms_source.start!
+    vva_source.start!
+  end
+
+  def vbms_source
+    sources.find_or_create_by(source: "VBMS")
+  end
+
+  def vva_source
+    sources.find_or_create_by(source: "VVA")
+  end
+
+  # If we do not yet have the veteran_first_name saved in Caseflow's DB, then
+  # we want to fetch it from BGS, save it to the DB, then return it
+  # TODO: these 3 methods are identical, let's simplify this
+  def veteran_first_name
+    super || begin
+      update_attributes(veteran_first_name: veteran.first_name || "") if veteran
+      super
+    end
+  end
+
+  def veteran_last_name
+    super || begin
+      update_attributes(veteran_last_name: veteran.last_name || "") if veteran
+      super
+    end
+  end
+
+  def veteran_last_four_ssn
+    super || begin
+      update_attributes(veteran_last_four_ssn: veteran.last_four_ssn || "") if veteran
+      super
+    end
+  end
+
+  private
+
+  def veteran
+    @veteran ||= Veteran.new(file_number: file_number).load_bgs_record!
+  end
 end
--- a/app/models/manifest_source.rb
+++ b/app/models/manifest_source.rb
@@ -9,4 +9,29 @@ class ManifestSource < ActiveRecord::Base
   has_many :records
 
   validates :manifest, :source, presence: true
+  validates :manifest, uniqueness: { scope: :source }
+  validates :source, inclusion: { in: %w(VBMS VVA) }
+
+  def start!
+    return if success? && fetched_at && fetched_at > 3.hours.ago
+
+    # TODO: what do we do if it is pending, hmmmm
+    # If it is pending, don't make another request
+    # https://bibwild.wordpress.com/2013/06/03/activerecord-atomic-check-and-update-through-optimistic-locking/
+    update(status: :pending)
+    V2::DownloadManifestJob.perform_now(service, self)
+    update(status: :success, fetched_at: Time.zone.now)
+
+    # TODO: handle error
+    # TODO: start downloading records from VBMS and save to S3
+  end
+
+  def service
+    case source
+    when "VBMS"
+      VBMSService
+    when "VVA"
+      VVAService
+    end
+  end
 end
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -3,4 +3,19 @@ class Record < ActiveRecord::Base
 
   validates :manifest_source, :external_document_id, presence: true
   validates :external_document_id, uniqueness: true
+
+  def self.create_from_external_document(manifest_source, document)
+    # TODO: add DB index for [manifest_source_id, external_document_id]
+    find_or_initialize_by(manifest_source: manifest_source, external_document_id: document.document_id).tap do |t|
+      t.assign_attributes(
+        type_id: document.type_id,
+        type_description: document.type_description,
+        mime_type: document.mime_type,
+        received_at: document.received_at,
+        jro: document.jro,
+        source: document.source
+      )
+      t.save!
+    end
+  end
 end
--- a/app/models/serializers/v2/manifest_serializer.rb
+++ b/app/models/serializers/v2/manifest_serializer.rb
@@ -1,0 +1,19 @@
+class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
+  type :manifest
+
+  attribute :veteran_first_name
+  attribute :veteran_last_name
+  attribute :created_at
+  attribute :updated_at
+
+  attribute :records do
+    object.records.order(:id).map do |document|
+      {
+        id: document.id,
+        type_id: document.type_id,
+        received_at: document.received_at,
+        external_document_id: document.external_document_id
+      }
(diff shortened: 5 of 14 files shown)


app/jobs/v2/download_manifest_job.rb
```diff
@@ -0,0 +1,27 @@
...
+    DocumentCreator.new(manifest_source: manifest_source).create(documents)
+  # TODO: make a parent class for these two errors
+  rescue VBMS::ClientError, VVA::ClientError, e
+    Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+    Raven.capture_exception(e)
+  end
+
+  private
+
+  # Use VBMS efolder new API
```
aroltsch: I am hoping we will turn on the new API on around Dec 10.

app/services/external_api/vbms_service.rb
```diff
@@ -21,7 +21,7 @@ def self.fetch_documents_for(download)
     @vbms_client ||= init_client
 
     request = if FeatureToggle.enabled?(:vbms_efolder_service_v1)
-                VBMS::Requests::FindDocumentSeriesReference.new(download.file_number)
+                VBMS::Requests::FindDocumentVersionReference.new(download.file_number)
```
aroltsch: This service is much faster.

db/migrate/20171127190302_rename_columns_in_records.rb
```diff
@@ -0,0 +1,6 @@
+class RenameColumnsInRecords < ActiveRecord::Migration
+  def change
+    rename_column :records, :vva_source, :source
```
aroltsch: Renaming these two columns to be named the same as for the old version for simplicity sake.

app/jobs/v2/download_manifest_job.rb
```diff
@@ -0,0 +1,27 @@
+class V2::DownloadManifestJob < ActiveJob::Base
+  queue_as :default
+
+  around_perform :set_vbms_version
+
+  def perform(service, manifest_source)
+    documents = service.fetch_documents_for(manifest_source.manifest)
+    DocumentCreator.new(manifest_source: manifest_source).create(documents)
+  # TODO: make a parent class for these two errors
```
aroltsch: In my next PR, I will think of a better way to handle these errors.

app/models/manifest.rb
```diff
@@ -4,4 +4,49 @@ class Manifest < ActiveRecord::Base
...
+    sources.find_or_create_by(source: "VBMS")
+  end
+
+  def vva_source
+    sources.find_or_create_by(source: "VVA")
+  end
+
+  # If we do not yet have the veteran_first_name saved in Caseflow's DB, then
+  # we want to fetch it from BGS, save it to the DB, then return it
+  # TODO: these 3 methods are identical, let's simplify this
```
aroltsch: I will think of a better way to handle this. Currently we will call `update` 3 times - no good.

app/jobs/v2/download_manifest_job.rb
```diff
@@ -0,0 +1,27 @@
+class V2::DownloadManifestJob < ActiveJob::Base
+  queue_as :default
+
+  around_perform :set_vbms_version
+
+  def perform(service, manifest_source)
+    documents = service.fetch_documents_for(manifest_source.manifest)
+    DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
+  # TODO: make a parent class for these two errors
+  rescue VBMS::ClientError, VVA::ClientError => e
```
aroltsch: In my next PR, I will think of a better way to handle these errors.

app/models/manifest.rb
mdbenjam: I think we handle this problem decently in the current download model. Really this is doing a similar thing that associated_vacols_model is doing in Caseflow. Maybe we can take a page from that.

app/models/manifest_source.rb
```diff
@@ -9,4 +9,22 @@ class ManifestSource < ActiveRecord::Base
...
+    update(status: :pending)
+    V2::DownloadManifestJob.perform_now(service, self)
+    update(status: :success, fetched_at: Time.zone.now)
+
+    # TODO: handle error
+    # TODO: start downloading records from VBMS and save to S3
+  end
+
+  def service
+    source == "VBMS" ? VBMSService : VVAService
```
mdbenjam: Can we use a case statement here? Not clear we will ever expand to other services, but if we do it'd be a nice pattern to follow. Also if it's not either VVA or VBMS we probably want to throw an error.

app/models/manifest_source.rb
```diff
@@ -9,4 +9,22 @@ class ManifestSource < ActiveRecord::Base
   has_many :records
 
   validates :manifest, :source, presence: true
+  validates :manifest, uniqueness: { scope: :source }
+
+  def start!
+    return if success? && fetched_at && fetched_at > 3.hours.ago
+
+    # TODO: what do we do if it is pending, hmmmm
+    # If it is pending, don't make another request
```
mdbenjam: Agree with that, we should be careful with this. If we do it right we can get rid of stale object errors I like this approach:

https://bibwild.wordpress.com/2013/06/03/activerecord-atomic-check-and-update-through-optimistic-locking/

Though, this means we have to do this job async and then check for the right status in the end. Might add some complexity.

app/services/document_creator.rb
```diff
@@ -0,0 +1,44 @@
...
+    ManifestSource.transaction do
+      external_documents.each do |document|
+        Record.create_from_external_document(manifest_source, document)
+      end
+    end
+  end
+
+  # Override the getter to return only non-restricted documents
+  def external_documents
+    @external_documents.reject { |document| RESTRICTED_TYPES.include?(document.type_id) || document.restricted? }
```
mdbenjam: What does the `document.restricted?` call do? I don't see that method.

app/models/record.rb
```diff
@@ -3,4 +3,19 @@ class Record < ActiveRecord::Base
...
+  def self.create_from_external_document(manifest_source, document)
+    # TODO: add DB index for [manifest_source_id, external_document_id]
+    find_or_initialize_by(manifest_source: manifest_source, external_document_id: document.document_id).tap do |t|
+      t.assign_attributes(
+        type_id: document.type_id,
+        type_description: document.type_description,
+        mime_type: document.mime_type,
+        received_at: document.received_at,
+        jro: document.jro,
+        source: document.source
```
mdbenjam: Wait why do we need the source, we can get that from the manifest_source right?

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -0,0 +1,19 @@
+class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
+  type :manifest
+
+  attribute :veteran_first_name
+  attribute :veteran_last_name
+  attribute :created_at
+  attribute :updated_at
```
mdbenjam: Does updated_at become our new manifest_fetched_at?

app/jobs/v2/download_manifest_job.rb
```diff
@@ -0,0 +1,27 @@
...
+    Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+    Raven.capture_exception(e)
+  end
+
+  private
+
+  # Use VBMS efolder new API
+  def set_vbms_version
+    FeatureToggle.enable!(:vbms_efolder_service_v1)
```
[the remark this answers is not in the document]
amprokop: Hmm yeah, agree with Mark, perhaps decoupling the rollout of VBMS v1 from v2 of our API is a good idea. v2 of our API might not be running in production before December 10th anyway, right?

app/models/manifest.rb
aroltsch: Yeah, doing a similar thing sounds good. My approach here avoids setting `@veteran_info_fetched = true` and instead I set an empty string on the column if BGS info is empty for the field, this way we don't try to set the field again.

app/models/manifest_source.rb
aroltsch: Yes! I will create a ticket for this.

app/models/record.rb
aroltsch: The `source` in this case a completely different source from what you are thinking. This is a `source` returned from VBMS and VVA.

app/models/manifest_source.rb
aroltsch: Yes!

app/models/serializers/v2/manifest_serializer.rb
aroltsch: Why would we need `manifest_fetched_at` on the manifest? Or are you referring to `fetched_at` on the ManifestSource object?

app/services/document_creator.rb
aroltsch: It is an attribute on the external_document https://github.com/department-of-veterans-affairs/connect_vbms/blob/master/lib/vbms/requests/find_document_series_reference.rb#L62. Would it help to re-name `document` to ` external_document` to avoid confusion?

app/jobs/v2/download_manifest_job.rb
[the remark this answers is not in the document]
mdbenjam: Okay as long as we don't use this in prod before turning on the API, sounds good to me.

app/models/manifest.rb
mdbenjam: Yeah you can get rid of `@veteran_info_fetched` with saving an empty string, but we can still gain by moving the updates all into one method.

app/models/serializers/v2/manifest_serializer.rb
mdbenjam: As in, we currently return `manifest_fetched_at` to Reader which was the most recent time we fetched from the manifest. Presumably we can use `updated_at` for that task now.

app/services/document_creator.rb
mdbenjam: No I think it's fine the way it is. I was just asking for my own understanding.

app/models/serializers/v2/manifest_serializer.rb
aroltsch: `updated_at` won't be updated each time the manifest is fetched. Instead `fetched_at` on the ManifestSource will be updated. Should we return `fetched_at` for each source? Why do you need this information?

app/models/manifest.rb
aroltsch: Yes, will take care of it in the next PR.

app/models/serializers/v2/manifest_serializer.rb
mdbenjam: Ah gotcha. We display the most recent fetched time to Reader users so they know how recent their information is.

Issue #675, opened 2017-11-03
[FE] Error Messaging for Hanging UI


Chingujo:
##  Story
As a eFolder Express user, when a download of an eFolder is timed out, I want to see an error message with appropriate next steps instead of an endless spinning icon. 

Steps to re-produce (locally):
1. Start rails s for eX, don't start Sidekiq
2. Search for any DEMO number, watch the spinner going forever. 

## Acceptance Criteria
- Verify that the below "We couldn't find this folder." error message is displayed when the request times out after the back end 
- Verify that the error message "We couldn't find this efolder" is in accordance with the message styling in the style guide.
- Verify that the copy below the error message reads " You can go back to the home page and Search for another efolder. If you continue  to experience problems, please use the Send feedback link to submit a ticket."
- Verify 'Search for another efolder' is a link that takes you back to the home page
- Verify that 'Send feedback' is a link that opens feedback 

## Mocks 

![image](https://user-images.githubusercontent.com/32683958/33615641-a49b1670-d9a8-11e7-8198-85c96b940ab0.png)

Chingujo:
@lakohl 

Is this good to go? Took the stylings from the style guide and mock reflects the copy with the highlighted actions as links

Thank you =D!

aroltsch:
Add steps to reproduce it

lakohl:
@Chingujo Adjusting copy slightly.
<img width="719" alt="screen shot 2017-11-20 at 2 15 30 pm" src="https://user-images.githubusercontent.com/4975959/33036728-4e610eb0-cdfd-11e7-8593-4d395caace84.png">

Chingujo:
Thanks @lark!! Updated the ac and the mock :)

@lowellnava My understanding was that we would eventually have to determine a timeout but that there isn't a BE ticket yet. However, Anya mentioned there may be a front end approach to this and we could build out the UI end.

@aroltsch to confirm or deny :)

lowellrex:
This case will arise when a sidekiq job throws an exception and the job is terminated without changing the Download's status from fetching_manifest. The solution here will be to catch all StandardErrors on the backend and change the Download's status to some error status if the current status is fetching_manifest. The front-end will display the message if the download's status is set to that error status.

lowellrex:
Updating the mock to reflect the line breaks as rendered in chrome on my machine.

[...]

PR #744, opened 2017-11-29
Display error page when sidekiq dies and download status is 0

lowellrex:
Connects #675.

I would like some in-progress feedback on this approach to solve the problem, while I finish up writing tests.

### DIFF
--- a/app/jobs/download_all_manifest_job.rb
+++ b/app/jobs/download_all_manifest_job.rb
@@ -24,5 +24,11 @@ class DownloadAllManifestJob < ActiveJob::Base
       download.update_attributes!(status: :pending_confirmation)
     end
     external_documents
+
+    # efolder issue #675: Set the download's status to manifest_fetch_error
+    # if this job fails and the download's status == fetching_manifest.
+  rescue StandardError => e
+    download.update_attributes!(status: :manifest_fetch_error) if download.fetching_manifest?
+    raise e
   end
 end
--- a/app/jobs/download_all_manifest_job.rb
+++ b/app/jobs/download_all_manifest_job.rb
@@ -25,8 +25,8 @@ class DownloadAllManifestJob < ActiveJob::Base
     end
     external_documents
 
-    # efolder issue #675: Set the download's status to manifest_fetch_error
-    # if this job fails and the download's status == fetching_manifest.
+    # efolder #675: Catch all errors and change the status so that the database row reflects
+    # that an erorr occurred and the efolder UI does not display a spinner forever.
   rescue StandardError => e
     download.update_attributes!(status: :manifest_fetch_error) if download.fetching_manifest?
     raise e
--- a/app/views/downloads/_confirm.html.erb
+++ b/app/views/downloads/_confirm.html.erb
@@ -43,6 +43,11 @@
       </div>
     </div>
 
+  <% elsif @download.manifest_fetch_error? %>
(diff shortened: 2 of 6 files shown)


aroltsch:
I think this approach is good for now. Hopefully when we finish the refactoring, we can remove this.

spec/jobs/download_all_manifest_job_spec.rb
```diff
@@ -95,5 +95,14 @@
         expect(download.reload).to be_pending_confirmation
       end
     end
+
+    context "when we encounter an unexpected error fetching manifests" do
+      # Mock a failure to one of the service-specific manifest fetches to mimic an unexpected error in the job.
+      before { allow(DownloadManifestJob).to receive(:perform_now).and_raise(StandardError) }
```
aroltsch: So was it the issue with the feature tests?

app/views/downloads/_confirm.html.erb
```diff
@@ -43,6 +43,11 @@
       </div>
     </div>
 
+  <% elsif @download.manifest_fetch_error? %>
+      <h1 class="cf-msg-screen-heading cf-error">We couldn't find this efolder.</h1>
```
aroltsch: If you take a look at the mock, I think they want the message split into two lines.

spec/jobs/download_all_manifest_job_spec.rb
```diff
@@ -95,5 +95,14 @@
         expect(download.reload).to be_pending_confirmation
       end
     end
+
+    context "when we encounter an unexpected error fetching manifests" do
+      # Mock a failure to one of the service-specific manifest fetches to mimic an unexpected error in the job.
```
aroltsch: You can probably remove this comment as line 101 says it all :)

app/jobs/download_all_manifest_job.rb
```diff
@@ -24,5 +24,11 @@ def perform(download)
       download.update_attributes!(status: :pending_confirmation)
     end
     external_documents
+
+    # efolder issue #675: Set the download's status to manifest_fetch_error
+    # if this job fails and the download's status == fetching_manifest.
```
aroltsch: We usually don't add issue numbers in the comments. I would say it would be better to describe the problem in the comment. Also, I think this comment "Set the download's status to manifest_fetch_error if this job fails and the download's status == fetching_manifest" is not needed because the code itself acts as that comment.

app/views/downloads/_confirm.html.erb
lowellrex: Just chatted with Phillip and the single line seems fine (though the red color is a little off and may result in an update to caseflow-commons). I'll update the screenshot in the original ticket.

spec/jobs/download_all_manifest_job_spec.rb
lowellrex: I just avoided attempting to make a job run in the feature test and instead tested it in two places. This is not ideal because the feature test is no longer a true end-to-end test (and the tests potentially rely too much on implementation of the feature and not its behaviour), but I figure it is a good enough solution for now.

app/views/downloads/_confirm.html.erb
lowellrex: Created a PR in caseflow-commons to address the mismatched color issue: https://github.com/department-of-veterans-affairs/caseflow-commons/pull/120

PR #745, opened 2017-11-30
Anya/701 manifests controller v2 part2

aroltsch:
connects #701 

This is a second part of the PR:
1. Implemented some of the TODO comments from the PR #1
2. Wrote tests

Things to do (I will create tickets):
1. Make the job asynchronous and handle pending state
2. Implement start caching records from VBMS to S3 once manifest is fetched. 

Steps to Validate:
1. 
```Curl the API with curl "http://localhost:3000/api/v2/manifests" -H "Authorization: Token token=token" -H "css_id: CF_Q_283" -H "station_id: 283" -H "veteran_id: 1234"```
2. Ensure statuses are correct
3. Ensure the record list is accurate

### CODE REVIEW

File: app/jobs/v2/download_manifest_job.rb [hunk 1]
```diff
@@ -3,18 +3,15 @@ class V2::DownloadManifestJob < ActiveJob::Base
 
   around_perform :set_vbms_version
 
-  def perform(service, manifest_source)
-    documents = service.fetch_documents_for(manifest_source.manifest)
+  def perform(manifest_source)
+    documents = ManifestFetcher.new(manifest_source: manifest_source).process
+    return if documents.blank?
     DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
-  # TODO: make a parent class for these two errors
```
aroltsch: Error handling was moved to the ManifestFetcher so the job logic is as simple as possible.

app/models/veteran.rb
```diff
@@ -22,11 +22,6 @@ def found?
     bgs_record != :not_found
   end
 
-  # def accessible?
```
aroltsch: We don't need this.

app/models/manifest.rb
```diff
@@ -20,33 +19,33 @@ def vva_source
...
   end
 
-  def veteran_last_four_ssn
-    super || begin
-      update_attributes(veteran_last_four_ssn: veteran.last_four_ssn || "") if veteran
-      super
-    end
+  def self.find_or_create_by_user(user:, file_number:)
+    manifest = Manifest.find_or_create_by(file_number: file_number)
+    manifest.user_manifests.find_or_create_by(user: user).touch
```
aroltsch: We will create a UserManifest record once per manifest and then just update `updated_at` timestamp.

app/models/manifest.rb
```diff
@@ -20,33 +18,33 @@ def vva_source
...
 
-  # If we do not yet have the veteran_first_name saved in Caseflow's DB, then
+  # If we do not yet have the veteran info saved in Caseflow's DB, then
   # we want to fetch it from BGS, save it to the DB, then return it
-  # TODO: these 3 methods are identical, let's simplify this
-  def veteran_first_name
-    super || begin
-      update_attributes(veteran_first_name: veteran.first_name || "") if veteran
-      super
+  %w(veteran_first_name veteran_last_name veteran_last_four_ssn).each do |name|
```
mdbenjam: RAILS MAGIC AF

app/models/manifest.rb
mdbenjam: Hmm I'm wondering if we should create a new user manifest file for each request. Where did we land on Shane's point about being able to identify which documents were downloaded by which user at a given point? If we update records we won't have that ability. Are there pitfalls to creating new `user_manifest` records for each search?

app/services/manifest_fetcher.rb
```diff
@@ -0,0 +1,23 @@
...
+  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError]
+
+  attr_accessor :manifest_source
+
+  def process
+    manifest_source.update(status: :pending)
+    documents = manifest_source.service.fetch_documents_for(manifest_source.manifest)
+    manifest_source.update(status: :success, fetched_at: Time.zone.now)
+    documents
+  rescue *EXCEPTIONS => e
```
mdbenjam: Why log exceptions here since we already do that in Api::V1::ApplicationController? What about catching, setting the status as failed, then re-throwing the error so our standard error handling code picks it up.

spec/jobs/v2/download_manifest_job_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    context "when document list is not empty" do
+      before do
+        allow_any_instance_of(ManifestFetcher).to receive(:process).and_return(documents)
+      end
+
+      it "creates document records" do
+        subject
+        expect(manifest.records.size).to eq 2
```
mdbenjam: Should we check anything more? Like what the fields of the documents are? Not sure if it's necessary, just curious of your opinion.

spec/models/manifest_source_spec.rb
```diff
@@ -0,0 +1,52 @@
...
+        expect(V2::DownloadManifestJob).to have_received(:perform_now)
+      end
+    end
+
+    context "when fetched less than 3 hours ago" do
+      before do
+        source.update_attributes!(fetched_at: Time.zone.now - 2.hours, status: :success)
+      end
+
+      it "starts the manifest job" do
```
mdbenjam: It should say "does not start the manifest job"

app/models/manifest_source.rb
```diff
@@ -13,19 +13,17 @@ class ManifestSource < ActiveRecord::Base
...
-    update(status: :pending)
-    V2::DownloadManifestJob.perform_now(service, self)
-    update(status: :success, fetched_at: Time.zone.now)
-
-    # TODO: handle error
+    V2::DownloadManifestJob.perform_now(self)
     # TODO: start downloading records from VBMS and save to S3
   end
 
+  def current?
```
mdbenjam: Might want to either makes these two methods private or test them.

spec/models/manifest_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      end
+    end
+
+    context "when veteran first name is nil" do
+      let(:name) { nil }
+
+      it "should set the veteran_first_name" do
+        expect(manifest.reload.veteran_first_name).to eq "June"
+      end
+    end
```
mdbenjam: Maybe one more case where BGS returns nil, make sure we set empty string.

spec/services/document_creator_spec.rb
```diff
@@ -0,0 +1,48 @@
+describe DocumentCreator do
```
mdbenjam: Maybe make `external_documents` private since we're not directly testing it. Thought not sure how that works since it's overriding the getter.

spec/services/manifest_fetcher_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+
+        it "saves manifest status as success and updated fetched at" do
+          expect(subject.size).to eq 2
+          expect(source.reload.status).to eq "success"
+          expect(source.reload.fetched_at).to_not be_nil
+        end
+      end
+
+      context "when VBMS client returns error" do
```
mdbenjam: VBMS should be VVA in this description.

app/models/manifest.rb
aroltsch: Yeah, true! I think it is okay if we create user_manifests on every search.

app/models/manifest.rb
aroltsch: haha

app/models/manifest_source.rb
aroltsch: Yes! True

app/services/manifest_fetcher.rb
aroltsch: Great point

spec/jobs/v2/download_manifest_job_spec.rb
aroltsch: I think it is okay the way it is. The idea is to obfuscate the deeper logic. The job should not worry about these little details and can trust DocumentCreator to do the right thing. We can check the exact fields for this method which I think I forgot to add the tests for: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/record.rb#L7

spec/models/manifest_source_spec.rb
aroltsch: good catch

spec/models/manifest_spec.rb
aroltsch: Good addition to the test

spec/services/document_creator_spec.rb
aroltsch: Because we are overriding the getter, we cannot make it private.

app/services/manifest_fetcher.rb
aroltsch: Hmm, actually thinking about this more. We need to gracefully handle the error here. If we re-throw the error, would the Api::V1::ApplicationController return 500 then?

spec/services/manifest_fetcher_spec.rb
aroltsch: you probably meant line 58, I corrected it

app/services/manifest_fetcher.rb
mdbenjam: It will, why do we want to handle this gracefully though? If there's an error isn't that a 500?

aroltsch: We want to catch the error, set the status and continue to the next manifest_source. Then in the response we want to return status (failed/success/pending) for each manifest_source instead of 500.

PR #751, opened 2017-12-05
Anya/add validations manifests controller

aroltsch:
connects #701

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -32,6 +32,12 @@ class ExternalApi::BGSService
     end
   end
 
+  def valid_file_number?(file_number)
+    number = (file_number || "").strip
+    return true if /^\d+$/ =~ number && number.length >= 8
+    false
+  end
+
   private
 
   def init_client
--- a/app/controllers/api/v2/manifests_controller.rb
+++ b/app/controllers/api/v2/manifests_controller.rb
@@ -35,8 +35,6 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
     @manifest ||= Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
   end
 
-  private
-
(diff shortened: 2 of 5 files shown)


app/services/external_api/bgs_service.rb
```diff
@@ -32,6 +32,12 @@ def check_sensitivity(file_number)
     end
   end
 
+  def valid_file_number?(file_number)
```
mdbenjam: Hmm, this doesn't use any BGS stuff. Should this be a class method?

app/services/external_api/bgs_service.rb
```diff
@@ -32,6 +32,12 @@ def check_sensitivity(file_number)
...
+    number = (file_number || "").strip
+    return true if /^\d+$/ =~ number && number.length >= 8
```
mdbenjam: Is there any upper bound on length?

app/services/external_api/bgs_service.rb
aroltsch: a class method on the BGSService?

app/services/external_api/bgs_service.rb
aroltsch: Hmm, good question. I don't know if there is. I just followed the logic that we currently have: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/search.rb#L17

app/services/external_api/bgs_service.rb
mdbenjam: Yes

app/services/external_api/bgs_service.rb
kierachell: BGS tells you:
```
<message>findFlashes: FileNumber is invalid, must be 8 or 9 digits</message><
```
screen shot 2017-12-05 at 7 37 23 pm

Issue #740, opened 2017-11-27
Sanitize incoming user data from CSS in single location


lowellrex:
This ticket exists to address the work that remains after doing the migrations in #569.

After chatting with @shanear on Wednesday, we determined that the tidiest way of sanitizing data we receive from CSS for use by our applications would be to move all of that logic into a single module. This module would allow us to be more resilient to changes made within CSS (for example, if we uppercase all CSS IDs at their point of entry into the Caseflow ecosystem then we do not care if the casing of CSS IDs change upstream), as well as allowing for parity across all Caseflow applications (for example, if we strip leading and trailing spaces from email addresses in one application and not in another it would be more difficult to associate a user in efolder with that same user in caseflow).

Currently, we authenticate users in SessionsController.create() via the information returned from CSS and Api::V1::ApplicationController.authenticate_or_authorize() when the API request contains the correct API key. SessionController creates a hash that represents the authenticated user and stores it in the session's user element which we can later use to create a User object through User.from_session() in BaseController.current_user() (or which the caseflow applications can use for their own authentication purposes). Api::V1::ApplicationController.authenticate_or_authorize() uses User.find_or_create_by() to instantiate a new User object.

I propose we create a new module to consolidate the handling of authenticated users in efolder, which I am calling AuthenticatedUser. This module could consolidate a lot of that logic as follows:

Existing | Proposed
---------|---------
User.from_css_auth_hash() | AuthenticatedUser.from_css_auth_hash(), AuthenticatedUser.serialize()
User.from_session() | AuthenticatedUser.deserialize(), User.from_authenticated_user()
Api::V1::ApplicationController.authenticate_or_authorize() | AuthenticatedUser.from_request_headers(), User.from_authenticated_user()

cc: @aroltsch 

## AC:
* Create module to handle sanitizing of data from CSS, serializing/deserializing that object for use by caseflow applications
* Verify that we only ever instantiate Users from CSS data (from_session, from_css_auth_hash) and replace it with the new module's way of instantiating
* Remove existing code that sanitizes data from CSS
* Update existing records in the database if they do not agree with the sanitized data format in the new module
* Ensure that changes to this module do not alter behaviour of downstream Caseflow applications

## Test plan:
* Make sure users can still log in to efolder UI
* Make sure Reader still works (made slight--non-breaking--changes to the authentication/authorization process for the efolder API)

kierachell:
like Caseflow commons?

[...]

PR #753, opened 2017-12-05
Normalize CSS data and decouple CSS authentication from User creation

lowellrex:
Connects #740.

Created new AuthenticatedVisitor model to normalize CSS data (specifically trimming email addresses and uppercasing CSS IDs) the first time it hits our system.

One nice side effect of this change is that the new AuthenticatedVisitor model has no knowledge of the User object (or any of its consumers), and its consumers have no knowledge of its functioning either (hopefully approaching the mythical black box with a well-defined interface!) so it is ripe for plucking out to its own service should we ever decide to go that direction.

Still need to add tests for the AuthenticatedVisitor model, but wanted to get this first pass at the design out the door so I could get some feedback as early as possible.


app/models/authenticated_visitor.rb
```diff
@@ -0,0 +1,41 @@
...
+          email: raw_css_response["http://vba.va.gov/css/common/emailAddress"],
+          name: "#{first_name} #{last_name}",
+          roles: raw_css_response.attributes["http://vba.va.gov/css/caseflow/role"],
+          station_id: raw_css_response["http://vba.va.gov/css/common/stationId"])
+    end
+  end
+
+  protected
+
+  def css_auth_hash
```
shanear: Since this sort of relies on the controller infrastructure, I think it should probably go in the controller and be passed into `AuthenticatedVisitor` as an argument.

app/models/authenticated_visitor.rb
```diff
@@ -0,0 +1,41 @@
...
+  include ActiveModel::Serializers::JSON
+
+  attr_accessor :id, :css_id, :email, :name, :roles, :station_id
+
+  def email=(value)
+    @email = value.try(:strip)
+  end
+
+  def css_id=(value)
+    @css_id = value.try(:upcase)
```
shanear: not sure what the standard syntax is, but I typically do this like this:

`@css_id = value && value.upcase`

app/models/authenticated_visitor.rb
```diff
@@ -0,0 +1,41 @@
+class AuthenticatedVisitor
```
shanear: for some reason I feel like `AuthenticatedVisitor` isn't really conveying what this class is doing. We never really use the word `Visitor`.

I feel like `SamlAuthentication` or `SamlAuthenticatedUser` or `CssAuthentication` or `CssAuthenticatedUser` all work better

app/models/user.rb
```diff
@@ -42,31 +40,21 @@ def denied?(thing)
...
-
-      {
-        id: auth_hash.uid,
-        css_id: auth_hash.uid,
-        email: raw_css_response["http://vba.va.gov/css/common/emailAddress"],
-        name: "#{first_name} #{last_name}",
-        roles: raw_css_response.attributes["http://vba.va.gov/css/caseflow/role"],
-        station_id: raw_css_response["http://vba.va.gov/css/common/stationId"]
-      }
+    def from_css_id_and_station(params)
```
shanear: you can rewrite this as `def from_css_id_and_station(css_id:, station_id:)`

Additionally i feel like this method would be more informative if it said `from_api_authentication_values` or something like that. It took me a bit of digging to even figure out why we have this.

shanear:
This is looking verrry nice 😎 

I left some comments and it still needs some more tests, but looking quite good.

app/models/authenticated_visitor.rb
lowellrex: I wanted to steer clear of the word "User" for this class because it would imply a strong relationship (possibly as a sub-class) between this class the User class, and one of the catalysts for this PR is to more cleanly separate the authentication functionality from the User model.

Since this class represents an entity I figured it should be a noun, and since this entity is only a transient presence in efolder (they will either become a User or be sent back home to their other Caseflow home), I settled on "Visitor".

I'm definitely up for changing the name (prefixing with "Css" seems like a good idea), but I think making the name a noun and making the distinction between this entity and the User model clear should be key considerations when naming this entity.

shanear: How about `CssAuthentication` then? That is way more clear to me.

lowellrex: For sure, I definitely think adding the "Css" prefix makes a lot of sense.

However, I still think it is important for the name of the entity to convey that objects of this class represent a single person who is somewhere in the process of getting authenticated through CSS. So I think it makes sense for the name to reflect that this object is an individual person. Maybe "CssAuthenticationVisitor"? (Once again trying to steer clear of the word "User" for the reasons listed in my above comment).

This may seem like a trivial thing to inspire so much back-and-forth, but names have a way of persisting for years and I want to make sure that this class is named properly.

spec/models/css_authentication_session_spec.rb
```diff
@@ -0,0 +1,75 @@
+describe CssAuthenticationSession do
+  let(:id) { "VACOSOLOH" }
+  let(:css_id) { "VACOSOLOH" }
+  let(:email) { "Han.Solo@va.gov" }
```
cyber-gov-shane: hahah yesssss

spec/models/user_spec.rb
```diff
@@ -76,37 +66,17 @@
...
-    let(:auth_hash_data) do
-      data = {
-        "http://vba.va.gov/css/common/fName" => "Kanye",
-        "http://vba.va.gov/css/common/lName" => "West",
-        "http://vba.va.gov/css/common/emailAddress" => "kanye@va.gov",
-        "http://vba.va.gov/css/common/stationId" => "123"
-      }
+  context ".from_api_authenticated_values" do
+    let(:css_id) { "lowercase_id" }
+    let(:uppercased_css_id) { css_id.upcase }
```
shanear: for tests, its usually best to hard code things when you can, id put `"LOWERCASE_ID"` here

shanear: you can write it directly into the `it` block if you want too.

shanear: Also, this applies to a bunch of your other tests.

app/models/user.rb
```diff
@@ -42,31 +40,21 @@ def denied?(thing)
...
   class << self
-    def from_session(session, request)
+    def from_session_and_request(session, request)
       return nil unless session["user"]
-      user = session["user"]
-      find_or_create_by(css_id: user["css_id"], station_id: user["station_id"]).tap do |u|
-        u.name = user["name"]
-        u.email = user["email"]
-        u.roles = user["roles"]
+      visitor = CssAuthenticationSession.new(session["user"])
```
shanear: `visitor` as the variable name doesn't make sense any more.

spec/models/css_authentication_session_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  let(:name) { "Han Solo" }
+  let(:roles) { "Reader" }
+  let(:station_id) { Random.rand(499) + 1 }
```
shanear: I typically avoid rand's in tests as they can produce non deterministic tests. This probably wont cause any issues, just a note.

app/controllers/sessions_controller.rb
```diff
@@ -3,7 +3,7 @@ class SessionsController < ApplicationController
   skip_before_action :authenticate, only: [:create]
 
   def create
-    session["user"] = User.from_css_auth_hash auth_hash
+    session["user"] = CssAuthenticationSession.from_css_auth_hash(css_auth_hash).as_json
```
shanear: ✨ beautiful ✨

app/models/user.rb
```diff
@@ -42,31 +40,21 @@ def denied?(thing)
...
-      {
-        id: auth_hash.uid,
-        css_id: auth_hash.uid,
-        email: raw_css_response["http://vba.va.gov/css/common/emailAddress"],
-        name: "#{first_name} #{last_name}",
-        roles: raw_css_response.attributes["http://vba.va.gov/css/caseflow/role"],
-        station_id: raw_css_response["http://vba.va.gov/css/common/stationId"]
-      }
+    def from_api_authenticated_values(css_id:, station_id:)
+      visitor = CssAuthenticationSession.new(css_id: css_id, station_id: station_id)
```
shanear: same as above `visitor` doesn't make a lot of sense as the variable name here.

spec/features/help_spec.rb
```diff
@@ -1,5 +1,7 @@
 RSpec.feature "Help" do
   scenario "Visiting the Help/FAQ page" do
+    User.authenticate!
```
shanear: should only authenticated users be able to visit the help page? This might not be necesary

lowellrex: I'm honestly kind of confused about how this test passed without the `User.authenticate!` line.

`visit("/")` [routes](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/config/routes.rb) to [DownloadsController.new()](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/downloads_controller.rb). [ApplicationController](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/application_controller.rb) runs a few filters before we enter DownloadsController.new() ([BaseController](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/base_controller.rb) also runs a few filters before we enter the target method, but--other than current_user() which turns out to be a no-op in this case--those do not have any impact on this code path).

The filter that causes this test to fail is [ApplicationController.authenticate()](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/application_controller.rb#L9). Since `current_user` will be nil when we reach this point in the code (since we haven't yet set it or faked the authentication), we will redirect to */auth/samlva* and exit the program.

The code changes in this PR did not change any of that behaviour, so I don't know why this only showed up as a failing test in this PR. However, I went ahead and ran this test locally on the master branch and this test fails in the same way without the `User.authenticate!` line.

Another alternative to this change would be to visit "/help" directly instead of visiting the efolder UI landing page. However, this would require changing this test file somewhat significantly and I also expect it will fail in the same way (since HelpController inherits from ApplicationController and does not explicitly skip the [ApplicationController.authenticate()](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/application_controller.rb#L9) filter).

shanear:
We will def want to thoroughly integration test these changes in UAT before deployment.

PR #755, opened 2017-12-06
[WIP] Add manual 304

mdbenjam:
Currently if a user requests a document for which they supply either an eTag or a last modified date, then we want to tell them to just use their cache. Currently this happens by us serving the request, rails hashing the response body, determining that it's the same, and sending a 304. Instead we'd like to short circuit and return the 304 immediately.

**Test Plan**
- [ ] Test with Reader hooked up to a local eFolder Express and ensure documents still download properly.


aroltsch:
@mdbenjam lemme know when you merge this, I'll update v2 API.

mdbenjam:
I don't think we should actually do this anymore.

Issue #700, opened 2017-11-14
[v2] Connect API RecordsController to the v2 data model


aroltsch:
# AC:
1. Create a new file called `controllers/api/v2/records_controller.rb`
2. Connect API::V2::DocumentsController to the new data model: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/database_structure.md
3. Ensure the new API endpoint works the same as `/api/v1/documents`

# Steps
Curl the API with curl "http://localhost:3000/api/v2/records/:id" -H "Authorization: Token token=token" -H "css_id: CF_Q_283" -H "station_id: 283"

kierachell:
**PASSED**

<img width="591" alt="screen shot 2017-12-13 at 8 22 41 am" src="https://user-images.githubusercontent.com/18075411/33941076-5226aca0-dfdf-11e7-880f-9a6c35e0459a.png">
<img width="605" alt="screen shot 2017-12-13 at 8 24 42 am" src="https://user-images.githubusercontent.com/18075411/33941077-5255ef24-dfdf-11e7-8e65-09c111324bc8.png">
<img width="678" alt="screen shot 2017-12-13 at 8 24 47 am" src="https://user-images.githubusercontent.com/18075411/33941078-528f5656-dfdf-11e7-839c-7a89e0073094.png">
<img width="634" alt="screen shot 2017-12-13 at 8 24 54 am" src="https://user-images.githubusercontent.com/18075411/33941079-530a0374-dfdf-11e7-8f05-c36b273afa8f.png">
<img width="930" alt="screen shot 2017-12-13 at 8 26 05 am" src="https://user-images.githubusercontent.com/18075411/33941080-53549a06-dfdf-11e7-9750-03fbc81d00b4.png">

PR #756, opened 2017-12-06
Connect RecordsController to the v2 data model

aroltsch:
connects #700 

TBD in the next PR:
* Automated tests

### DIFF
--- a/app/controllers/api/v2/records_controller.rb
+++ b/app/controllers/api/v2/records_controller.rb
@@ -34,7 +34,7 @@ class Api::V2::RecordsController < Api::V1::ApplicationController
   end
 
   def validate_access
-    forbidden("sensitive record") unless BGSService.new.check_sensitivity(record.file_number)
+    forbidden("sensitive record") unless record.accessible_by?(current_user)
   rescue ActiveRecord::RecordNotFound
     record_not_found
   end
--- a/app/controllers/api/v2/records_controller.rb
+++ b/app/controllers/api/v2/records_controller.rb
@@ -1,0 +1,50 @@
+class Api::V2::RecordsController < Api::V1::ApplicationController
+  before_action :validate_access
+  before_action :enable_caching
+
+  def show
+    result = record.fetch!
+    return document_failed if record.failed?
+
+    send_data(
+      result,
+      type: record.mime_type,
+      disposition: "attachment",
+      filename: record.s3_filename
+    )
+  end
+
+  private
+
+  def document_failed
+    render json: {
+      "errors": [
+        "title": "Document download failed",
+        "detail": "An upstream dependency failed to fetch document contents."
+      ]
+    }, status: 502
+  end
+
+  def enable_caching
+    expires_in 30.days, public: true
+  end
+
+  def record
+    @record ||= Record.includes(:manifest_source).find(params[:id])
+  end
+
+  def validate_access
+    forbidden("sensitive record") unless BGSService.new.check_sensitivity(record.file_number)
+  rescue ActiveRecord::RecordNotFound
+    record_not_found
+  end
+
+  def record_not_found
+    render json: {
+      "errors": [
+        "title": "Record not found",
+        "detail": "A record with that ID was not found in our systems."
+      ]
+    }, status: 404
+  end
+end
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -4,6 +4,44 @@ class Record < ActiveRecord::Base
   validates :manifest_source, :external_document_id, presence: true
   validates :external_document_id, uniqueness: true
 
+  enum status: {
+    pending: 0,
+    success: 1,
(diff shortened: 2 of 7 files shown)


app/controllers/api/v2/records_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def show
+    result = record.fetch!
+    return document_failed if record.failed?
+
+    send_data(
+      result,
+      type: record.mime_type,
+      disposition: "attachment",
+      filename: record.s3_filename
```
aroltsch: I changed the filename to `record.s3_filename` instead of `document.filename`. Let me know if you are using the very ling filename for anything

app/controllers/api/v2/records_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+  def enable_caching
+    expires_in 30.days, public: true
+  end
+
+  def record
+    @record ||= Record.includes(:manifest_source).find(params[:id])
+  end
+
+  def validate_access
+    forbidden("sensitive record") unless BGSService.new.check_sensitivity(record.file_number)
```
aroltsch: Manifest might have been download by some other user so we might not have ManifestUser record for the current_user so I thought it makes sense to check the access with BGS. Thoughts?

app/services/record_fetcher.rb
```diff
@@ -0,0 +1,29 @@
+class RecordFetcher
+  include ActiveModel::Model
+
+  attr_accessor :record
+
+  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError].freeze
+
+  def process
```
aroltsch: I am not a big fan of this method. It does 3 things: fetch, convert and store - nu bueno. However, I can't think of a better way to handle this :(.

app/controllers/api/v2/records_controller.rb
mdbenjam: Reader does not use that.

app/controllers/api/v2/records_controller.rb
mdbenjam: I think this is a bad idea. Instead of this, we can just disallow users from downloading a document if they don't have a corresponding user_manifest.

app/controllers/api/v2/records_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+    render json: {
+      "errors": [
+        "title": "Document download failed",
+        "detail": "An upstream dependency failed to fetch document contents."
+      ]
+    }, status: 502
+  end
+
+  def enable_caching
+    expires_in 30.days, public: true
```
mdbenjam: I'm actually thinking we should up this time to a year.

app/models/record.rb
```diff
@@ -4,6 +4,44 @@ class Record < ActiveRecord::Base
...
+  before_create :adjust_mime_type
+
+  delegate :manifest, :service, to: :manifest_source
+  delegate :file_number, to: :manifest
+
+  def fetch!
+    fetcher.process
+  end
+
+  def ssn
```
mdbenjam: Where is ssn used?

app/services/record_fetcher.rb
mdbenjam: I think it's okay. Though I have a higher tolerance than others about long methods.

app/controllers/api/v2/records_controller.rb
aroltsch: cool

app/controllers/api/v2/records_controller.rb
aroltsch: why do you think it is a bad idea, just curious?

app/services/record_fetcher.rb
aroltsch: It is okay but not good :)

app/models/record.rb
aroltsch: It is used in the request to the VVA service to get document_content: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/services/external_api/vva_service.rb#L26

app/controllers/api/v2/records_controller.rb
mdbenjam: It's just we'll hit this a ton. I think this API causes a good bit of our problems so keeping it as simple as possible is ideal.

app/controllers/api/v2/records_controller.rb
mdbenjam: Also, I'm adding a change to the v1 API that relates to caching. We should probably add it here too: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/755/files. It prevents us from doing the entire request if the client has cached it.

aroltsch: Ok, I will add that once you merge this PR.

PR #759, opened 2017-12-07
More tests for the records controller

aroltsch:
connects #700

### DIFF
--- a/spec/requests/api/v2/records_spec.rb
+++ b/spec/requests/api/v2/records_spec.rb
@@ -33,7 +33,7 @@ describe "Records API v2", type: :request do
       expect(response.code).to eq("404")
     end
 
-    context "when user owns corresponding download record" do
+    context "when user owns corresponding manifest record" do
       let!(:user_manifest) { UserManifest.create(user: current_user, manifest: manifest) }
 
       it "returns a document" do
@@ -45,8 +45,8 @@ describe "Records API v2", type: :request do
       end
 
       it "returns 502 if there is a VBMS/VVA error" do
-        allow(Fakes::DocumentService).to receive(:fetch_document_file)
-          .and_raise([VBMS::ClientError, VVA::ClientError].sample)
+        allow_any_instance_of(RecordFetcher).to receive(:process).and_return(nil)
+        record.update(status: :failed)
 
         get "/api/v2/records/#{record.id}"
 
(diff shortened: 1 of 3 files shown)


spec/requests/api/v2/records_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+        expect(response.code).to eq("401")
+      end
+    end
+
+    it "returns 404 if document ID is not found" do
+      get "/api/v2/records/8888"
+      expect(response.code).to eq("404")
+    end
+
+    context "when user owns corresponding download record" do
```
mdbenjam: You should change download here and below to `manifest record`

spec/requests/api/v2/records_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        expect(response.code).to eq("502")
+
+        json = JSON.parse(response.body)
+        expect(json["errors"].length).to eq(1)
+        expect(json["errors"].first["title"]).to eq("Document download failed")
+        expect(json["errors"].first["detail"]).to eq("An upstream dependency failed to fetch document contents.")
+      end
+
+      it "returns 500 if there is a StaleObjectError error" do
```
mdbenjam: Let's hope we never actually see this one :D 🤞

spec/services/record_fetcher_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    context "when the file is in s3 after it has been cached" do
+      before do
+        allow(S3Service).to receive(:fetch_content).and_return(nil)
+        allow(Fakes::DocumentService).to receive(:fetch_document_file).and_return(fake_pdf_content)
+      end
+
+      it "should cache in s3 from VBMS and then serve from s3" do
+        expect(subject).to eq fake_pdf_content
+        allow(S3Service).to receive(:fetch_content).and_return("from s3")
```
mdbenjam: Don't we want to see that S3 returns the right thing rather than mocking it out?

aroltsch: Hmm, how would you test that? When you call `RecordFetcher.new(record: record).process` second time, how would you test that it actually got the content from s3 instead of VBMS? Because both would return the same string if you don't mock the fetch_content for s3.

mdbenjam: Can you do this?
`expect(S3Service).to receive(:fetch_content)`

Set an expectation rather than an allow?

aroltsch: I could but it would not give accurate test results. You would test that fetch_content was called (and yes it would be called every time before fetching from VBMS). But what results it would return?

PR #761, opened 2017-12-08
Run test files in random order to surface leaking state

lowellrex:
[Prior to the addition](https://github.com/department-of-veterans-affairs/caseflow-efolder/commit/a6fc320576376bb0fbc7ea312bc09af26df0a709#diff-cc581a8e6ec5512b31550fa03b0ba9f7) of the `User.authenticate!` call in help_spec.rb, the test would fail if run by itself because the user was not allowed to visit the particular page the test directs them to. However, this test failure was masked by the fact that the state of previous tests leaked into the help_spec.rb test allowing this test to pass when run after a test where the user had already authenticated.

Randomizing the test order does not solve the issue of leaking state, however, it will likely reveal tests that pass as a result of their position in the test plan.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,8 @@ gem "sentry-raven"
 
 gem 'newrelic_rpm'
 
+gem "dogstatsd-ruby"
+
 # SSOI
 gem 'omniauth-saml-va', git: 'https://github.com/department-of-veterans-affairs/omniauth-saml-va', branch: 'paultag/css'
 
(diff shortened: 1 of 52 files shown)


shanear:
and... it turns out that test wasn't the only one that was order dependent 😬

Awesome that we are finally airing all of these out.

kierachell:
💪 RANDOMIZE ALL THE TESTS YASSSS

nickheiner-usds:
Lololol yes, this was very easy. Thanks for doing this, @lowellrex!

lowellrex:
Closing this out since it is not a priority at the moment.

PR #772, opened 2017-12-13
File number must be 8 or 9 char in length

aroltsch:
connects #701


aroltsch:
@lowellrex Actually you bring up a good point, I can change the error message to give more context.

Issue #736, opened 2017-11-27
No implicit conversion of nil into String


aroltsch:
Sentry link: https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/650/. It appears that `document.filepath` is nil.

mdbenjam:
Can't wait for when we calculate this rather than save it in the DB!

Chingujo:
Possibly fixed by refactoring making a comment for @aroltsch due to GH issues

astewarttistatech:
@lowellrex @aroltsch how is this to be validated if necessary?

lowellrex:
If a user can download an efolder from the UI, then this works.

astewarttistatech:
**PASSED**

Validated in: Prod/Preprod

Smoke test

PR #774, opened 2017-12-13
Set document filepath before downloading document from S3

lowellrex:
Connects #736.

I think this happened because somebody clicked "Try retrieving efolder again" while another job was waiting to be completed. This called `DownloadsController.retry()` which called `Download.reset!()` which sets `Document.filepath` to nil. I can confirm this with more investigation into the logs if we need.

This fix simply sets the document's filepath if it is nil so we have somewhere to download the document to.

### DIFF
--- a/app/models/document.rb
+++ b/app/models/document.rb
@@ -29,6 +29,10 @@ class Document < ActiveRecord::Base
     Zaru.sanitize! "#{cropped_type_name}-#{filename_date}-#{filename_doc_id}.#{preferred_extension}"
   end
 
+  def path
+    @path ||= File.join(download.download_dir, id.to_s)
+  end
+
   def fetch_content!(save_document_metadata:)
     return {
       content: fetcher.content(save_document_metadata: save_document_metadata),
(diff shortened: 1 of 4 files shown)


app/models/document.rb
```diff
@@ -101,7 +101,10 @@ def save_locally(content, index)
         f.write(content)
       end
     end
-    update_attributes!(filepath: filepath)
+  end
+
+  def build_filepath_with(index)
```
aroltsch: @lowellrex do you think we can spend some more time figuring why the filepath is nil in some documents.

app/services/download_documents.rb
```diff
@@ -109,6 +109,7 @@ def package_contents
 
     Zip::File.open(zip_path, Zip::File::CREATE) do |zipfile|
       @download.documents.success.each_with_index do |document, index|
+        document.build_filepath_with(index) if document.filepath.nil?
```
aroltsch: How can you be sure the document will get exactly the same index as here: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/services/download_documents.rb#L71

aroltsch: Maybe we can store `index` in the document column instead of a filepath and then calculate filepath on the fly. What is the purpose of attaching index anyways? Can we get rid of it?

Issue #562, opened 2017-08-01
[v2] Manifests Endpoint | Move calls to be async 


mdbenjam:
# AC
1. Verify when api/v2/manifests is called it should return immediately with whatever files exist in the DB
1. Verify when api/v2/manifests is called it starts a background job that downloads the document list and returns status as `pending`
1. Verify api/v2/manifests returns a fetched_at time for each source (VBMS and VVA) of when the background job last finished.

aroltsch:
Tech details: 
The files endpoint is using `perform_now` which is synchronous. 

Method peform_later is asynchronous. When you use this method, the job is not run at the moment, but rather is pushed in a job's queue. Method perform_now is synchronous and will run the job at the moment, no matter what is the job's state.

astewarttistatech:
**PASSED**

Validated: UAT | Postman

Screenshots:

<img width="860" alt="screen shot 2017-12-21 at 10 21 05 am" src="https://user-images.githubusercontent.com/23038505/34263863-e41e6c64-e63e-11e7-9180-178e9f8fd315.png">
<img width="861" alt="screen shot 2017-12-21 at 11 04 31 am" src="https://user-images.githubusercontent.com/23038505/34263864-e425ec5a-e63e-11e7-81dd-feb4bde3ec56.png">

PR #775, opened 2017-12-14
Fetch manifest async

aroltsch:
connects #562

### DIFF
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -1,6 +1,6 @@
 class Manifest < ActiveRecord::Base
-  has_many :sources, class_name: "ManifestSource"
-  has_many :user_manifests
+  has_many :sources, class_name: "ManifestSource", dependent: :destroy
+  has_many :user_manifests, dependent: :destroy
   has_many :records, through: :sources
   has_many :users, through: :user_manifests
 
--- a/app/models/manifest_source.rb
+++ b/app/models/manifest_source.rb
@@ -1,23 +1,21 @@
 class ManifestSource < ActiveRecord::Base
   enum status: {
-    pending: 0,
-    success: 1,
-    failed: 2
+    initialized: 0,
+    pending: 1,
+    success: 2,
+    failed: 3
   }
 
   belongs_to :manifest
-  has_many :records
+  has_many :records, dependent: :destroy
 
   validates :manifest, :source, presence: true
   validates :manifest, uniqueness: { scope: :source }
   validates :source, inclusion: { in: %w(VBMS VVA) }
 
   def start!
-    return if current?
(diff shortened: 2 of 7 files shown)


app/models/manifest_source.rb
```diff
@@ -1,23 +1,21 @@
 class ManifestSource < ActiveRecord::Base
   enum status: {
-    pending: 0,
-    success: 1,
-    failed: 2
+    initialized: 0,
```
aroltsch: We need this additional state because we only want to set `pending` when the manifest is being fetched.

app/models/manifest_source.rb
```diff
@@ -1,23 +1,21 @@
...
   validates :manifest, :source, presence: true
   validates :manifest, uniqueness: { scope: :source }
   validates :source, inclusion: { in: %w(VBMS VVA) }
 
   def start!
-    return if current?
-    # TODO: If it is pending, don't make another request
-    # https://bibwild.wordpress.com/2013/06/03/activerecord-atomic-check-and-update-through-optimistic-locking/
-    V2::DownloadManifestJob.perform_now(self)
-    V2::SaveFilesInS3Job.perform_later(self)
```
aroltsch: Removing it from here because we only need to start this job when Manifest has been fetched.

app/models/manifest.rb
```diff
@@ -1,6 +1,6 @@
 class Manifest < ActiveRecord::Base
-  has_many :sources, class_name: "ManifestSource"
-  has_many :user_manifests
+  has_many :sources, class_name: "ManifestSource", dependent: :destroy
```
mdbenjam: Oh I didn't know about this. So this means if we delete a Manifest we delete it's sources as well?

spec/requests/api/v2/manifests_spec.rb
```diff
@@ -68,9 +70,11 @@
     end
 
     it "returns empty array" do
-      get "/api/v2/manifests", nil, headers
-      expect(response.code).to eq("200")
-      expect(response.body).to eq(response_body)
+      perform_enqueued_jobs do
```
mdbenjam: Do we actually need to check anything about the job completing?

app/models/manifest.rb
aroltsch: Yes, it is called `cascade delete` in SQL. Not like we will be deleting anything anyways.

spec/requests/api/v2/manifests_spec.rb
aroltsch: We know it completed because the status is `success` for both manifests.

mdbenjam: 👍

Issue #763, opened 2017-12-11
Clean up Sentry Errors when receiving error responses from VBMS/VVA


mdbenjam:
Currently when VBMS or VVA have a problem we trigger a bunch of Sentry errors. Some examples are:
https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/685/
https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/612/ ([when VBMS is in maintenance mode](https://github.com/department-of-veterans-affairs/caseflow/issues/3643))
https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/671/

We should catch these errors in connect_vbms and connect_vva and return a reasonable error to eFolder which can then pass that error on to users.

PR #777, opened 2017-12-14
Sentry - Ignore exceptions when VBMS/VVA are in maintenance

aroltsch:
connects #763 

When VVA/VBMS is down, connect_vva and connect_vbms correctly catch the exception that is then passed to the eX and download gets a status of vva_connection_error or vbms_connection_error. 

When VBMS or VVA is in maintenance, do not capture these errors in Sentry.

### Note
Unfortunately, Sentry `excluded_exceptions` doesn't provide support for regular expressions so I created our own  class to exclude certain exceptions. 
See here what Raven does: https://github.com/getsentry/raven-ruby/blob/00c572b281b0862e2fbb7872b28c95e7c2578949/lib/raven/configuration.rb#L331

### DIFF
--- a/app/services/exception_logger.rb
+++ b/app/services/exception_logger.rb
@@ -1,0 +1,13 @@
+class ExceptionLogger
+  DEPENDENCY_MAINTENANCE = [/Could not get JDBC Connection/, /Maintenance.+VBMS/, /upstream connect error or disconnect/].freeze
+
+  def self.capture(e)
+    Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+    Raven.capture_exception(e) unless ignore_exception?(e)
+  end
+
+  # Do not send Sentry alerts when external systems are in maintenance mode
+  def self.ignore_exception?(e)
+    DEPENDENCY_MAINTENANCE.select { |m| m =~ e.to_s }.present?
+  end
+end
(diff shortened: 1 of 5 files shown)


app/services/exception_logger.rb
```diff
@@ -0,0 +1,13 @@
+class ExceptionLogger
+  DEPENDENCY_MAINTENANCE = [/Could not get JDBC Connection/, /Maintenance.+VBMS/, /upstream connect error or disconnect/].freeze
```
mdbenjam: I'm wondering if it's easier to check the status code: 418? What do we think is more likely to change. Just a thought, I'll leave it up to you.

mdbenjam: Does this check for just different instances of maintenance? Or also when they're down.

aroltsch: So far we know of these 3 exceptions when VVA/VBMS are in maintenance. We can add more if we see more. I think VVA returns the same error when it is down or in maintenance.  I am not sure about VBMS.

app/services/manifest_fetcher.rb
```diff
@@ -12,12 +12,7 @@ def process
     documents
   rescue *EXCEPTIONS => e
     manifest_source.update(status: :failed)
-    log_error(e)
+    ExceptionLogger.capture(e)
```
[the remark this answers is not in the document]
aroltsch: When a document download fails, we don't send any Sentry exceptions, we just set the status as failed.

PR #789, opened 2017-12-18
Switch DataDog to use dogstatsd

nickheiner-usds:
I used the same testing process as https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/776. I installed the macOS DataDog agent so this new approach could work.

In the screenshot below, you can see the data showing up ([sample notebook](https://app.datadoghq.com/notebook/23056/Delete%20this%20after%20caseflow-efolder%23789%20is%20merged)):

image

Here is the data from UAT:

image

### DIFF
--- a/app/services/data_dog_service.rb
+++ b/app/services/data_dog_service.rb
@@ -22,7 +22,7 @@ class DataDogService
 
   private_class_method def self.get_tags(app_name, attrs)
     extra_tags = attrs.reduce([]) do |tags, (key, val)|
-      tags + "#{key}:#{val}"
+      tags + ["#{key}:#{val}"]
     end
     [
       "app:#{app_name}",
--- a/app/services/data_dog_service.rb
+++ b/app/services/data_dog_service.rb
@@ -1,40 +1,35 @@
-require "dogapi"
+require "datadog/statsd"
 
 class DataDogService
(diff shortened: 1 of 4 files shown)


app/services/data_dog_service.rb
```diff
@@ -0,0 +1,35 @@
+require "datadog/statsd"
+
+class DataDogService
+  @statsd = Datadog::Statsd.new
+  @host = `curl http://instance-data/latest/meta-data/instance-id --silent || echo "not-ec2"`.strip
```
sunil-sadasivan: Didn't realize you can do this.  Is it better to use `ENV['HOSTNAME']`.  EC2 host names are in the form of their given ip address.

nickheiner-usds: That gives us something like:

```
$ echo $HOSTNAME
ip-172-30-70-179
```

whereas `meta-data/instance-id` gives:

```
$ curl http://instance-data/latest/meta-data/instance-id
i-0084c393557b25a64
```

Alan expressed a preference for the latter. If you feel strongly about this, we can change it.

sunil-sadasivan: No strong preference @nickheiner-usds.  Internal IPs can be re-used so I understand the preference for using instance-ids.

PR #794, opened 2017-12-20
Lowell test fixes 1

lowellrex:
The [pull request to run efolder tests in a random order](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/761) has been revealing a lot of tests that should fail but pass due to the order in which they are run. I had initially planned to fix all of those cases (thinking there would be few) in the same PR that randomizes the test order in Travis, however due to the large number of falsely passing tests, I think it best to incorporate in distinct PRs so they can be merged to master as soon as possible. This is the first batch of those test fixes.

### DIFF
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -56,13 +56,14 @@ RSpec.feature "Downloads" do
   end
 
   scenario "Creating a download" do
-    Fakes::BGSService.veteran_info = {
+    veteran_info = {
       "12341234" => {
         "veteran_first_name" => "Stan",
         "veteran_last_name" => "Lee",
         "veteran_last_four_ssn" => "2222"
       }
     }
+    allow_any_instance_of(Fakes::BGSService).to receive(:veteran_info).and_return(veteran_info)
 
     visit "/"
     expect(page).to_not have_content "Recent Searches"
@@ -71,7 +72,7 @@ RSpec.feature "Downloads" do
     click_button "Search"
 
     # Test that Caseflow caches veteran name for a download
-    Fakes::BGSService.veteran_info = {}
+    allow_any_instance_of(Fakes::BGSService).to receive(:veteran_info).and_return(nil)
 
     @download = @user_download.last
     expect(@download).to_not be_nil
@@ -91,13 +92,14 @@ RSpec.feature "Downloads" do
   end
 
   scenario ": initial wait" do
-    Fakes::BGSService.veteran_info = {
+    veteran_info = {
       "1234" => {
         "veteran_first_name" => "Stan",
         "veteran_last_name" => "Lee",
         "veteran_last_four_ssn" => "2222"
       }
     }
+    allow_any_instance_of(Fakes::BGSService).to receive(:veteran_info).and_return(veteran_info)
 
     visit "/"
 
@@ -110,13 +112,14 @@ RSpec.feature "Downloads" do
   end
 
   scenario "Searching for an errored download tries again" do
-    Fakes::BGSService.veteran_info = {
+    veteran_info = {
       "55555555" => {
         "veteran_first_name" => "Stan",
         "veteran_last_name" => "Lee",
         "veteran_last_four_ssn" => "2222"
       }
     }
+    allow_any_instance_of(Fakes::BGSService).to receive(:veteran_info).and_return(veteran_info)
 
     @user_download.create!(
       file_number: "55555555",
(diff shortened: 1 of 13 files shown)


lowellrex:
Documenting a few trends that I consider to be testing anti-patterns here:
1. We often set attributes of Fakes::BGSService (specifically sensitive_files and veteran_info) to values in tests which has the side effect of those values persisting as the result of calling those attributes (the way Fakes::BGSService is written I would expect each of those attributes to be nil unless they are specifically set within a test). We should probably instead be mocking the return values of those attributes `allow(Fakes::BGSService).to accept(:sensitive_files).and_return( ... )` in order to properly scope the changes to just the test scenario or file.
1. [Webdriver caching in Capybara](http://www.rubydoc.info/github/jnicklas/capybara/Capybara/Session#reset!-instance_method) can cause sessions and requests from previous tests to infect later tests. I don't yet know of a good universal fix for this ([reseting the webdriver cache does not appear straightforward](https://makandracards.com/makandra/16117-how-to-clear-cookies-in-capybara-tests-both-selenium-and-rack-test)), but beware that this could cause test false failures/successes.
1. Hard coded counters. If we are testing whether some value increments during some process, we should get the value before, run the process, and check that the value after == value before + 1. If we want to test what the value is before the process starts we can test that separately.
1. Hard coded IDs. If a specific value has meaning (`get_weaknesses("Lowell")` spawns an error, for example), it may be worthwhile to test that value. However, we should avoid using hard coded values (say 123456789 to stand in for an SSN) in tests simply so we can address the same same entity in multiple points across the test. We should prefer to generate random values and assign them to variables (something like `def rand_ssn; (Array.new(9) { ("0".."9").to_a.sample }).join; end` in the SSN example, or caseflow uses a third party Fakes library). This prevents conflicts across tests and protects us from instances where tests incorrectly pass/fail due to some fluke of the particular value we chose.
1. Function and FeatureToggles leaking state. Every test needs to assume a clean slate and explicitly set up the environment that it expects to exist (assign any Functions we expect a user to have, turn on any FeatureToggles we require, load database rows as needed). Similarly tests need to clean up after themselves by resetting the environment (or every test should clean up before running). It will be slower, but it may be easier to just have a universal `reset_test_env` script that clears the database, redis, browser cache, etc. that we could run before/after each test.

spec/jobs/save_files_in_s3_spec.rb
```diff
@@ -14,10 +14,13 @@
         veteran_last_name: "Washington"
       )
     end
+    let(:document_id) do
+      "{#{random_int_of_length(4)}-#{random_int_of_length(4)}}"
```
aroltsch: @lowellrex we already have something like that, let's re-use that and put it in spec/rails_helper.rb 
https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/83167b54e360eb798b7f2bfd32d2f42dc0ed69d3/spec/services/download_documents_spec.rb#L5

spec/requests/api/v1/file_spec.rb
```diff
@@ -40,6 +40,7 @@
   end
 
   before do
+    User.unauthenticate!
```
aroltsch: @lowellrex instead of doing it here, let's unauthenticate user before each test (similar to what is in Caseflow)
https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/rails_helper.rb#L167

spec/features/download_spec.rb
```diff
@@ -214,6 +214,9 @@ def assert_coachmark_does_not_exist
 
     search = Search.where(user: @user).first
     expect(search).to be_access_denied
+
+    Fakes::BGSService.sensitive_files = nil
```
aroltsch: Is it something we can move to rails_helper?

spec/jobs/save_files_in_s3_spec.rb
lowellrex: `generate_id` returns an 8 digit hex string (`"f4b25d4d1"` or similar) whereas this document id is expected to be formatted like `"{1893-5316}"`. I'm happy to reuse the code in download_documents_spec (test still passes when using the `SecureRandom.hex` function) if you'd prefer but I think we should aspire to have our test data appear as much like real data as possible.

spec/features/download_spec.rb
lowellrex: Changed this to use `allow()` for mocking instead of setting a class variable.

spec/jobs/save_files_in_s3_spec.rb
aroltsch: We can make it formatted like "{1893-5316}", let's just move it to the rails_helper so we can re-use it.

Issue #778, opened 2017-12-15
Remove Document.filepath column from database


lowellrex:
#736 replaced the use of the `Document.filepath` attribute in favor of a `Document.path` method that generates the filepath based on other attributes of the document (document ID and download path). We no longer use `Document.filepath` attribute anywhere, however there may be some jobs that were started prior to this change being pushed to production that depend on the `Document.filepath` attribute so we decided to delay removing the column so that those jobs would not fail.

This ticket exists to remove the filepath column after all of jobs that may have been started before the change in #736 have completed. I don't imagine that any of those jobs would run longer than a few minutes or hours, so this change is fine to be pushed live any time after that.

## Acceptance Criteria
* filepath column does not exist in documents table of database

PR #797, opened 2017-12-21
Remove documents filepath column from db

lowellrex:
Connects #778.

Confirmed that we are using the new strategy to generate document file paths that we incorporated in #736 and released on Monday by hopping into the rails console on a production box and running
```
Document.find_by_sql("select * from documents where filepath is not null order by id desc limit 1")[0].created_at
```
revealing that the most recently created document row that has a filepath value was created at ~10am Monday which lines up with when the initial change (#774) was pushed live.

### DIFF
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -164,7 +164,7 @@ class Download < ActiveRecord::Base
   def reset!
     Download.transaction do
       update_attributes!(status: :pending_documents)
-      documents.update_all(filepath: nil, download_status: 0, completed_at: nil)
+      documents.update_all(download_status: 0, completed_at: nil)
     end
   end
 
--- a/app/models/download.rb
+++ b/app/models/download.rb
@@ -164,7 +164,7 @@ class Download < ActiveRecord::Base
   def reset!
     Download.transaction do
       update_attributes!(status: :pending_documents)
-      documents.update_all(download_status: 0, completed_at: nil)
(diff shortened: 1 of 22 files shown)


app/models/download.rb
```diff
@@ -164,7 +164,7 @@ def package_filename
   def reset!
     Download.transaction do
       update_attributes!(status: :pending_documents)
-      documents.update_all(filepath: nil, download_status: 0, completed_at: nil)
+      documents.update_all(download_status: 0, completed_at: nil)
```
aroltsch: Ugh, can we have a situation where a migration will finish but the application will still use the "old" code? Maybe we need to deploy this change first and then the migration.

lowellrex: Hmmm I would guess this is unlikely but it definitely reveals that I don't know exactly what the build process looks like. I don't imagine we would start the rails server until we have run all the migrations, but that is a total guess. Where would the build code live? [Appeals deployment repo](https://github.com/department-of-veterans-affairs/appeals-deployment)?

aroltsch: https://wiki.openstack.org/wiki/DBMigrationBestPractices#Best_Practices

aroltsch: we should separate this change from the migration to ensure backwards compatibility.

lowellrex:
This PR should not be deployed until [the related PR](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/800) that removes all references to `Document.filepath` from the app logic is running live.

PR #799, opened 2017-12-21
V2 api - use VBMS ListDocuments

aroltsch:
Since we can't turn on VBMS eFolder V1 yet, we need to continue using the old eDocument service for the v2 api. 

Tested using VBMS and VVA file numbers pointed to UAT env.

### DIFF
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -51,10 +51,10 @@ class Record < ActiveRecord::Base
     find_or_initialize_by(manifest_source: manifest_source, external_document_id: document.document_id).tap do |t|
       t.assign_attributes(
         type_id: document.type_id,
-        type_description: document.type_description,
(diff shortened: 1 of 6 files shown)


app/models/record.rb
```diff
@@ -51,10 +51,10 @@ def self.create_from_external_document(manifest_source, document)
     find_or_initialize_by(manifest_source: manifest_source, external_document_id: document.document_id).tap do |t|
       t.assign_attributes(
         type_id: document.type_id,
-        type_description: document.type_description,
+        type_description: document.try(:type_description),
```
aroltsch: We need these `try` statements; because the old VBMS API is using a Document class instead of OpenStruct so if we call a method that doesn't exist on the class it will break and methods `type_description, jro, restricted` do not exist on the Document model in connect_vbms.

PR #802, opened 2017-12-27
Set manifest pending status synchronously

aroltsch:
1. Set manifest status to pending before starting asynchronous job
2. Catch StandardError while fetching manifest to change the status from pending to failed.

### DIFF
--- a/app/models/manifest_source.rb
+++ b/app/models/manifest_source.rb
@@ -15,6 +15,7 @@ class ManifestSource < ActiveRecord::Base
 
   def start!
     return if current? || pending?
+    update(status: :pending)
     V2::DownloadManifestJob.perform_later(self)
   end
 
--- a/app/models/serializers/v2/manifest_serializer.rb
+++ b/app/models/serializers/v2/manifest_serializer.rb
@@ -22,7 +22,9 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
         id: document.id,
         type_id: document.type_id,
         received_at: document.received_at,
-        external_document_id: document.external_document_id
+        external_document_id: document.external_document_id,
+        created_at: document.created_at,
+        updated_at: document.updated_at
       }
     end
   end
--- a/app/services/manifest_fetcher.rb
+++ b/app/services/manifest_fetcher.rb
@@ -3,10 +3,10 @@ class ManifestFetcher
 
   attr_accessor :manifest_source
 
-  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError].freeze
+  # Catch StandardError in case there is an error to avoid manifests being stuck in pending state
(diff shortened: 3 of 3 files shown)


app/models/manifest_source.rb
```diff
@@ -15,6 +15,7 @@ class ManifestSource < ActiveRecord::Base
 
   def start!
     return if current? || pending?
+    update(status: :pending)
```
aroltsch: Set status to `pending` before starting the job; otherwise the server might respond before the status is set to `pending`

app/services/manifest_fetcher.rb
```diff
@@ -3,10 +3,10 @@ class ManifestFetcher
 
   attr_accessor :manifest_source
 
-  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError].freeze
+  # Catch StandardError in case there is an error to avoid manifests being stuck in pending state
+  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError, StandardError].freeze
```
aroltsch: Catch `StandardError` when fetching manifest and set the status to failed if anything otherwise the manifest will be stuck in pending state.

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -22,7 +22,9 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
         id: document.id,
         type_id: document.type_id,
         received_at: document.received_at,
-        external_document_id: document.external_document_id
+        external_document_id: document.external_document_id,
+        created_at: document.created_at,
```
mdbenjam: Why add these to the serializer?

app/services/manifest_fetcher.rb
mdbenjam: What happens if the job just dies without an error being thrown. I.E. box shuts down. Presumably our job system will just restart it right?

app/models/serializers/v2/manifest_serializer.rb
aroltsch: Makes it easier for testing purposes.

app/services/manifest_fetcher.rb
aroltsch: I don't think our current approach has a way to restart the job if the box shuts down in th middle of the job, however, if we move to SQS/Shoryuken it will handle it the restart. 
```
With SQS, it is much more durable. They also have the notion of in-flight messages. This means that the message is pulled off the queue but never deleted until the delete command is sent for that message id. So if you lose your worker mid-processing of the event, that event isn’t lost for good. The message will be timed out after being in-flight for 5 minutes and then dropped back onto the available queue. ```

mdbenjam: Sounds good, moving to SQS should make this airtight 🤞

PR #815, opened 2018-01-03
Call new Image Magick Service

mdbenjam:
This PR does two things:

1) Removes code calling Image Magick directly and instead calls our new image conversion service.
2) Changes both the v1 and v2 APIs to fail gracefully if conversion fails. If it fails we now return the original file (i.e. TIFF image) instead of failing.

**Test Plan**
- Ensure the `:convert_tiff_images` feature flag is turned on.
- Testing locally by commenting out this line: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/mdbenjam-call-convert-service/lib/fakes/document_service.rb#L88 so that we only return TIFFs then going through the UI, new and old document APIs to ensure the following:
   - If the conversion service is running (see new readme instructions), then the images are converted from TIFF to PDF
   - If the conversion service is not running or throws an error, then the TIFF images are returned.
- Once someone takes an initial look and thinks this is the right direction, I'll do more testing in UAT to ensure this works on real cases.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -75,6 +75,8 @@ gem 'request_store'
 
 gem "mini_magick"
 
+gem 'curb'
+
 gem 'connect_vbms', git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", ref: "e7fc3e6825f45176950c0b7ce37b38068ec5aea8"
 gem 'connect_vva', git: "https://github.com/department-of-veterans-affairs/connect_vva.git", ref: "f7036e7d3c17aae6e66e345051ea61c6badc5de2"
 gem 'bgs', git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git", :branch => '9266698668bf361bf5df06990d06da59abb7c608'
--- a/app/models/fetcher.rb
+++ b/app/models/fetcher.rb
@@ -16,26 +16,26 @@ class Fetcher
     ImageConverterService.converted_mime_type(document.mime_type)
   end
 
-  def converted_content
+  def cached_converted_content
     document.converted_mime_type = converted_mime_type
-    @converted_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
-                                                 service: :s3,
-                                                 name: "fetch_content") do
+    @cached_converted_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
+                                                        service: :s3,
+                                                        name: "fetch_content") do
       S3Service.fetch_content(document.s3_filename)
     end
   end
 
-  def original_content
+  def cached_original_content
     document.converted_mime_type = document.mime_type
-    @original_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
-                                                service: :s3,
-                                                name: "fetch_content") do
+    @cached_original_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
+                                                       service: :s3,
+                                                       name: "fetch_content") do
       S3Service.fetch_content(document.s3_filename)
     end
   end
 
   def cached_content
-    converted_content || original_content
+    cached_converted_content || cached_original_content
   end
 
   def download_from_service
--- a/app/models/fetcher.rb
+++ b/app/models/fetcher.rb
@@ -12,19 +12,45 @@ class Fetcher
 
   private
 
-  def cached_content
-    @cached_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
-                                              service: :s3,
-                                              name: "fetch_content") do
+  def converted_mime_type
+    ImageConverterService.converted_mime_type(document.mime_type)
+  end
+
+  def converted_content
+    document.converted_mime_type = converted_mime_type
+    @converted_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
+                                                 service: :s3,
+                                                 name: "fetch_content") do
       S3Service.fetch_content(document.s3_filename)
     end
+  end
+
+  def original_content
+    document.converted_mime_type = document.mime_type
+    @original_content ||= MetricsService.record("S3: fetch content for: #{document.s3_filename}",
+                                                service: :s3,
+                                                name: "fetch_content") do
+      S3Service.fetch_content(document.s3_filename)
+    end
+  end
+
+  def cached_content
+    converted_content || original_content
   end
 
   def download_from_service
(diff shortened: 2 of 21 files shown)


Gemfile
```diff
@@ -75,6 +75,8 @@ gem 'request_store'
 
 gem "mini_magick"
 
+gem 'curb'
```
mdbenjam: Not sure if I need to install this. But I couldn't figure out how to post files through HTTPI

app/services/download_documents.rb
```diff
@@ -80,9 +80,11 @@ def download_contents(save_locally: true)
 
   def fetch_from_s3(document)
     # if the file exists on the filesystem, skip
-    return if File.exist?(document.path)
+    # return if File.exist?(document.path)
 
-    S3Service.fetch_file(document.s3_filename, document.path)
```
mdbenjam: Not sure the best thing to do here... But by always going through the fetcher, we know what the document type is. If it's saved locally we're not sure. Though this change will cause us to download the same doc twice from S3.

app/services/image_converter_service.rb
```diff
@@ -20,36 +22,26 @@ def self.converted_mime_type(mime_type)
     mime_type
   end
 
-  private
+  # :nocov:
```
mdbenjam: Not sure how to test this method without the service running.

app/services/download_documents.rb
```diff
@@ -79,10 +79,9 @@ def download_contents(save_locally: true)
   end
 
   def fetch_from_s3(document)
-    # if the file exists on the filesystem, skip
-    return if File.exist?(document.path)
-
-    S3Service.fetch_file(document.s3_filename, document.path)
+    # All the content should already be in S3.
+    fetch_result = document.fetch_content!(save_document_metadata: true)
+    document.save_locally(fetch_result[:content])
```
aroltsch: Why are we saving it locally here?

app/services/record_fetcher.rb
```diff
@@ -19,11 +19,37 @@ def process
...
+    record.converted_mime_type = record.mime_type
+    content
+  end
+
+  def converted_mime_type
+    ImageConverterService.converted_mime_type(record.mime_type)
+  end
+
+  def cached_converted_content
+    record.converted_mime_type = converted_mime_type
```
aroltsch: So converted_mime_type might actually return the original mime type so when you call cached_converted_content it is actually returning the file with original mime type?

app/services/record_fetcher.rb
```diff
@@ -19,11 +19,37 @@ def process
 
   private
 
-  def cached_content
-    @cached_content ||= MetricsService.record("S3: fetch content for: #{record.s3_filename}",
-                                              service: :s3,
-                                              name: "fetch_content") do
+  def convert_to_pdf(content)
```
aroltsch: Is this code identical to v1 code? Maybe it makes sense to move it into a shared file.

mdbenjam: I would say since we're planning on removing v1, I don't think it makes sense to put too much effort into it.

app/services/download_documents.rb
mdbenjam: So if we call S3Service.fetch_file(document.s3_filename, document.path) we go outside of the fetcher it doesn't know to look for both versions of the file. Calling save locally saves it to the same path as the above fetch_file. The downside, is this could download the file twice.

lib/fakes/document_service.rb
```diff
@@ -85,7 +85,7 @@ def self.sleep_manifest_load(wait)
 
   # Randomly send a pdf or tiff
   def self.document_type
-    return { ext: "pdf", mime_type: "application/pdf" } if rand(2)
+    return { ext: "pdf", mime_type: "application/pdf" } if rand(2) == 1
```
mdbenjam: Apparently 0 is truthy...

app/services/metrics_service.rb
```diff
@@ -38,6 +38,8 @@ def self.record(description, service: nil, name: "unknown")
       increment_datadog_counter("request_error", service, name)
     end
 
+    Rails.logger.info("RESCUED #{description}")
```
mdbenjam: Just so we can better track failing calls in the logs.

app/models/fetcher.rb
```diff
@@ -12,19 +12,48 @@ def content(save_document_metadata: true)
...
+    @cached_converted_content ||= cached_content
+  end
+
+  def cached_original_content
+    document.s3_stored_file_mime_type = document.mime_type
+    @cached_original_content ||= cached_content
+  end
+
+  def cached_converted_or_original_content
+    @cached_converted_or_original_content ||= (cached_converted_content || cached_original_content)
```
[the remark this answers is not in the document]
mdbenjam: We need to set it because this call comes after the `s3_stored_file_mime_type` is set to the new converted mime_type.

app/services/record_fetcher.rb
```diff
@@ -19,6 +19,18 @@ def process
 
   private
 
+  def convert_to_pdf(content)
+    # Every time we can't find the file in S3 and are asked to convert the version from
+    # VBMS, we try to do the conversion, even if it failed last time. In case some change
+    # has enabled it to be converted this time.
+    converted_content = ImageConverterService.new(image: content, mime_type: record.mime_type).process
```
aroltsch: Even if the file was pdf originally, it will call ImageConverterService and then immediately return. Do we still want to set `conversion_status` to success even if we didn't convert anything? Maybe it makes sense to update conversion_status in `ImageConverterService` and you can catch `ImageConverterService::ImageConverterError` in the service itself also.

app/models/record.rb
```diff
@@ -68,7 +74,7 @@ def fetcher
   end
 
   def mime
-    MIME::Types[ImageConverterService.converted_mime_type(mime_type)].first
+    MIME::Types[conversion_success? ? ImageConverterService.converted_mime_type(mime_type) : mime_type].first
```
aroltsch: This is beautiful.

db/migrate/20180106211625_add_conversion_status_to_records.rb
```diff
@@ -0,0 +1,5 @@
+class AddConversionStatusToRecords < ActiveRecord::Migration
+  def change
+    safety_assured { add_column :records, :conversion_status, :integer, default: 0 }
```
mdbenjam: This is safe because there are 0 records in prod, so the table won't have to be locked.

PR #821, opened 2018-01-08
Anya/810 start retrieving files

mdbenjam:
connects #810 

New Routes:
```
POST "manifests", to: "manifests#start"
GET "manifests", to: "manifests#progress"
POST "manifests/:manifest_id/files_downloads", to: "files_downloads#start"
GET "manifests/:manifest_id/files_downloads", to: "files_downloads#progress"
```

Note: TODO comments will be addresses in the subsequent PR. 
Added `/api/v2/manifests/manifest_id/records`. This API endpoint is used by eX UI and will kick off
the job to download files from VBMS/VVA and save in s3 and create a zip file.

### DIFF
--- a/app/controllers/api/v2/manifests_controller.rb
+++ b/app/controllers/api/v2/manifests_controller.rb
@@ -3,15 +3,18 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
   before_action :validate_header
   before_action :validate_access
 
-  def index
+  def start
+    manifest.start!
+    render json: json_manifests
+  end
+
+  def progress
     render json: json_manifests
   end
 
   private
 
   def json_manifests
-    manifest.start!
-
     ActiveModelSerializers::SerializableResource.new(
       manifest,
       each_serializer: Serializers::V2::ManifestSerializer
@@ -32,7 +35,7 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
   end
 
   def manifest
-    Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
+    @manifest ||= Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
   end
 
   def bgs_service
--- a/app/controllers/api/v2/manifests_controller.rb
+++ b/app/controllers/api/v2/manifests_controller.rb
@@ -32,7 +32,7 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
   end
 
   def manifest
-    @manifest ||= Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
+    Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
   end
 
   def bgs_service
--- a/app/controllers/api/v2/records_controller.rb
+++ b/app/controllers/api/v2/records_controller.rb
@@ -1,6 +1,8 @@
 class Api::V2::RecordsController < Api::V1::ApplicationController
-  before_action :validate_access
+  before_action :validate_access, only: :show
 
+  # /api/v2/records/:id
+  # This API returns document content
   def show
     result = record.fetch!
     return document_failed if record.failed?
@@ -16,7 +18,24 @@ class Api::V2::RecordsController < Api::V1::ApplicationController
     )
   end
 
+  # /api/v2/manifests/manifest_id/records
+  # This API endpoint is used by eX UI and will kick off
+  # the job to download files from VBMS/VVA and save in s3
+  def index
+    return record_not_found unless user_manifest
+    render json: json_records
+  end
+
   private
+
+  def json_records
+    user_manifest.save_files_and_package!
+
+    ActiveModelSerializers::SerializableResource.new(
+      user_manifest,
+      each_serializer: Serializers::V2::UserManifestSerializer
+    ).as_json
+  end
 
   def document_failed
     render json: {
@@ -29,6 +48,11 @@ class Api::V2::RecordsController < Api::V1::ApplicationController
 
   def enable_caching
     expires_in 30.days, public: true
+  end
+
+  def user_manifest
+    UserManifest.includes(:manifest)
+                .find_by(manifest_id: params[:manifest_id], user_id: current_user.id)
   end
 
   def record
--- a/app/controllers/api/v2/records_controller.rb
+++ b/app/controllers/api/v2/records_controller.rb
@@ -1,5 +1,5 @@
 class Api::V2::RecordsController < Api::V1::ApplicationController
-  before_action :validate_access, only: :show
+  before_action :validate_access
 
   # /api/v2/records/:id
   # This API returns document content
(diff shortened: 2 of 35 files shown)


app/controllers/api/v2/manifests_controller.rb
```diff
@@ -32,7 +32,7 @@ def validate_header
   end
 
   def manifest
-    @manifest ||= Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
+    Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
```
aroltsch: Need to reload a new one from memory each time to get the latest status.

app/controllers/api/v2/records_controller.rb
```diff
@@ -31,6 +50,11 @@ def enable_caching
     expires_in 30.days, public: true
   end
 
+  def user_manifest
+    UserManifest.includes(:manifest)
```
aroltsch: Need to reload a new one from memory each time to get the latest status.

app/models/manifest.rb
```diff
@@ -40,8 +42,7 @@ def veteran
 
   def self.find_or_create_by_user(user:, file_number:)
     manifest = Manifest.find_or_create_by(file_number: file_number)
-    # Create a record every time for auditing purposes
-    manifest.user_manifests.create(user: user)
+    manifest.user_manifests.find_or_create_by(user: user)
```
aroltsch: This method is called many times during polling so we don't want to re-create user_manifests each time for the same user.

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -17,7 +17,7 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
   end
 
   attribute :records do
-    object.records.order(:id).map do |document|
+    object.records.map do |document|
```
aroltsch: Records will be ordered by the received_at timestamp.

app/jobs/v2/save_files_in_s3_job.rb
```diff
@@ -2,7 +2,9 @@ class V2::SaveFilesInS3Job < ActiveJob::Base
   queue_as :default
 
   def perform(manifest_source)
-    manifest_source.records.each(&:fetch!)
+    manifest_source.records.each do |record|
```
mdbenjam: I think we should use a more robust locking mechanism. There's a subtle race condition (a real edge case). If two of these jobs are happening at the same time, both could enter the `fetch!` method at the same time, before the status is marked as pending. Then you could be fetching the same document twice at the same time.

I think either locking the job so that only one thread can enter for a given manifest_source, or locking the fetch! method so only one thread can enter for a given record might be more robust.

app/models/user_manifest.rb
```diff
@@ -1,6 +1,27 @@
...
+
+  enum status: {
+    initialized: 0,
+    pending: 1,
+    finished: 2,
+    failed: 3
+  }
+
+  def save_files_and_package!
+    return if current? || pending?
```
mdbenjam: We should really lock the DB when we test if it's `pending?`otherwise there's a small window for a race condition here.

app/services/record_fetcher.rb
```diff
@@ -6,7 +6,11 @@ class RecordFetcher
   EXCEPTIONS = [VBMS::ClientError, VVA::ClientError].freeze
 
   def process
-    return cached_content if cached_content
+    if cached_content
+      record.update(status: :success)
```
mdbenjam: Why are we updating success here? I'm unsure it's necessary.

aroltsch: That's for documents that were already cached in s3 before we release this code. The initial status is `initialized`. If we get a cached copy, the state will stay in `initialized` which is wrong.

app/jobs/v2/save_files_in_s3_job.rb
aroltsch: I agree we need some kind of a locking mechanism. However if we lock the entire job, won't it create problems? For example, one user #1 starts the job, another user #2 makes an HTTP request to start the same job, the lock for the job is taken so the user #2 can't start the job and will need to wait for the job to finish for the lock to be released while in http request?

app/jobs/v2/package_files_job.rb
```diff
@@ -0,0 +1,17 @@
+class V2::PackageFilesJob < ActiveJob::Base
+  queue_as :default
+
+  def perform(files_download)
+    # TODO: keep calling fetch if status is pending and create zip file at the end
+    files_download.records.each(&:fetch!)
```
mdbenjam: Might be useful to reuse-`save_files_in_s3_job.rb` synchronously. But we can leave that till the next PR.

app/models/files_download.rb
```diff
@@ -0,0 +1,27 @@
...
+  enum status: {
+    initialized: 0,
+    pending: 1,
+    finished: 2,
+    failed: 3
+  }
+
+  def start!
+    return if current? || pending?
+    update(status: :pending)
```
mdbenjam: Like we mentioned very small race condition here. We can probably ignore it, but we will get a stale object error if something falls through.

app/models/serializers/v2/files_download_serializer.rb
```diff
@@ -0,0 +1,20 @@
...
+  attribute :fetched_files_at
+
+  attribute :records do
+    object.records.map do |document|
+      {
+        id: document.id,
+        type_id: document.type_id,
+        status: document.status,
+        received_at: document.received_at,
+        external_document_id: document.external_document_id,
```
mdbenjam: If you merge the other one first, this should probably be series_id.

app/services/manifest_fetcher.rb
```diff
@@ -3,8 +3,7 @@ class ManifestFetcher
 
   attr_accessor :manifest_source
 
-  # Catch StandardError in case there is an error to avoid manifests being stuck in pending state
-  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError, StandardError].freeze
+  EXCEPTIONS = [VBMS::ClientError, VVA::ClientError].freeze
```
mdbenjam: Why are we taking out StandardError?

app/jobs/v2/package_files_job.rb
aroltsch: Yup! I'll see if I can do it in the next PR!

app/services/manifest_fetcher.rb
aroltsch: I moved it to the FetchManifestJob, to be higher up.

app/models/files_download.rb
aroltsch: Yes, I will try to handle these race conditions in the next PR

app/models/serializers/v2/files_download_serializer.rb
aroltsch: Fixed.

spec/requests/api/v2/records_spec.rb
```diff
@@ -37,7 +37,7 @@
     end
 
     context "when user has access to the corresponding manifest record" do
-      let!(:user_manifest) { UserManifest.create(user: current_user, manifest: manifest) }
+      let!(:files_download) { FilesDownload.create(user: current_user, manifest: manifest) }
```
mdbenjam: nit, we changed the let, but it doesn't seem like it's used anywhere.

aroltsch: you need this record for permissions checking

PR #822, opened 2018-01-09
Attempting to update rubocop

mdbenjam:
Our version of Rubocop had a vulnerability. Upgrading causes new rules to be added. I mimicked most of what was done in: https://github.com/department-of-veterans-affairs/caseflow/pull/4312/files. Where only one thing was flagged by a rule, I fixed that one instance, and called it out in the PR.

**Test Plan**
- [ ] The only way to really test this is to rely on automated tests. Quick smoke test also passed.

### DIFF
--- a/app/controllers/api/v1/documents_controller.rb
+++ b/app/controllers/api/v1/documents_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class Api::V1::DocumentsController < Api::V1::ApplicationController
   before_action :can_access?
 
@@ -22,7 +24,7 @@ class Api::V1::DocumentsController < Api::V1::ApplicationController
     status = 500
     message = "Caseflow eFolder failed to fetch document contents."
 
-    if error_kind == :vva_error || error_kind == :vbms_error
+    if [:vva_error, :vbms_error].include?(error_kind)
       status = 502
       message = "An upstream dependency failed to fetch document contents."
     end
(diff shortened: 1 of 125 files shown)


lib/tasks/support/shell_command.rb
```diff
@@ -9,7 +9,7 @@ def self.run(command)
         puts(line)
       end
 
-      success = thread.value == 0
+      success = thread.value.zero?
```
mdbenjam: Changed by hand. Tested by running rake commands which all rely on it.

app/services/document_creator.rb
```diff
@@ -1,13 +1,14 @@
 class DocumentCreator
   include ActiveModel::Model
 
-  attr_accessor :manifest_source, :external_documents
+  attr_accessor :manifest_source
+  attr_writer :external_documents
```
mdbenjam: Changed by hand. Should work if tests pass since external_documents is written to in several jobs.

app/models/css_authentication_session.rb
```diff
@@ -2,14 +2,15 @@ class CssAuthenticationSession
   include ActiveModel::Model
   include ActiveModel::Serializers::JSON
 
-  attr_accessor :id, :css_id, :email, :name, :roles, :station_id
+  attr_accessor :id, :name, :roles, :station_id
```
mdbenjam: Changed by hand.

app/controllers/api/v1/documents_controller.rb
```diff
@@ -22,7 +22,7 @@ def document_download_failed(error_kind)
     status = 500
     message = "Caseflow eFolder failed to fetch document contents."
 
-    if error_kind == :vva_error || error_kind == :vbms_error
+    if [:vva_error, :vbms_error].include?(error_kind)
```
mdbenjam: Changed by hand. Tested in `documents_spec.rb` test: `returns 502 if there is a VBMS error`

Issue #818, opened 2018-01-04
[BE] Add series_id and version to the records table


aroltsch:
# AC
1. Rename `external_document_id` in records table to `version_id` 
1. Add `series_id` and `version` to the response in connect_vbms gem
1. Add `series_id` and `version` to the `records` table in eX

PR #823, opened 2018-01-10
Anya/818 add series id and version

aroltsch:
connects #818 
This PR adds series_id and version. The VBMS eFolder new API will only be integrated into eX v2 code. 
Reader will need to start using series_id to identify the document. Need to think how to migrate Reader.


db/migrate/20180109194733_rename_external_document_id_to_version_id_in_records.rb
```diff
@@ -0,0 +1,5 @@
+class RenameExternalDocumentIdToVersionIdInRecords < ActiveRecord::Migration
+  def change
+    rename_column :records, :external_document_id, :version_id
```
aroltsch: These migrations are safe because nothing is using this code in production.

Issue #812, opened 2018-01-03
[FE] Add React infrastructure to efolder


lowellrex:
Set up the react infrastructure for efolder similar to what was done for the [main caseflow repo](https://github.com/department-of-veterans-affairs/caseflow/pull/469). The react environment should should be set up such that subsequent react front-end work can be limited to adding/altering only those files required to strictly implement the features they are supposed to. 

## Acceptance Criteria:
* React environment is set up in efolder repo

lowellrex:
There is still work to be done in order to fully set up the react infrastructure for efolder express (webpack, etc.), but I am hoping we will be able to knock that stuff out as we work on the other efolder react tickets. I am moving this ticket back to "Current Sprint" as a reminder of that work still to be accomplished.

PR #824, opened 2018-01-10
First pass at efolder backend serving sample react app

lowellrex:
Connects #812.

First pass at setting up the core infrastructure from which we will build the rest of the efolder express front-end. Still plenty left to do (add the `yarn build` step to the deploy pipeline, set up hot reloading, incorporate components of [appeals-frontend-toolkit](https://github.com/department-of-veterans-affairs/appeals-frontend-toolkit), use react-router to route on the frontend, actually write the components and containers, etc.), but I wanted to get this out for others to critique as I think it contains the key elements of this approach.

I've got plenty of open questions about this approach, so if you've got thoughts on any of the following, let me know.

### Passing data from the back to the front-end
Currently I accomplish this by calling `efx.init(<%= raw( initial_react_data ) %>);` in an HTML template. This seemed like an expedient way to get this up and running, but I have no idea if there are better ways to solve that problem.

### Organization
Things are kind of all over the place right now. `src/index.jsx` kicks off the whole react app, but there isn't a clear distinction between what it is doing and what `InitContainer` does (this is partially because I foresee InitContainer doing the routing which it doesn't yet). Are the files separated in an otherwise sensible fashion? Files contain too many different pieces? Pieces exist in several files when they could exist in a single spot?

### Idiomatic React
This is a first hacks at efolder express as a react app and my first attempt at whipping up a react app. I am sure I have blind spots where I did things in a totally haphazard fashion that do not align with the idiomatic way of writing react apps. We should fix those. Please identify those for me because I almost certainly don't know what those are yet.


app/controllers/downloads_controller.rb
```diff
@@ -17,6 +17,15 @@ def create
...
   end
 
+  def react
+    render "_react", layout: false
+  end
+
+  def initial_react_data
+    { text: "placeholder text" }.to_json
+  end
+  helper_method :initial_react_data
```
lowellrex: This will undoubtedly move (`DownloadsController` won't exist after we get rid of all v1 code), but I something like these methods will exist in the final version to serve the html page that contains the react app (`react()`) and to pass whatever data we want to the backend from the front (`initial_react_data()`).

Also, before we merge this to master we probably aught to hide this behind a feature flag.

app/views/downloads/_react.html.erb
```diff
@@ -0,0 +1,10 @@
+<html>
+  <head>
+    <title>Getting Started</title>
+    <%= javascript_include_tag 'react-app.js' %>
+  </head>
+  <body>
+    <div id="efx_app_v2"></div>
+    <script>efx.init(<%= raw( initial_react_data ) %>);</script>
+  </body>
+</html>
```
lowellrex: Just a placeholder template right now, but I don't imagine this will change all that much. I figure we will build the skeleton of the application UI (what currently exists in [the application layout](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/8043995a00e4ef8fd28ccbf1c76a1b6d0c03d1dd/app/views/layouts/application.html.erb)) in react (possibly as a shared component among the different caseflow apps?).

client/src/index.jsx
```diff
@@ -0,0 +1,18 @@
...
+import { render } from 'react-dom';
+import { Provider } from 'react-redux';
+import { createStore } from 'redux';
+
+import reducer from './reducer';
+import InitContainer from './containers/InitContainer';
+
+module.exports = {
+  init: function(props) {
+    var initState = { text: "some text" };
```
nickheiner-usds: There are a few things in this PR that will be cleaned up once you pull in [our shared lint config](https://github.com/department-of-veterans-affairs/eslint-config-appeals). You can Caseflow's usage for an example.

client/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "efolder-express-ui",
+  "private": true,
+  "version": "1.0.0",
+  "license": "UNLICENSED",
+  "main": "index.jsx",
```
nickheiner-usds: We don't need a `main` field because this node module will not be consumed by any other node modules.

nickheiner-usds:
> Still plenty left to do (add the `yarn build` step to the deploy pipeline, set up hot reloading, incorporate components of [appeals-frontend-toolkit](https://github.com/department-of-veterans-affairs/appeals-frontend-toolkit), use react-router to route on the frontend, actually write the components and containers, etc.)

I would certainly invite you to do multiple manageable PRs instead of feeling like you need to dump everything in to one! :smile:

app/views/downloads/_react.html.erb
```diff
@@ -0,0 +1,10 @@
+<html>
+  <head>
+    <title></title>
```
nickheiner-usds: We should have a real `title` value.

client/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "scripts": {
+    "build": "webpack --config webpack.config.js"
+  },
+  "dependencies": {
+    "babel-loader": "^7.1.2",
```
nickheiner-usds: This warning occurs when running `yarn`. We should be sure to clean it up before merging.

```
warning "babel-loader@7.1.2" has unmet peer dependency "babel-core@6 || 7 || ^7.0.0-alpha || ^7.0.0-beta || ^7.0.0-rc".
```

client/src/index.jsx
nickheiner-usds: Also, we should add https://eslint.org/docs/rules/prefer-const to this project. I'd like to enable it in Caseflow, but there's too much code that does not abide by it to migrate. We can add it up-front here.

app/views/downloads/_react.html.erb
mdbenjam: Yeah we already have header and footer components, which is the majority of what is in the application layout.

client/webpack.config.js
```diff
@@ -0,0 +1,22 @@
+var path = require('path');
```
mdbenjam: What's the advantage of creating a new webpack config here rather than using the shared one?

client/src/index.jsx
```diff
@@ -0,0 +1,18 @@
...
+    var store = createStore(reducer, {...initState, ...props});
+
+    render( <Provider store={store}>
```
mdbenjam: Any reason to not use the ReduxBase file in the extracted package?

app/controllers/downloads_controller.rb
mdbenjam: I'd just add a feature toggle now so we don't forget later.

client/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
```
nickheiner-usds: Cool. Can you turn on [prefer-const](https://eslint.org/docs/rules/prefer-const) as well?

lowellrex: I think it is on by default, but I will confirm and add it in if it is not.

lowellrex: Changed prefer-const from a warning to an error.

app/controllers/downloads_controller.rb
```diff
@@ -17,6 +17,27 @@ def create
...
+      redirect_to "/"
+    end
+  end
+
+  def initial_react_data
+    { text: "placeholder text" }.to_json
+  end
+  helper_method :initial_react_data
+
+  def can_access_react_app?
```
nickheiner-usds: Can't this be simplified?

```rb
  def can_access_react_app?
    FeatureToggle.enabled?(:efolder_react_app, user: current_user) || Rails.env.development?
  end
```

client/.eslintrc.js
```diff
@@ -0,0 +1,9 @@
...
+  env: {
+    browser: true,
+    mocha: true
```
nickheiner-usds: We do not want `mocha` to be on for all our code. It should only be on for the tests. (This is a mistake in Caseflow.)

nickheiner-usds:
This LGTM as a start, but I'll let @mdbenjam give final signoff.

lowellrex:
FeatureToggle is working (tested in UAT). However, accessing /react if the FeatureToggle is activated results in an error "Uncaught ReferenceError: efx is not defined at react:8" because we are not currently spitting the bundle out to the assets directory. This is a problem that will have to be solved before too long (with webpacker? react-on-rails?) and certainly before we go live. For the time being, I think it is fine that we cannot test this live and we should get this merged to master so we can continue work on the related tickets.

I have been getting around this constraint by running the following command locally (from the top level caseflow-efolder directory) to set up the environment: `$ cd client && yarn run build && cd .. && rails s`

PR #830, opened 2018-01-17
Anya/810 package files part3

aroltsch:
PackageFilesJob will create a zip file without saving files on disk. It will only append files to the zip file that were downloaded successfully from VBMS or S3.


app/models/manifest.rb
```diff
@@ -47,6 +47,10 @@ def number_failed_documents
     records.failed.count
   end
 
+  def s3_filename
+    "#{id}-manifest.zip"
```
mdbenjam: So elegant. There's only ever one zip for a veteran.

app/services/zipfile_creator.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_accessor :manifest
+
+  def process
+    records = manifest.records
+    return if records.empty?
+
+    t = Tempfile.new
+    index = 0
+
+    Zip::OutputStream.open(t.path) do |z|
```
mdbenjam: https://stackoverflow.com/questions/4943757/rubyzip-vs-native-os-compression

I'm interested to see if this reduces our CPU or memory usage at all. If it doesn't we might want to try experimenting with using a zip tool directly. But this is super elegant, so seems like a good choice for now.

app/services/zipfile_creator.rb
```diff
@@ -0,0 +1,32 @@
...
+      end
+    end
+    S3Service.store_file(manifest.s3_filename, t.path, :filepath)
+    manifest.update(zipfile_size: File.size(t.path))
+    t.close
+  end
+
+  private
+
+  def unique_filename(record, index)
```
mdbenjam: Is this deterministic? Will the order always be the same? I'm wondering, could we use the record id instead of index? Might make reproducing issues easier. Though it's not a big deal.

app/services/zipfile_creator.rb
```diff
@@ -0,0 +1,32 @@
...
+    Zip::OutputStream.open(t.path) do |z|
+      records.each do |record|
+        z.put_next_entry(unique_filename(record, index))
+        content = record.fetch!
+        z.print(content) and index += 1 if content
+      end
+    end
+    S3Service.store_file(manifest.s3_filename, t.path, :filepath)
+    manifest.update(zipfile_size: File.size(t.path))
+    t.close
```
mdbenjam: we should also unlink the tempfile: http://ruby-doc.org/stdlib-1.9.3/libdoc/tempfile/rdoc/Tempfile.html#method-i-close

spec/jobs/v2/package_files_job_spec.rb
```diff
@@ -17,7 +27,8 @@
       it "sets status to finished" do
         allow(S3Service).to receive(:store_file).and_return(nil)
         subject
-        expect(S3Service).to have_received(:store_file).twice
+        # 2 times for uploading documents and 1 time for uploading a zip file
+        expect(S3Service).to have_received(:store_file).exactly(3).times
```
mdbenjam: Should we do some check of the zip file? Make sure it's actually a zip file, or has two files in it or something? Or do you think that's too detailed?

app/services/zipfile_creator.rb
aroltsch: When you call manifests.records they are always ordered by the recieved_at timestamp so yes the order will be the same.

app/models/manifest.rb
aroltsch: Yup, love love love it

spec/jobs/v2/package_files_job_spec.rb
aroltsch: I will add more tests once you approve the direction on this PR.

app/services/zipfile_creator.rb
aroltsch: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/manifest.rb#L7

spec/jobs/v2/package_files_job_spec.rb
aroltsch: I think for the PackageFilesJob it might be a bit detailed. However, in the zipfile_creator_spec.rb I will add some detailed checks

PR #833, opened 2018-01-18
Mutex manifest download job and package files job

aroltsch:
1. Handle edge case: when two users might start the same job for the same manifest/source at the same time. 
2. Add additional logic if a current user doesn't have eX role, start saving files in s3 job

### DIFF
--- a/app/jobs/v2/download_manifest_job.rb
+++ b/app/jobs/v2/download_manifest_job.rb
@@ -1,14 +1,23 @@
 class V2::DownloadManifestJob < ActiveJob::Base
   queue_as :default
 
-  def perform(manifest_source)
-    ManifestFetcher.new(manifest_source: manifest_source).process
-  # Start downloading files if it is not eX user
-  # V2::SaveFilesInS3Job.perform_later(manifest_source)
-  # Catch StandardError in case there is an error to avoid manifests being stuck in pending state
+  SECONDS_TO_AUTO_UNLOCK = 180
+
+  def perform(manifest_source, ui_user)
+    s = Redis::Semaphore.new("download_manifest_source_#{manifest_source.id}".to_s,
+                             connection: Rails.application.secrets.redis_url_sidekiq)
+
+    s.lock(SECONDS_TO_AUTO_UNLOCK)
+    return if manifest_source.current?
+
+    documents = ManifestFetcher.new(manifest_source: manifest_source).process
+
+    V2::SaveFilesInS3Job.perform_later(manifest_source) if documents.present? && !ui_user
   rescue StandardError => e
     manifest_source.update!(status: :failed)
     raise e
+  ensure
+    s.unlock
   end
 
   def max_attempts
--- a/app/jobs/v2/package_files_job.rb
+++ b/app/jobs/v2/package_files_job.rb
@@ -1,12 +1,23 @@
 class V2::PackageFilesJob < ActiveJob::Base
   queue_as :default
 
+  SECONDS_TO_AUTO_UNLOCK = 43_200
+
   def perform(manifest)
+    s = Redis::Semaphore.new("package_files_#{manifest.id}".to_s,
+                             connection: Rails.application.secrets.redis_url_sidekiq)
+
+    s.lock(SECONDS_TO_AUTO_UNLOCK)
+
+    return if manifest.recently_downloaded_files?
+
     ZipfileCreator.new(manifest: manifest).process
   # Catch StandardError in case there is an error to avoid files downloads being stuck in pending state
   rescue StandardError => e
     manifest.update!(fetched_files_status: :failed)
     raise e
+  ensure
+    s.unlock
   end
 
   def max_attempts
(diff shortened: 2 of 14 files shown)


app/jobs/v2/download_manifest_job.rb
```diff
@@ -2,14 +2,20 @@ class V2::DownloadManifestJob < ActiveJob::Base
...
   def perform(manifest_source)
+    s = Redis::Semaphore.new("download_manifest_source_#{manifest_source.id}".to_s,
+      connection: Rails.application.secrets.redis_url_sidekiq)
+    s.lock(90)
+    return if manifest_source.current?
     documents = ManifestFetcher.new(manifest_source: manifest_source).process
-    return if documents.blank?
+    s.unlock and return if documents.blank?
     DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
+    s.unlock
```
mdbenjam: We should move this to an ensure block.

aroltsch: Good point!

app/jobs/v2/download_manifest_job.rb
```diff
@@ -1,16 +1,26 @@
 class V2::DownloadManifestJob < ActiveJob::Base
   queue_as :default
 
-  def perform(manifest_source)
+  TIME_TO_AUTO_UNLOCK = 90
```
aroltsch: I wonder if 90 seconds is enough.

app/jobs/v2/download_manifest_job.rb
```diff
@@ -1,16 +1,26 @@
...
+    return if manifest_source.current?
     documents = ManifestFetcher.new(manifest_source: manifest_source).process
     return if documents.blank?
+
     DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
-  # Start downloading files if it is not eX user
-  # V2::SaveFilesInS3Job.perform_later(manifest_source)
-  # Catch StandardError in case there is an error to avoid manifests being stuck in pending state
+    # If the user is eX UI user, then don't start saving files in s3
+    V2::SaveFilesInS3Job.perform_later(manifest_source) unless (roles || []).include?("Download eFolder")
```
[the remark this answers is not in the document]
mdbenjam: I'd rather just pass a flag tot he job, not roles. I'm not sure we currently have this case, but someone _could_ have both Reader and eX roles.

app/jobs/v2/download_manifest_job.rb
mdbenjam: My thought around this. We should make this number pretty high. The downside of having a high number is if our job gets stuck, we won't release the lock. But we don't have any reason right now to think that will happen. The downside to too small a number is we could have the jobs operating at the same time, which seems bad. So I'd say high is good.

Now is 90 seconds high enough... I think it probably is. Here's numbers from the current manifest endpoint right now: 
screen shot 2018-01-22 at 10 20 26 am

The 99% percentile isn't that high. That said, I'd still be okay upping this to say 3 minutes. I don't think that we really lose anything by upping it.

app/jobs/v2/package_files_job.rb
```diff
@@ -1,13 +1,24 @@
 class V2::PackageFilesJob < ActiveJob::Base
   queue_as :default
 
+  TIME_TO_AUTO_UNLOCK = 90
```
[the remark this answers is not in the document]
mdbenjam: Do we have any metrics on how long this endpoint takes now? My guess is this is too short. This feels like a pretty heavy job that could take many minutes.

app/services/record_fetcher.rb
```diff
@@ -4,40 +4,21 @@ class RecordFetcher
...
-    content
+    s = Redis::Semaphore.new("record_#{record.id}".to_s,
+                             connection: Rails.application.secrets.redis_url_sidekiq)
+    s.lock(SECONDS_TO_AUTO_UNLOCK)
+    content_from_s3 || content_from_vbms
   rescue *EXCEPTIONS
-    record.update(status: :failed)
     nil
-  # Catch StandardError in case there is an error to avoid records being stuck in pending state
-  rescue StandardError => e
```
aroltsch: We don't need it here anymore because we are not setting statuses here.

app/services/record_fetcher.rb
```diff
@@ -4,40 +4,21 @@ class RecordFetcher
...
-    record.update(status: :failed)
-    raise e
+  ensure
+    s.unlock
   end
 
   private
 
-  def wait_while_pending
```
aroltsch: So happy to get rid of this method!

app/models/record.rb
```diff
@@ -8,9 +8,8 @@ class Record < ActiveRecord::Base
 
   enum status: {
     initialized: 0,
-    pending: 1,
```
aroltsch: We can get rid of `pending` state now.

aroltsch:
Tested manually using PUT statements to ensure jobs are waiting for mutex to become available.

Issue #525, opened 2017-07-06
eFolder Express | Demo jobs and non-demo jobs diverge


mdbenjam:
### AC:
Ensure Demo numbers work for v2 endpoints. 

For validation reference this document: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/endpoints.md 
Ensure the above endpoints work for all scenarios from here: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/lib/fakes/document_service.rb#L7

PR #835, opened 2018-01-22
Anya/525 demo file numbers

aroltsch:
connects #525

### DIFF
--- a/app/jobs/v2/download_manifest_job.rb
+++ b/app/jobs/v2/download_manifest_job.rb
@@ -2,9 +2,7 @@ class V2::DownloadManifestJob < ActiveJob::Base
   queue_as :default
 
   def perform(manifest_source)
-    documents = ManifestFetcher.new(manifest_source: manifest_source).process
-    return if documents.blank?
-    DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
+    ManifestFetcher.new(manifest_source: manifest_source).process
   # Start downloading files if it is not eX user
   # V2::SaveFilesInS3Job.perform_later(manifest_source)
   # Catch StandardError in case there is an error to avoid manifests being stuck in pending state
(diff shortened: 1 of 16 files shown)


app/jobs/v2/download_manifest_job.rb
```diff
@@ -2,9 +2,7 @@ class V2::DownloadManifestJob < ActiveJob::Base
   queue_as :default
 
   def perform(manifest_source)
-    documents = ManifestFetcher.new(manifest_source: manifest_source).process
-    return if documents.blank?
-    DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create
```
aroltsch: I moved DocumentCreator into ManifestFetcher - much cleaner this way.

mdbenjam: Cool why did we remove the return if documents.blank? Presumably because DocumentCreator.new(manifest_source: manifest_source, external_documents: documents).create returns nil if there are no documents?

app/services/external_api/vva_service.rb
```diff
@@ -13,9 +13,15 @@ def self.fetch_documents_for(download)
     documents
   end
 
-  # TODO: remove when switched to VBMS eFolder API
-  def self.v2_fetch_documents_for(download)
-    fetch_documents_for(download)
+  def self.v2_fetch_documents_for(source)
```
mdbenjam: Is this just to make things more explicit? We plan on removing the v1 method when we're done?

aroltsch: v2 will reference the source object instead of manifest so these two methods are different. And yes we will remove the v1 method.

app/jobs/v2/download_manifest_job.rb
aroltsch: I honestly don't understand why I added that check because DocumentCreator won't create any documents if there are empty array. Maybe that check was for `nil` situations.

PR #837, opened 2018-01-23
Rename manifest source - source column

aroltsch:
Rename `manifest_source.source` to `manifest_source.name` as it is a better name for the column.


db/migrate/20180123192853_rename_source_to_name_in_sources.rb
```diff
@@ -0,0 +1,5 @@
+class RenameSourceToNameInSources < ActiveRecord::Migration
+  def change
+    rename_column :manifest_sources, :source, :name
```
aroltsch: This migration is safe because the code is not running in production.

Issue #813, opened 2018-01-03
[BE] Return historical cases searched for the landing page


aroltsch:
###AC

1. New v2 endpoint is created that returns historical cases for the current user
2. It should have an expiration date (3 days is what we have currently)

Sjones352:
**PASSED**

Validate Postman
1. Check the status of the files download
`GET <url>/api/v2/manifests/:manifest_id/files_downloads`

2. It should have an expiration date (3 days is what we have currently
<img width="958" alt="screen shot 2018-02-02 at 12 02 44 pm" src="https://user-images.githubusercontent.com/23080951/35746843-4bb13f0c-0816-11e8-86e9-81da05619693.png">

Issue #804, opened 2018-01-02
[FE] Add functionality to React landing page


lowellrex:
Add functionality to the landing page as part of Efolder Express UI React build-out initiative.

This ticket completes (and is blocked by) the work in #803. All functionality that currently is possible on the landing page will be enabled by this ticket. This includes, at a minimum (with the caveat that I may be forgetting additional functionality):
* Searching for file numbers initiates a manifest start and routes to the correct page
* Previously searched file numbers are displayed and linked to properly

## Acceptance Criteria:
* All features of the current landing page are enabled in the React version
* Searching works
* Linking to previous searches works

Issue #803, opened 2018-01-02
[FE] Rewrite landing page in React


lowellrex:
Rewrite the landing page as part of Efolder Express UI React build-out initiative.

This ticket is limited to making the landing page appear visually indistinguishable from the current user interface. Later tickets will add functionality to the landing page (linking to download results pages, enabling search, etc.). The work in this ticket should be able to be completed without communicating to the efolder API.

## Acceptance Criteria:
* Landing page looks visually indistinguishable from the current version:
![image](https://user-images.githubusercontent.com/32683958/34500248-e4c2f9a8-efd7-11e7-9f1a-3a94cf8cbdf0.png)

PR #838, opened 2018-01-23
Lowell 803 welcome page react

lowellrex:
Connects #803.

More work still to be done as part of #804 (which is blocked by #813 as well), but this PR just gets all the html elements in the correct spot looking somewhat passable visually.


client/package.json
```diff
@@ -9,21 +9,30 @@
     "lint:fix": "yarn run lint --fix"
   },
   "dependencies": {
+    "@department-of-veterans-affairs/appeals-frontend-toolkit": "^2.1.0",
+    "babel-core": "^6.26.0",
```
nickheiner-usds: `babel-*` deps should really be on the `devDependencies` side, but it doesn't make a huge difference since this module will have no dependents.

client/src/components/AlertIcon.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export default class AlertIcon extends React.Component {
```
nickheiner-usds: Please always default to `PureComponent`, since it avoids wasted renders.

client/src/components/RecentDownloadRow.jsx
```diff
@@ -0,0 +1,36 @@
...
+    // TODO: Use this.props.download.expiration_date or equivalent in UserManifest when it is built out.
+    const expirationDate = '01/26';
+
+    // TODO: Switch the link text based on this.props.download.in_progress or equivalent.
+    const downloadInProgress = true;
+
+    // TODO: Add alert icon based on this.props.download.had_errors or equivalent.
+    const downloadEncounteredErrors = true;
+
+    return <tr id={`download-${this.props.download.id}`}>
```
nickheiner-usds: Why do we need this `id`?

client/src/components/RecentDownloadRow.jsx
```diff
@@ -0,0 +1,36 @@
...
+      <td>
+        {veteranFullName}
```
nickheiner-usds: You could also just do:

```jsx
{this.props.download.veteran_first_name} {this.props.download.veteran_last_name}
```

Whatever seems cleaner to you.

client/src/components/RecentDownloadRow.jsx
```diff
@@ -0,0 +1,36 @@
...
+        <span className="cf-subtext"> ({this.props.download.file_number}) - Expires on {expirationDate} </span>
+      </td>
+      <td className="ee-actions-cell">
+        <Link to={`/downloads/${this.props.download.id}`}>
+          { downloadEncounteredErrors ? <AlertIcon /> : '' }
```
nickheiner-usds: The idiomatic way to do this is:

```jsx
{ downloadEncounteredErrors && <AlertIcon /> }
```

client/src/containers/InitContainer.jsx
```diff
@@ -1,28 +1,47 @@
...
+import WelcomeContainer from './WelcomeContainer';
 
+class InitContainer extends React.Component {
   render() {
-    return <div>
-      <textarea defaultValue={this.props.text} onChange={this.onTextareaChange.bind(this)} />
-      <h1>{this.props.text.length}</h1>
-    </div>;
+    return <BrowserRouter>
+      <div>
```
nickheiner-usds: What is this `div` for?

client/src/containers/InitContainer.jsx
```diff
@@ -1,28 +1,47 @@
...
+            <NavigationBar
+              appName="eFolder Express"
+              logoProps={{
+                accentColor: '#F0835e',
+                overlapColor: '#F0835e'
+              }}
+              userDisplayName={this.props.userDisplayName}
+              dropdownUrls={this.props.dropdownUrls}
+              defaultUrl="/react">
+              <div className="cf-wide-app section--document-list">
```
nickheiner-usds: Do you really need `section--document-list`?

We must be sure that all class names and IDs are really needed. This is the easiest time to remove them if not; it's harder to do it in an error-free way once the code is older.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -0,0 +1,55 @@
...
+            </i></p>
+          </div>
+          <RecentDownloadsContainer />
+        </div>
+      </div>
+    </main>;
+  }
+}
+
+const mapStateToProps = (state) => {
```
nickheiner-usds: This can be simplified:

```js
const mapStateToProps = (state) => _.pick(state, 'authenticityToken');
```

Generally, please default to `() => expr` instead of `() => { return expr; }` when the function is just a single return value.

client/src/containers/RecentDownloadsContainer.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import RecentDownloadRow from '../components/RecentDownloadRow';
+
+class RecentDownloadsContainer extends React.Component {
+  render() {
+    if (this.props.recentDownloads.length === 0) {
```
nickheiner-usds:
```js
if (!this.props.recentDownloads.length) {
```

client/src/containers/WelcomeContainer.jsx
nickheiner-usds: And please apply this change throughout the PR. :smile:

client/src/containers/WelcomeContainer.jsx
```diff
@@ -0,0 +1,55 @@
...
+            <h1>Welcome to eFolder Express</h1>
+            <p>eFolder Express allows VA employees to bulk-download VBMS eFolders.
+              <br />Search for a Veteran ID number below to get started.
+            </p>
+          </div>
+          <div className="ee-search">
+            <form
+              className="usa-search usa-search-big cf-form"
+              id="new_download"
+              action="/downloads"
```
nickheiner-usds: This is not necessary. We're making an SPA now, so we're not going to use a `form action` which does a full page refresh. Instead, we'll make an XHR request to trigger the necessary server actions. That can be in a different PR. For now, let's just remove the unnecessary content.

client/src/containers/RecentDownloadsContainer.jsx
```diff
@@ -0,0 +1,35 @@
...
+      <h2 className="ee-recent-searches">History</h2>
+      <table className="usa-table-borderless" summary="List of recent downloads and links to download their contents">
+        <thead>
+          <tr className="usa-sr-only">
+            <th scope="col">File Number</th>
+            <th scope="col">Actions</th>
+          </tr>
+        </thead>
+        <tbody>
+          {this.props.recentDownloads.map((dl, i) => <RecentDownloadRow download={dl} key={i} />)}
```
nickheiner-usds: Please use longer variable names.

Using indexes for a key is bad, because it makes it hard for React to efficiently render the content. Are there any actual identifiers we can use?

nickheiner-usds:
The page does not work:

image

(Ideally, the page should work before you submit it for CR. :smile:)

client/src/components/AlertIcon.jsx
lowellrex: For future reference, more information: https://stackoverflow.com/questions/41340697/react-component-vs-react-purecomponent#43936258

nickheiner-usds:
I posted a message scolding you for breaking the page but it was my fault lol. Disregard. :smile:

client/src/components/RecentDownloadRow.jsx
lowellrex: Grepping through the codebase `git grep download-` reveals that the reason we have this `id` in the table row element is so we can inspect it in tests. Going to leave the `id` around so we can maintain those tests.

client/src/components/RecentDownloadRow.jsx
lowellrex: The v1 Download model [currently implements](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/b50f527fa8793edddb95d602d3fad7a80cf17aec/app/models/download.rb#L46) some (potentially incorrect?) logic when building `veteran_name`. I think it would make sense to replicate this logic in the new model on the backend (falling back to some sensible value other than `undef` if there is no last name) and just pass that forward to the frontend. Leaving this as a variable for now, to be replaced by the attribute when the backend implements that method.

client/src/containers/InitContainer.jsx
lowellrex: Good catch! It was left over from before `BrowserRouter` wrapped the child elements.

lowellrex: I spoke too soon. We need it because `BrowserRouter` barfs with the following message if we don't have it, "A <Router> may have only one child element".

nickheiner-usds:
Once my a-f-t (https://github.com/department-of-veterans-affairs/appeals-frontend-toolkit/pull/9) PR lands, here's a diff to use it.

```diff
diff --git a/client/src/containers/WelcomeContainer.jsx b/client/src/containers/WelcomeContainer.jsx
index a555474..0ea76ef 100644
--- a/client/src/containers/WelcomeContainer.jsx
+++ b/client/src/containers/WelcomeContainer.jsx
@@ -2,12 +2,13 @@ import React from 'react';
 import { connect } from 'react-redux';
 
 import RecentDownloadsContainer from './RecentDownloadsContainer';
+import AppSegment from '@department-of-veterans-affairs/appeals-frontend-toolkit/components/AppSegment';
 
 class WelcomeContainer extends React.PureComponent {
   render() {
     return <main className="usa-grid">
       <div className="cf-app">
-        <div className="cf-app-segment cf-app-segment--alt ee-new-download">
+        <AppSegment extraClassNames="ee-new-download" filledBackground>
           <div className="ee-heading">
             <h1>Welcome to eFolder Express</h1>
             <p>eFolder Express allows VA employees to bulk-download VBMS eFolders.
@@ -40,7 +41,7 @@ Note: eFolder Express now includes Virtual VA documents from the Legacy Content
             </i></p>
           </div>
           <RecentDownloadsContainer />
-        </div>
+        </AppSegment>
       </div>
     </main>;
   }

```

nickheiner-usds:
The a-f-t PR landed, so please pull in `AppSegment` and use it here.

nickheiner-usds:
There's a big margin at the bottom here that feels unnecessary. Let's look in person at how we can remove it:

image

PR #839, opened 2018-01-24
Return user's historical downloads

aroltsch:
connects #813


app/models/manifest.rb
```diff
@@ -28,7 +31,7 @@ def download_and_package_files!
   end
 
   def recently_downloaded_files?
-    finished? && fetched_files_at && fetched_files_at > 3.days.ago
+    finished? && fetched_files_at && fetched_files_at > 3.hours.ago
```
aroltsch: We don't really need this check. We only need this check to avoid starting the job when it just ran. For example, whenever job grabs a mutex after waiting for another job to finish.

app/models/serializers/v2/files_download_serializer.rb
```diff
@@ -1,23 +0,0 @@
-class Serializers::V2::FilesDownloadSerializer < ActiveModel::Serializer
```
aroltsch: We can get rid of this serializer and just re-use manifest serializer.

app/models/manifest.rb
```diff
@@ -15,20 +15,23 @@ class Manifest < ActiveRecord::Base
...
+  UI_HOURS_UNTIL_EXPIRY = 72
+  API_HOURS_UNTIL_EXPIRY = 3
+
   def start!
     vbms_source.start!
     vva_source.start!
   end
 
   def download_and_package_files!
-    return if pending? || recently_downloaded_files?
```
aroltsch: Remove this in order to allow to retry at any time.

app/models/manifest.rb
mdbenjam: Maybe use API_HOURS_UNTIL_EXPIRY here? Or maybe it's just a coincidence its the same time length.

lib/fakes/document_service.rb
```diff
@@ -56,7 +56,8 @@ def self.v2_fetch_documents_for(source)
 
     sleep_and_check_for_error(demo, source.name)
 
-    return [] if source.records.count == demo[:num_docs]
+    # clean up existing records
+    source.records.delete_all
```
mdbenjam: Is this because we return a random set of docs each time? And we can't match them up with what we returned last time?

app/models/manifest.rb
aroltsch: It is a coincidence. I will create a constant though.

lib/fakes/document_service.rb
aroltsch: Yes! If you use the same file number, the number of docs will double each time.

app/models/manifest.rb
aroltsch: Actually I think a constant would not make sense here.

lib/fakes/document_service.rb
mdbenjam: That's a great question! I'm not sure, we've assumed no but that might not be right. Paging @nicholasholtz

aroltsch: I will create a separate PR to handle the above case ^^.

Issue #836, opened 2018-01-23
[BE] Return time estimate for files download


aroltsch:
### AC:
As part of the files_downloads response, return time estimate how long it is left to download the files

Sjones352:
PASSED
Validate Postman

Start retrieving manifest
1. Returns  ```time_to_complete": "less than a minute```
2. Returns    ```       "seconds_left": 40```
<img width="975" alt="screen shot 2018-02-04 at 4 17 52 pm" src="https://user-images.githubusercontent.com/23080951/35782517-0a824834-09c7-11e8-883d-025c630e8dd8.png">

PR #840, opened 2018-01-25
Anya/836 download files estimate

aroltsch:
connects #836


app/models/record.rb
```diff
@@ -29,6 +29,8 @@ class Record < ActiveRecord::Base
   delegate :manifest, :service, to: :manifest_source
   delegate :file_number, to: :manifest
 
+  AVERAGE_DOWNLOAD_TIME = 2
```
mdbenjam: nit: could we add _IN_SECONDS to this const? Just makes it super clear.

app/models/serializers/v2/files_download_serializer.rb
```diff
@@ -5,6 +5,7 @@ class Serializers::V2::FilesDownloadSerializer < ActiveModel::Serializer
   attribute :fetched_files_at
   attribute :number_successful_documents
   attribute :number_failed_documents
+  attribute :time_to_complete
```
mdbenjam: Do we need to send seconds left as well? How will we draw the progress bar?

aroltsch: Ah, great point!

app/models/record.rb
aroltsch: good idea!

Issue #808, opened 2018-01-02
[FE] Add functionality for React manifest retrieval components


lowellrex:
This ticket completes (and is blocked by) the work in #807. All behaviour that currently occurs during the course of a manifest retrieval on the search results page will be enabled by this ticket. This includes, at a minimum (with the caveat that I may be forgetting additional functionality):
* Progress indicators appearing at correct time
* UI elements (buttons, alerts, status text) appear/disappear as expected
* Documents are moved into correct tabs
* Tabs are selected/unselected when appropriate
* "Download efolder" button works

## Acceptance Criteria:
* All manifest retrieval behaviour on the current search results page are enabled in the React version
* Progress indicators appearing at correct time
* UI elements (buttons, alerts, status text) appear/disappear as expected
* Documents are moved into correct tabs
* Tabs are selected/unselected when appropriate
* "Download efolder" button works

lowellrex:
Deployment of this feature is blocked by #810, though the response can be mocked in testing so that ticket should not block progress on this ticket entirely.

Issue #806, opened 2018-01-02
[FE] Add functionality to React search results page


lowellrex:
This ticket completes (and is blocked by) the work in #805. All functionality that currently is possible on the search results page will be enabled by this ticket. This includes, at a minimum (with the caveat that I may be forgetting additional functionality):
* Manifest for given file number is properly displayed
* Showing the pending results spinner
* Making the efolder retrieve button start a document download and route to next step
* Copy file number to clipboard

## Acceptance Criteria:
* All features of the current search results page are enabled in the React version
* Manifest for given file number is properly displayed
* Showing the pending results spinner
* Making the efolder retrieve button start a document download and route to next step
* Copy file number to clipboard

Issue #805, opened 2018-01-02
[FE] Rewrite search results page in React


lowellrex:
Rewrite the search results page as part of Efolder Express UI React build-out initiative.

This ticket is limited to making the search results page appear visually indistinguishable from the current user interface. Later tickets will add functionality to the search results page (pending results spinner, efolder retrieve button works, copy file number to clipboard, etc.). The work in this ticket should be able to be completed without communicating to the efolder API.

## Acceptance Criteria:
* Search results page looks visually indistinguishable from the current version:
![image](https://user-images.githubusercontent.com/32683958/34500820-4fc28802-efda-11e7-9d38-fa7cb89369e9.png)

PR #841, opened 2018-01-25
Add out of service page to react app

lowellrex:
In the course of working on #805 I realized we would need to convert the out of service page and the `/help` page to react and we had overlooked making tickets for this work. This PR encompasses the work to make the out of service notification work in the new react single page app.

This isn't the ideal arrangement for how I imagine an out of service notification would function in a strict single page app, but this PR draws the UI elements, works well enough for now, and retains flexibility for doing things the more SPA-way as we incorporate the API requests into the UI work that is happening first.

Ideally, every single request to `efolder.cf.ds.va.gov/*` would route to our single page app (with the possible exception of session login/logout), the app would then fire off the requests based on the path, and build the page based on API responses. In that model, we would display the out of service notification when prompted by an API response. As we work on #804, #806, and #808 we will see if it is feasible to go to that approach entirely.


nickheiner-usds:
Let's use `StatusMessage` from https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/pull/11 so we can see it working.

nickheiner-usds:
We should fix this peer dependency warning before merging this PR:

```
λ yarn
yarn install v1.2.1
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
warning "babel-loader@7.1.2" has unmet peer dependency "babel-core@6 || 7 || ^7.0.0-alpha || ^7.0.0-beta || ^7.0.0-rc".
[4/4] 📃  Building fresh packages...
✨  Done in 5.27s.
```

nickheiner-usds:
Here's a PR to use `StatusMessage`: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/847.

PR #842, opened 2018-01-25
Remove unneeded documents

aroltsch:
Delete records if there were perviously created in Caseflow DB but no longer exist in VBMS.

### DIFF
--- a/app/services/document_creator.rb
+++ b/app/services/document_creator.rb
@@ -34,10 +34,16 @@ class DocumentCreator
 
   def create
     ManifestSource.transaction do
+      valid_records = []
       external_documents.each do |document|
-        Record.create_from_external_document(manifest_source, document)
+        valid_records << Record.create_from_external_document(manifest_source, document)
       end
+      remove_deleted_records(valid_records)
     end
+  end
+
+  def remove_deleted_records(valid_records)
+    (manifest_source.records - valid_records).each(&:delete)
   end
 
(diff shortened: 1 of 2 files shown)


app/services/document_creator.rb
```diff
@@ -34,12 +34,19 @@ class DocumentCreator
...
-        Record.create_from_external_document(manifest_source, document)
+        valid_records << Record.create_from_external_document(manifest_source, document)
       end
+      # User can delete documents from VBMS so clean up these records if they were previously created
+      remove_deleted_records(valid_records)
     end
   end
 
+  def remove_deleted_records(valid_records)
+    (manifest_source.records - valid_records).each(&:delete)
```
mdbenjam: I learned yesterday that `:destroy` might be better than `:delete` since it runs on destroy hooks while delete doesn't. I feel like we should destroy instead. Though I'm not sure we have any hooks right now!

app/services/document_creator.rb
```diff
@@ -34,12 +34,19 @@ class DocumentCreator
 
   def create
     ManifestSource.transaction do
+      valid_records = []
```
mdbenjam: nit, not a big deal, but this could be cleaned up by changing the below each to a map :)

aroltsch: Now I remember why I didn't use map. We don't want to override `external_documents` variable.

PR #843, opened 2018-01-26
Move help page into react

lowellrex:
In the course of working on #805 I realized we would need to convert the help page to react and we had overlooked making tickets for this work. This PR encompasses the work to make the help page work in the new react single page app.

Ideally we wouldn't need to get the path to the different assets from the store and could instead get them from some other API response, but I think this is good enough for now. If we are able to remove the other properties we are passing in through `initial_react_data` in the work that happens in #804, #806, and #808 then we should investigate alternative approaches to getting the paths to those assets.

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@department-of-veterans-affairs/appeals-frontend-toolkit": "^2.1.0",
+    "@department-of-veterans-affairs/caseflow-frontend-toolkit": "^2.2.0",
     "es5-shim": "^4.5.10",
     "glamor": "^2.20.40",
     "history": "^4.7.2",
--- a/client/src/containers/HelpContainer.jsx
+++ b/client/src/containers/HelpContainer.jsx
@@ -1,0 +1,309 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+import AppSegment from '@department-of-veterans-affairs/appeals-frontend-toolkit/components/AppSegment';
+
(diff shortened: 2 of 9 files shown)


client/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@department-of-veterans-affairs/appeals-frontend-toolkit": "^2.1.0",
+    "@department-of-veterans-affairs/caseflow-frontend-toolkit": "^2.2.0",
```
lowellrex: #841 updates all other occurrences of `appeals-frontend-toolkit` with `caseflow-frontend-toolkit`, so I didn't bother doing that as part of this PR.

client/src/containers/HelpContainer.jsx
```diff
@@ -0,0 +1,309 @@
...
+import AppSegment from '@department-of-veterans-affairs/appeals-frontend-toolkit/components/AppSegment';
+
+class HelpContainer extends React.PureComponent {
+  render() {
+    return <main className="usa-grid">
+      <AppSegment extraClassNames="cf-help-content" filledBackground>
+
+        <h1 id="#top">Welcome to the eFolder Express Help Page!</h1>
+        <p>Here you will find <a href="#training-videos">Training Videos</a> and answers to the
+          <a href="#faq">Frequently Asked Questions (FAQs)</a> for eFolder Express, as well as links to the
```
mdbenjam: I'd like to use the shared link component vs anchor tags. Just for consistency. It should already be in the toolkit.

PR #844, opened 2018-01-26
Add unauthorized alert to react app

lowellrex:
Once the frontend toolkit [PR to add the StatusMessage](https://github.com/department-of-veterans-affairs/caseflow-frontend-toolkit/pull/11) is merged I will move this over to use that shared StatusMessage component instead of `<AppSegment>`.

Tested that this displays correctly by adding a route in `InitContainer` and navigating to that route. Since I think this will more usually be used as an alert (and not a page to which folks are redirected), I did not include that route in this PR. I imagine this alert will probably respond to some global state in `InitContainer` that we set off an action based on an API response.

### DIFF
--- a/client/src/containers/UnauthorizedContainer.jsx
+++ b/client/src/containers/UnauthorizedContainer.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
-import { Link } from 'react-router-dom';
 
+import Link from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/Link';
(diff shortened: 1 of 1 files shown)


client/src/containers/UnauthorizedContainer.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
```
mdbenjam: We should import the link from the toolkit as well. It's just a thin wrapper over this.

Issue #807, opened 2018-01-02
[FE] Rewrite manifest retrieval components in React


lowellrex:
Rewrite the manifest retrieval components of the search results page as part of Efolder Express UI React build-out initiative.

This ticket is limited to making the manifest retrieval components of the search results page appear visually indistinguishable from the current user interface. Later tickets will add functionality (progress indicators appearing at correct time, documents are moved into correct tabs, tabs are selected/unselected etc.). The work in this ticket should be able to be completed without communicating to the efolder API.

This ticket is blocked by #805.

## Acceptance Criteria:
* Manifest retrieval components of the search results page looks visually indistinguishable from the current version:

![image](https://user-images.githubusercontent.com/32683958/34501063-3d1abe6c-efdb-11e7-9a79-56d4a7387af7.png)

PR #850, opened 2018-01-31
Welcome page submit search button interactivity

lowellrex:
Connects #804.

First pass at adding functionality to the welcome page's submit search button.

Decided to tackle this ticket before #807 because the downloads page lacks functionality until the fetch manifest job has been kicked off. This PR contains the changes to kick off that job.


app/controllers/api/v1/application_controller.rb
```diff
@@ -35,7 +35,7 @@ def authenticate_with_token
   end
 
   def user_has_role
-    current_user && (current_user.can?("Reader") || current_user.can?("Hearing Prep") || current_user.admin?)
+    current_user && (current_user.can?("Reader") || current_user.can?("Hearing Prep") || current_user.can?("Download eFolder") || current_user.admin?)
```
lowellrex: Had to add the "Download eFolder" role to this method since the efolder express is now accessing the backend through the API so we hit this method instead of setting the user in `BaseController.current_user` as we were previously when using the rails templates.

client/src/index.jsx
```diff
@@ -1,16 +1,15 @@
...
 
 import reducer from './reducer';
 import InitContainer from './containers/InitContainer';
 
 module.exports = {
   init(props) {
-    const initState = {};
-    const store = createStore(reducer, { ...initState,
-      ...props });
+    const store = createStore(reducer, props, applyMiddleware(thunk));
```
lowellrex: Artifact from a different approach I tried (using the action/reducer pattern) and abandoned for this particular use case. We will use action/reducer for other actions so I've left this around because it will be useful for later pieces.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -24,17 +60,23 @@ export default class WelcomeContainer extends React.PureComponent {
...
-              <input type="search" name="file_number" id="file_number" />
-              <button type="submit" className="cf-submit">
+              <input
+                type="search"
+                name="file_number"
+                id="file_number"
+                onChange={this.updateSearchInput.bind(this)}
+                onKeyPress={this.handleKeyPress.bind(this)}
+              />
+              <button type="submit" className="cf-submit" onClick={this.submitSearchInput.bind(this)}>
```
lowellrex: Is there a better way to reference these function without using bind?

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
...
+
+    request.
+      post('/api/v2/manifests').
+      set(headers).
+      send().
+      use(nocache).
+      then(
+        (resp) => {
+          // TODO: We will need to be able to query for the status of the manifest fetch based on the
+          // manifest ID.
```
aroltsch: You can't use the header? get `'/api/v2/manifests'` accepts the header

client/src/containers/WelcomeContainer.jsx
mdbenjam: So if you use arrow notation it automatically binds. `submitSearch = () => {}`

nickheiner-usds: Yes. In fact, we don't want to use `bind` in `render,` because it will cause the component to re-render every time, because props will always be different, because `bind` creates a new function each time.

Instead, just define `submitSearchInput` as a `=>` function above.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
 });
 
 export default class WelcomeContainer extends React.PureComponent {
+  updateSearchInput(event) {
+    this.searchInputText = event.target.value;
```
mdbenjam: We should rarely set instance variables in react. I might be able to explain better in person--but the general idea is you want to keep these kinds of values in the state so that changes to them trigger render updates to the page. This page explains it more fully: https://reactjs.org/docs/forms.html

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
...
+    }
+  }
+
+  submitSearchInput() {
+    const headers = {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+      FILE_NUMBER: this.searchInputText,
+      'X-CSRF-Token': document.querySelector('meta[name="csrf-token"]').content,
+      'X-Requested-With': 'XMLHttpRequest'
```
nickheiner-usds: I don't think we do this for our other requests. I don't think it's necessary.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
...
+      this.submitSearchInput();
+    }
+  }
+
+  submitSearchInput() {
+    const headers = {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+      FILE_NUMBER: this.searchInputText,
+      'X-CSRF-Token': document.querySelector('meta[name="csrf-token"]').content,
```
nickheiner-usds: Do we really need to get the CSRF token out of a DOM element? Can we not pass it directly from Rails into the the `init` function?

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
...
+      'X-Requested-With': 'XMLHttpRequest'
+    };
+
+    request.
+      post('/api/v2/manifests').
+      set(headers).
+      send().
+      use(nocache).
+      then(
+        (resp) => {
```
nickheiner-usds: We should also handle the error state, although that can be a future PR.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -24,17 +60,23 @@ export default class WelcomeContainer extends React.PureComponent {
...
                 Search for a Veteran ID number below to get started.
               </label>
-              <input type="search" name="file_number" id="file_number" />
-              <button type="submit" className="cf-submit">
+              <input
+                type="search"
+                name="file_number"
+                id="file_number"
+                onChange={this.updateSearchInput.bind(this)}
+                onKeyPress={this.handleKeyPress.bind(this)}
```
nickheiner-usds: It could be simpler to keep the parent as a `form` and use `onSubmit`: https://reactjs.org/docs/forms.html. It's up to you.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
...
+      post('/api/v2/manifests').
+      set(headers).
+      send().
+      use(nocache).
+      then(
+        (resp) => {
+          // TODO: We will need to be able to query for the status of the manifest fetch based on the
+          // manifest ID.
+          // https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/endpoints.md
+          this.props.history.push(`/downloads/${resp.body.data.id}`);
```
nickheiner-usds: Is there a loading state we want to display before the response returns?

nickheiner-usds:
I left a few comments but none are blocking.

client/src/containers/WelcomeContainer.jsx
mdbenjam: One pattern we've used a lot in Reader, and that might be applicable here is to do requests in `componentDidMount`. In that model clicking submit would just move you to the next page in the SPA. Then when that page mounts, we make the request, display a loading spinner, and do something with the result. One reason this might be useful, is what happens if someone direct links to the next page. Should it trigger the request? I'm not sure what the answer is, but it seems worth considering.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -11,6 +13,40 @@ const searchBarNoteTextStyling = css({
...
+  submitSearchInput() {
+    const headers = {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+      FILE_NUMBER: this.searchInputText,
+      'X-CSRF-Token': document.querySelector('meta[name="csrf-token"]').content,
+      'X-Requested-With': 'XMLHttpRequest'
+    };
+
+    request.
```
mdbenjam: At some point, it might be worth factoring this out into a utility class, like we have in Caseflow just so we don't have to add the CSRF headers to each request. But for now it's fine.

client/src/containers/WelcomeContainer.jsx
mdbenjam: And if we update the state we probably should do it in Redux, but we can chat more about that.

client/src/containers/WelcomeContainer.jsx
lowellrex: Correct, I meant for the GET request once we land on the downloads page. In this instance we will have the FILE_NUMBER available to us (through the state store or POST response), but if, for example, we navigate directly to `/downloads/12/` we only know the manifest ID and not the FILE_NUMBER.

Perhaps we can get around this by using `GET /api/v2/manifests/:manifest_id/files_downloads` (or consolidating that request with `GET /api/v2/manifests/`!)... Will research more.

client/src/containers/WelcomeContainer.jsx
lowellrex: Removed "X-Requested-With" field.

client/src/containers/WelcomeContainer.jsx
nickheiner-usds: lol GH wasn't showing me Mark's comment when I made mine.

client/src/containers/WelcomeContainer.jsx
lowellrex: I'm not super clear on what the error state would be. Display a message to the user and do not redirect?

Currently (in the production html templates version and this react PR), if the request fails we just never redirect to the downloads page.

client/src/containers/WelcomeContainer.jsx
lowellrex: I'd be open to that strategy here, however I'm not sure how we'd get the URL of the next page since it contains the manifest ID which we do not know until we get a response to the request we send. Or would you envision rendering `/downloads` and then appending the manifest ID to the URL after we have received the response to that request?

client/src/containers/WelcomeContainer.jsx
lowellrex: I didn't want to put the `searchInputText` variable into the global state because it does not have meaning outside of the `WelcomeContainer` object. I guess I'm not clear on how the class' state is different from an instance variable, but I'll make the change to make this live in local state.

client/src/containers/WelcomeContainer.jsx
lowellrex: Currently we have a little spinner on the button element, so I think that will be included in a later PR.

mdbenjam: ah gotcha, that's a fair point. The manifest id is separate from the file number. Sigh. This makes sense then.

client/src/containers/WelcomeContainer.jsx
mdbenjam: Hmmm that's a really interesting point. Maybe we make `GET /api/v2/manifests/:id` an alias for the FILE_NUMBER version.

client/src/containers/WelcomeContainer.jsx
mdbenjam: We don't display anything? Like a status message or something? Seems like we should if we aren't right now.

lowellrex: I don't think so. I think the form submit action is effectively a redirect for a standard html POST form. So if the request succeeds we will redirect to whichever page the form tells us to, and if the request fails then we won't... I think.

lowellrex:
The latest change spits out the following warning, that I don't really understand why we're getting. Perhaps because "controlled components" usually keep track of their own state? Do we have examples of this in the main caseflow repo that I can take a look at?
> Warning: A component is changing an uncontrolled input of type search to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components

nickheiner-usds:
Here's the solution to that warning. (I'm going to assume you clicked through to the link in the error message and read it.)

There are two ways to do components in React:

```jsx
// Controlled
<input value={providedValue} />

// Uncontrolled
<input />
```

In `WelcomeContainer`, we have the following component that is generating this warning:

```jsx
<input value={this.props.searchInputText} />
```

The first time we render `WelcomeContainer`, `this.props.searchInputText` is `undefined`. Thus, what we render is: 

```jsx
// All these are equivalent
<input value={this.props.searchInputText} />
<input value={undefined} />
<input /> // uncontrolled component
```

Once we type in the input field, we fire an action to update `searchInputText`. Then, our input becomes:

```jsx
// All these are equivalent
<input value={this.props.searchInputText} />
<input value={'entered text'} /> // controlled component
```

The solution is to initialize `searchInputText` to `''`, instead of `undefined`. I'll leave how to do that as an exercise for the reader. :smile:

client/src/actions.jsx
```diff
@@ -0,0 +1,6 @@
+import { UPDATE_SEARCH_TEXT } from './actionTypes';
+
+export const updateSearchInputText = (text) => (dispatch) => {
```
nickheiner-usds: This can be simplified:

```js
export const updateSearchInputText = (text) => ({ 
  type: UPDATE_SEARCH_TEXT,
  payload: text 
});
```

client/src/containers/WelcomeContainer.jsx
mdbenjam: I think all of these can show up on this page based on what id you search for: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/views/downloads/new.html.erb#L2

We don't have to address it in this PR, but we should make sure that every request handles errors.

nickheiner-usds: If a HTML form POST fails, then the server will display or redirect to an error page. We should have equivalent functionality in this SPA world.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -47,3 +87,14 @@ Note: eFolder Express now includes Virtual VA documents from the Legacy Content
     </main>;
   }
 }
+
+const mapStateToProps = (state) => ({
+  csrfToken: state.csrfToken,
+  searchInputText: state.searchInputText
+});
+
+const mapDispatchToProps = (dispatch) => ({
```
nickheiner-usds: `...` is for merging, and here we're merging the result of the `bindActionCreators` call with nothing else. Just use that value directly:

```js
const mapDispatchToProps = (dispatch) => bindActionCreators({ updateSearchInputText }, dispatch);
```

client/src/index.jsx
mdbenjam: Doesn't have to be in this PR, but I think it would be useful to either use the BaseRedux component Nick made, or to at least add the line to enable redux tools.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -47,3 +87,14 @@ Note: eFolder Express now includes Virtual VA documents from the Legacy Content
     </main>;
   }
 }
+
+const mapStateToProps = (state) => ({
```
nickheiner-usds: This can also be:

```js
const mapStateToProps = (state) => _.pick(state, 'csrfToken', 'searchInputText');
```

Not sure how @mdbenjam feels about that.

mdbenjam: I'm fine either way. I tend to prefer not using pick, but I think it's whatever Lowell prefers.

lowellrex:
@nickheiner-usds Implemented the fix you suggested in [your comment](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/850#issuecomment-362379793) by setting a default value for `searchInputText` in [index.jsx](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/850/commits/6859a79a800270daa377e9735466b9a5e2aef713). Is that a decent place to put that default? I tried setting it in reducer.js to no avail but am happy to move that default state definition around to wherever is most idiomatic.

nickheiner-usds:
Yep, what you did is exactly what I would have done! :smile:

PR #855, opened 2018-02-02
When VBMS is down, do not send error to Sentry

aroltsch:
This fixes sentry error: https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/612/


app/services/exception_logger.rb
```diff
@@ -1,5 +1,8 @@
 class ExceptionLogger
-  DEPENDENCY_MAINTENANCE = [/Could not get JDBC Connection/, /Maintenance.+VBMS/, /upstream connect error or disconnect/].freeze
+  DEPENDENCY_MAINTENANCE = [/Could not get JDBC Connection/,
+                            /Maintenance.+VBMS/,
+                            /upstream connect error or disconnect/,
+                            /A system error occurred/].freeze
```
aroltsch: I can't use Sentry excluded_exceptions because it doesn't allow for regex matchers.

PR #856, opened 2018-02-03
Backend additions for search results

lowellrex:
Connects #805.

These changes add components on the backend that we need for the frontend to function properly.


app/controllers/api/v2/manifests_controller.rb
```diff
@@ -30,7 +31,7 @@ def recent_downloads
   end
 
   def file_number
-    request.headers["HTTP_FILE_NUMBER"]
+    @file_number ||= request.headers["HTTP_FILE_NUMBER"] || params[:id] && Manifest.find(params[:id]).file_number
```
lowellrex: Need to add this because a user can navigate directly to `http://efolder.cs.ds.va.gov/downloads/1989` (perhaps via a bookmark) which bypasses the search page where we would normally capture the veteran ID (aka file number). As such, we need a way to look up the file number by manifest ID (1989 in the example URL).

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -42,6 +43,10 @@ def validate_header
     invalid_file_number unless bgs_service.valid_file_number?(file_number)
   end
 
+  def set_file_number_header
+    response.headers["HTTP_FILE_NUMBER"] = file_number
+  end
```
lowellrex: Need this for [the same reason](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/856/files#r165823571) that we need to be able to look up the file number by manifest ID. If a user navigates directly to an individual manifest page, we want to display the veteran ID (file number) of that manifest on the page, so we need to return the veteran ID in the HTTP response.

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -29,13 +29,25 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
       {
         id: document.id,
         type_id: document.type_id,
+        type_description: document.type_description,
         received_at: document.received_at,
         version_id: document.version_id,
+        source: document.manifest_source.name,
```
lowellrex: `type_description` and `source` are fields that are displayed on the manifest list page so we need to return them to the frontend.

config/routes.rb
```diff
@@ -12,7 +12,7 @@
 
   post 'increment_vva_coachmarks_status', to: 'downloads#increment_vva_coachmarks_status'
 
-  match '/react(/:any)' => 'gui#react', via: [:get]
+  match '/react(/*path)' => 'gui#react', via: [:get]
```
lowellrex: This allows paths to have multiple segments at the end of the URL (e.g. `/react/downloads/1989`) instead of just a single segment in the previous version (`/react/downloads/` matched the previous route, but `/react/downloads/1989` did not match).

config/routes.rb
```diff
@@ -35,8 +35,8 @@
     namespace :v2 do
       namespace :manifests, only: [] do
         post "/", action: :start
-        get "/", action: :progress
         get :history
+        get "/(:id)", action: :progress
```
lowellrex: The root `/manifests/` route needs to follow the `/manifests/history` route now that we added the optional manifest ID parameter otherwise the route `/manifests/history` would be match the pattern `/manifests/:id` with "history" being interpreted as the manifest ID.

app/controllers/api/v2/manifests_controller.rb
mdbenjam: Why not put the file number in the ManifestSerializer? We only need it in the header on our requests to the server to keep it out of the url.

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -29,13 +29,25 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
...
         series_id: document.series_id,
         created_at: document.created_at,
         updated_at: document.updated_at
       }
     end
   end
+
+  attribute :manifest_fetch_complete do
+    status = true
+    object.sources.map { |s| status = false if %w[initialized pending].include? s.status }
```
mdbenjam: This should be an each, not a map if we're not going to use the results.

mdbenjam: Why not check this on the frontend? This seems like a redundant field.

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -29,13 +29,25 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
...
+    status
+  end
+
+  attribute :veteran_full_name do
```
mdbenjam: Why not construct this on the frontend?

app/models/serializers/v2/manifest_serializer.rb
aroltsch: We did talk about it with Lowell and I actually do agree with Mark.

config/routes.rb
aroltsch: Should we make GET manifests/:id to always use the ID of the manifest instead of the file number? If we left both options it might confuse people. Although it will require changes to the Reader code.

app/controllers/api/v2/manifests_controller.rb
lowellrex: You're thinking that we should add some explicit error handling for that case? Maybe something like the following?
```
raise NoFileNumberError unless @file_number || request.headers["HTTP_FILE_NUMBER"] || params[:id]
```

config/routes.rb
lowellrex: I'm fine with allowing both options (file number in the header or manifest id in URL), but I'd be willing to remove the file number in the header code from the efolder backend if you'd like.

mdbenjam: I was sorta thinking about removing the file_number path too. But I don't have strong feelings.

app/controllers/api/v2/manifests_controller.rb
aroltsch: 1. Hmm, I was trying to say that if Manifest.find(params[:id]) doesn't find anything, it will return `nil` so you will end up calling `nil.file_number`; calling a method on the nil object.
2. I would separate this method into two methods to make it clearer/cleaner.

config/routes.rb
aroltsch: To make it clearer, I would have only two functions:
1. POST /manifests with FILE_NUMBER in the header
2. GET /manifests/:id to check status

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -1,6 +1,8 @@
 # TODO: create Api::V2::ApplicationController
 class Api::V2::ManifestsController < Api::V1::ApplicationController
-  before_action :validate_header, except: :history
+  before_action :set_file_number_from_header, only: :start
+  before_action :set_file_number_from_param, only: :progress
```
aroltsch: I think you can still have the same method: set_file_number. You can use `action_name` to check whether it is `start` or `progress` to use the appropriate mechanism

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -29,19 +31,27 @@ def recent_downloads
     @recent_downloads ||= current_user.recent_downloads
   end
 
-  def file_number
-    @file_number ||= request.headers["HTTP_FILE_NUMBER"] || params[:id] && Manifest.find(params[:id]).file_number
+  attr_reader :file_number
+
+  def set_file_number_from_param
+    m = Manifest.find(params[:id])
+    return invalid_manifest_id unless params[:id] && m
```
aroltsch: I don't think you need to check for params[:id]. If params[:id] is not there, the url would be incorrect.

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -53,4 +63,8 @@ def bgs_service
   def invalid_file_number
     render json: { status: "File Number is invalid, must be 8 or 9 digits" }, status: 400
   end
+
+  def invalid_manifest_id
+    render json: { status: "Manifest ID does not match any existing manifests" }, status: 400
```
aroltsch: I am actually noticing we can reuse `record_not_found` method. We can out it in the application_controller and remove it from all v2 controllers.

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -29,19 +30,29 @@ def recent_downloads
...
-  def validate_access
-    forbidden("sensitive record") unless bgs_service.check_sensitivity(file_number)
+  def set_file_number
+    if action_name == "start"
+      return missing_header("File Number") unless request.headers["HTTP_FILE_NUMBER"]
+      @file_number = request.headers["HTTP_FILE_NUMBER"]
+    elsif action_name == "progress"
+      begin
+        @file_number = Manifest.find(params[:id]).file_number
+      rescue StandardError
```
aroltsch: You probably want to rescue RecordNotFound https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/1ec875d3efac0884f70d34d110f5d372b18c17bb/app/controllers/api/v2/records_controller.rb#L40

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -29,19 +30,29 @@ def recent_downloads
...
 
-  def file_number
-    request.headers["HTTP_FILE_NUMBER"]
-  end
+  attr_reader :file_number
 
-  def validate_access
-    forbidden("sensitive record") unless bgs_service.check_sensitivity(file_number)
+  def set_file_number
+    if action_name == "start"
```
aroltsch: I would use a case statement instead of if/else, it is much cleaner.

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -1,6 +1,7 @@
 # TODO: create Api::V2::ApplicationController
 class Api::V2::ManifestsController < Api::V1::ApplicationController
-  before_action :validate_header, except: :history
```
aroltsch: I would leave it as a before_action for only: :start. `set_file_number` should only worry about setting the file_number.

PR #857, opened 2018-02-03
Frontend work for search results page

lowellrex:
Connects #805.

This is the beginning of the frontend work on this ticket. More work still remains (enabling downloading the actual efolder zip), but I want to get a gut check on the approach in this first step. I expect the change here that enable searching from the welcome page and displaying the search results will have only minor changes--if any--after this is merged.

Blocked by the backend work contained in [PR 856](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/856).


client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,114 @@
...
+  // MANIFEST_FETCH_STATUS_DOWNLOADING,
+  // MANIFEST_FETCH_STATUS_COMPLETE,
+  MANIFEST_FETCH_STATUS_ERRORED
+} from '../Constants';
+import DownloadListContainer from './DownloadListContainer';
+import DownloadSpinnerContainer from './DownloadSpinnerContainer';
+
+// Reader polls every second for a maximum of 20 seconds. Match that here.
+const MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS = 1;
+const MAX_MANIFEST_FETCH_RETRIES = 20;
```
lowellrex: I'm not sure why we settled on polling at 1 second intervals for a maximum of 20 seconds in reader, but I've matched that here.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+  }
+
+  pollManifestFetchEndpoint(retryCnt = 0) {
+    const headers = {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+      'X-CSRF-Token': this.props.csrfToken
+    };
+
+    request.
```
lowellrex: When we make a request in a third place in the code I will examine more thoroughly whether we can factor the request logic out to something like the main caseflow repo's `ApiUtil`.

client/src/components/DocumentListRow.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export default class DocumentListRow extends React.PureComponent {
```
mdbenjam: This component seems like overkill to me. Why not just return inline this below, or call it out in a function? I feel like factoring out something small like this that isn't reused in other files, reduces readability.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+          if (resp.body.data.attributes.manifest_fetch_complete === false) {
+            if (retryCnt < MAX_MANIFEST_FETCH_RETRIES) {
+              const sleepTime = MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS * 1000;
+
+              setTimeout(() => {
+                this.pollManifestFetchEndpoint(retryCnt + 1);
+              }, sleepTime);
+            } else {
+              const sleepLengthSeconds = MAX_MANIFEST_FETCH_RETRIES * MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS;
+              const errMsg = `Manifest fetch timed out after ${sleepLengthSeconds} seconds`;
```
mdbenjam: Is this the message we currently show users? I'd opt for something a little more user friendly. Not sure people will understand what time out means.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    request.
+      get(`/api/v2/manifests/${this.props.match.params.manifestId}`).
+      set(headers).
+      send().
+      use(nocache).
+      then(
+        (resp) => {
+          if (resp.body.data.attributes.manifest_fetch_complete === false) {
```
mdbenjam: nit: don't need to check `=== false` can just use (!resp.body.data.attributes.manifest_fetch_complete)

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+              setTimeout(() => {
+                this.pollManifestFetchEndpoint(retryCnt + 1);
+              }, sleepTime);
+            } else {
+              const sleepLengthSeconds = MAX_MANIFEST_FETCH_RETRIES * MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS;
+              const errMsg = `Manifest fetch timed out after ${sleepLengthSeconds} seconds`;
+
+              this.props.setManifestFetchErrorMessage(errMsg);
+              this.props.setManifestFetchStatus(MANIFEST_FETCH_STATUS_ERRORED);
```
mdbenjam: I'm somewhat skeptical of the need for this state machine set by `setManifestFetchStatus`. Every time you set status, you also set another variable. Either an error, or response. Can we just use those to determine the status? In these setups I worry about us in some situations setting the status but no the data or vice versa.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+
+// TODO: Add modal for confirming that the user wants to download even when the zip does not contain the entire
+// list of all documents.
+class DownloadContainer extends React.PureComponent {
+  componentDidMount() {
+    this.props.clearManifestFetchState();
+    this.pollManifestFetchEndpoint(0);
+  }
+
+  pollManifestFetchEndpoint(retryCnt = 0) {
```
nickheiner-usds: It's our convention to fully spell out names, so this should be `retryCount`.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+    return 'VVA/LCM';
+  }
+
+  return source;
+};
+
+const formatDateString = function(str) {
+  const date = new Date(str);
+
+  // TODO: Current version of efolder displays months and dates with leading zeroes if < 10.
```
mdbenjam: If there's not a reason for that, I think it's fine not to display leading zeros.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+    request.
+      get(`/api/v2/manifests/${this.props.match.params.manifestId}`).
+      set(headers).
+      send().
+      use(nocache).
+      then(
+        (resp) => {
+          if (resp.body.data.attributes.manifest_fetch_complete === false) {
+            if (retryCnt < MAX_MANIFEST_FETCH_RETRIES) {
+              const sleepTime = MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS * 1000;
```
nickheiner-usds: Please include a unit of measure on variable names where appropriate. This should be `sleepTimeMs`.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+  return `${date.getMonth() + 1}/${date.getDate()}/${date.getFullYear()}`;
+};
+
+class DownloadListContainer extends React.PureComponent {
+  componentDidMount() {
+    const resp = this.props.manifestFetchResponse;
+
+    this.props.setVeteranName(resp.body.data.attributes.veteran_full_name);
```
mdbenjam: We shouldn't put computed state in the redux store. If we have response, we have all the needed values. It might be easier to put `resp.body.data.attributes` in the store instead of putting the whole `resp`. You could also flatten the response as you put it in the store. But now we're duplicating data.

client/src/containers/DownloadContainer.jsx
nickheiner-usds: Unless we're explicitly testing that the value is `false` and not `''`, `0`, `undefined`, or `null`, this can just be:

```js
if (!resp.body.data.attributes.manifest_fetch_complete) {
```

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,112 @@
...
+              this.props.setManifestFetchErrorMessage(errMsg);
+              this.props.setManifestFetchStatus(MANIFEST_FETCH_STATUS_ERRORED);
+            }
+          } else {
+            this.props.setManifestFetchResponse(resp);
+            this.props.setManifestFetchStatus(MANIFEST_FETCH_STATUS_LISTED);
+          }
+        },
+        (err) => {
+          const errMsg = `${err.response.statusCode} (${err.response.statusText}) ${err.response.body.status}`;
```
nickheiner-usds: Is this meant to be shown directly to end users?

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+import { bindActionCreators } from 'redux';
+
+import AppSegment from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/AppSegment';
+
+import { setVeteranId, setVeteranName } from '../actions';
+import DocumentListRow from '../components/DocumentListRow';
+import DownloadPageFooter from '../components/DownloadPageFooter';
+import DownloadPageHeader from '../components/DownloadPageHeader';
+
+const aliasForSource = function(source) {
```
nickheiner-usds: Unless you're using dynamic `this` binding, please default to the `=>` function syntax.

This can just be:

```js
const aliasForSource = (source) => source === 'VVA' ? 'VVA/LCM' : source;
```

client/src/containers/DownloadListContainer.jsx
nickheiner-usds: https://momentjs.com is a useful date formatting library in JS that you can use if the vanilla JS dates aren't powerful enough.

client/src/containers/DownloadListContainer.jsx
nickheiner-usds: This seems like a bit of an odd pattern. It looks like we're reading data out of one part of the Redux state and writing that same data into another? This seems like an unnecessary complication.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+    this.props.setVeteranName(resp.body.data.attributes.veteran_full_name);
+    this.props.setVeteranId(resp.headers.http_file_number);
+  }
+
+  render() {
+    const resp = this.props.manifestFetchResponse;
+
+    const docSources = resp.body.data.attributes.sources;
+    const totalDocumentsCount = docSources.reduce((cnt, src) => cnt + src.number_of_documents, 0);
```
nickheiner-usds: `reduce` is actually a fairly low-level function, and there's often a cleaner, higher-level alternative:

```js
const totalDocumentsCount = _.sumBy(docSources, 'number_of_documents');
```

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+    const documentCountNote = docSources.map((src) => (
+      `${src.number_of_documents} from ${aliasForSource(src.source)}`)).join(' and ');
+
+    return <main className="usa-grid">
```
nickheiner-usds: I see us writing `<main className="usa-grid">` a lot. Can it be factored out to a root level? Or is there a lot of variation on whether each component needs it?

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+        <p>
+          <input className="cf-submit cf-retrieve-button" type="submit" value={this.props.startDownloadButtonLabel} />
+        </p>
+
+        <div className="ee-document-list">
+          <table className="usa-table-borderless ee-documents-table" summary="Files in veteran's eFolder">
+            <thead>
+              <tr>
+                <th className ="document-col" scope="col">Document Type</th>
```
nickheiner-usds: It's a shame that our linter isn't catching this, but there should not be a space here:

```
className ="document-col"
```

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+                <th className ="upload-col" scope="col">Receipt Date</th>
+              </tr>
+            </thead>
+
+            <tbody className ="ee-document-scroll" >
+              { resp.body.data.attributes.records.map((record) => (
+                <DocumentListRow
+                  type={record.type_description}
+                  source={aliasForSource(record.source)}
+                  received_at={formatDateString(record.received_at)}
```
nickheiner-usds: Our convention is to use `camelCase` for identifiers in JS. (It's fine to have it be `record.received_at` since we got that value directly from the server.)

client/src/index.jsx
```diff
@@ -7,7 +7,14 @@ import thunk from 'redux-thunk';
 import reducer from './reducer';
 import InitContainer from './containers/InitContainer';
 
-const initState = { searchInputText: '' };
+// TODO: Should we set up all of our defaults in the GuiController?
+const initState = {
+  searchInputText: '',
+  // TODO: Does startDownloadButtonLabel need to be a variable?
```
nickheiner-usds: Good question! :smile: If we never change it, then it would be better living in a constants file instead of in the Redux state.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,97 @@
...
+          {this.props.veteranName} #{this.props.veteranId}. Verify the Veteran ID and click the&nbsp;
+          {this.props.startDownloadButtonLabel} button below to start retrieving the eFolder.
+        </p>
+
+        <p>
+          <input className="cf-submit cf-retrieve-button" type="submit" value={this.props.startDownloadButtonLabel} />
+        </p>
+
+        <div className="ee-document-list">
+          <table className="usa-table-borderless ee-documents-table" summary="Files in veteran's eFolder">
```
mdbenjam: Why not use either our table component or maybe one on NPM?

client/src/containers/WelcomeContainer.jsx
```diff
@@ -21,6 +21,8 @@ class WelcomeContainer extends React.PureComponent {
   }
 
   handleFormSubmit = (event) => {
+    this.props.setVeteranId(this.props.searchInputText);
```
mdbenjam: We already have this in the URL right? I don't think we also need it in the state.

client/src/actions.jsx
```diff
@@ -1,6 +1,45 @@
...
+  SET_MANIFEST_FETCH_ERROR_MESSAGE,
+  SET_MANIFEST_FETCH_RESPONSE,
+  SET_MANIFEST_FETCH_STATUS,
+  SET_VETERAN_ID,
+  SET_VETERAN_NAME,
+  UPDATE_SEARCH_TEXT
+} from './actionTypes';
+import { MANIFEST_FETCH_STATUS_LOADING } from './Constants';
+
+export const setManifestFetchErrorMessage = (msg) => ({
```
mdbenjam: It might be interesting to start adding analytics now, for things like success and failures. It'll be easier than adding them later.

client/src/containers/DownloadContainer.jsx
lowellrex: We don't currently have a timeout on the frontend for the document list page.

client/src/containers/DownloadContainer.jsx
lowellrex: Yes

client/src/containers/DownloadListContainer.jsx
lowellrex: I decided to leave it in the components because it is serving double duty as a wrapping `<div>`, but we could move it up to InitContainer and replace it with divs where it was being used before if you think that is best.

client/src/containers/DownloadListContainer.jsx
lowellrex: I just grabbed the table markup that efolder was using before and dropped it in here. I'd be open to using a shared table component from the frontend toolkit if we want to put one in there.

client/src/containers/DownloadListContainer.jsx
nickheiner-usds: I think your latter suggestion is a good one, but it's not that big a deal. It's up to you.

client/src/containers/DownloadContainer.jsx
nickheiner-usds: Hmm... this seems out of line with the rest of our user-facing messaging. Was this in a mockup, or is this what eFolder does currently?

client/src/containers/DownloadListContainer.jsx
nickheiner-usds: Actually, we can use [Fragments](https://reactjs.org/docs/fragments.html) instead of wrapping `div`s. Then we can refactor `main` to `InitContainer`.

client/src/containers/WelcomeContainer.jsx
lowellrex: Talked it out in person. The manifest ID is the parameter in the URL so we do in fact need this in global state.

client/src/containers/DownloadContainer.jsx
nickheiner-usds: Let's rope in a designer to see what they think... @Chingujo?

If it becomes a blocker for this PR, we can just merge as-is, then make a ticket to circle back and change the copy to something else.

client/src/components/DownloadPageHeader.jsx
```diff
@@ -0,0 +1,24 @@
...
+import AppSegment from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/AppSegment';
+
+import CopyIcon from '../components/CopyIcon';
+
+// TODO: Get copy-paste feature working.
+export default class DownloadPageHeader extends React.PureComponent {
+  render() {
+    return <AppSegment extraClassNames="cf-efolder-header">
+      <h3 className="cf-push-left cf-name-header cf-efolder-name">{this.props.veteranName}</h3>
+      <div className="cf-txt-uc cf-efolder-id-control cf-push-right ">Veteran ID &nbsp;
```
nickheiner-usds: nit: extra space in `className="... cf-push-right "`.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,120 @@
...
+  setVeteranName
+} from '../actions';
+import DownloadListContainer from './DownloadListContainer';
+import DownloadSpinnerContainer from './DownloadSpinnerContainer';
+
+// Reader polls every second for a maximum of 20 seconds. Match that here.
+const MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS = 1;
+const MAX_MANIFEST_FETCH_RETRIES = 20;
+
+const manifestFetchInProgress = (responseAttrs) => {
```
nickheiner-usds: I would write this as:

```js
const manifestFetchInProgress = ({sources}) => _(sources)
    .map('status')
    .some((status) => ['initialized', 'pending'].includes(status))
```

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,120 @@
...
+                this.pollManifestFetchEndpoint(retryCount + 1);
+              }, sleepTimeMs);
+            } else {
+              const sleepLengthSeconds = MAX_MANIFEST_FETCH_RETRIES * MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS;
+              const errMsg = `Failed to fetch list of documents within ${sleepLengthSeconds} second time limit`;
+
+              this.props.setErrorMessage(errMsg);
+            }
+          } else {
+            this.props.setDocuments(respAttrs.records);
```
nickheiner-usds: It looks like `setDocuments` and `setDocumentsSources` are only ever called together. In situations like that, my general preference is to have a single `setDataFromBackend` action that sets everything it needs to, because that feels simpler to me. But this is fine too. It's up to you.

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,120 @@
...
+
+          this.props.setErrorMessage(errMsg);
+        }
+      );
+  }
+
+  // TODO: Add display for in progress.
+  // TODO: Add display for download complete.
+  render() {
+    // Every manifest we fetch should have at least 2 sources so that is our indicator that manifest fetch is complete.
```
nickheiner-usds: This comment is odd with the next line. This comment says we must have at least 2 sources, but the next line is only checking that we don't have 0 sources.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,72 @@
...
+
+import AppSegment from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/AppSegment';
+
+import DownloadPageFooter from '../components/DownloadPageFooter';
+import DownloadPageHeader from '../components/DownloadPageHeader';
+import { START_DOWNLOAD_BUTTON_LABEL } from '../Constants';
+
+const aliasForSource = (source) => source === 'VVA' ? 'VVA/LCM' : source;
+
+const formatDateString = (str) => {
```
nickheiner-usds: If it felt better to you, you could make this its own component:

```jsx
<FormattedDate rawStr={str} />
```

nickheiner-usds:
When I type "DEMO!@#" as the veteran ID from the search page, I get the following messages:

image

The second seems undesirable.

client/src/containers/InitContainer.jsx
```diff
@@ -22,14 +23,17 @@ class InitContainer extends React.PureComponent {
           userDisplayName={this.props.userDisplayName}
           dropdownUrls={this.props.dropdownUrls}
           defaultUrl="/">
-          <Route exact path="/" component={WelcomeContainer} />
-          <Route exact path="/out-of-service" component={OutOfServiceContainer} />
-          <Route exact path="/help" component={HelpContainer} />
+          <main className="usa-grid">
```
nickheiner-usds: 👍 This feels better to me.

client/src/containers/UnauthorizedContainer.jsx
```diff
@@ -5,12 +5,10 @@ import StatusMessage from '@department-of-veterans-affairs/caseflow-frontend-too
...
-      <StatusMessage title="Drat!">
-        You aren't authorized to use eFolder Express yet.
-        <br />
-        <Link to="/help">Click here</Link> to find out how you can request access to eFolder Express.
-      </StatusMessage>
-    </main>;
+    return <StatusMessage title="Drat!">
+      You aren't authorized to use eFolder Express yet.
+      <br />
+      <Link to="/help">Click here</Link> to find out how you can request access to eFolder Express.
```
nickheiner-usds: I'm sure you took this from how it was, but "[click here]()" links are a huge usability fail pet-peeve of mine. Would you mind changing this to "[Visit the help page]()"? Not a blocker.

nickheiner-usds:
If I go directly to http://localhost:3000/react/downloads/55 without searching for anything, I get:

image

I know we were going to work with a designer on that error message, but we should at least have it not say `undefined`.

client/src/containers/DownloadListContainer.jsx
lowellrex: I'm wary of adding another dependency for a project where the purpose is to prepare the codebase for maintenance mode. I understand the arguments in favor of lodash (that it is widely used, well tested, well known, well documented), but I still prefer to use built-in functions when possible in order to reduce our dependency on third parties.

In this particular case, if the `reduce()` function is too obtuse then I'd be in favor of summing in a loop. Maybe something like the following?
```
let totalDocumentsCount = 0;
this.props.documentSources.forEach( (src) => totalDocumentsCount += src.number_of_documents );
```

client/src/containers/DownloadListContainer.jsx
```diff
@@ -0,0 +1,72 @@
...
+              <tr>
+                <th className="document-col" scope="col">Document Type</th>
+                <th className="sources-col" scope="col">Source</th>
+                <th className="upload-col" scope="col">Receipt Date</th>
+              </tr>
+            </thead>
+
+            <tbody className="ee-document-scroll" >
+              { this.props.documents.map((doc) => (
+                <tr>
```
nickheiner-usds: We need a `key` here. Otherwise, we get a warning.

image

Does `doc` have an `id` we can use?

client/src/components/DownloadPageFooter.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+export default class DownloadPageFooter extends React.PureComponent {
+  render() {
+    return <AppSegment>
+      <input
+        className="cf-submit cf-retrieve-button ee-right-button"
+        type="submit"
+        value={this.props.label}
+      />
+      <Link href="/" className="ee-button-align">Start over</Link>
```
nickheiner-usds: Hmmm... this link actually goes back to `/` and not `react/`, which is what I'd expect. I don't know if we care.

client/src/containers/DownloadContainer.jsx
lowellrex: I find that a little tough to read myself, though I really like passing the list of sources in instead of the response object's attributes. Perhaps something like the following?
```
const manifestFetchInProgress = (sources) => {
  sources.forEach( (src) => if (['initialized', 'pending'].includes(src.status)) return true );
  return false;
}
```

client/src/containers/DownloadContainer.jsx
lowellrex: I don't have a strong preference either way. I have a slight preference to the current implementation because we are explicit about which bits of state we are setting.

client/src/containers/DownloadContainer.jsx
lowellrex: Fair enough. I was just trying to indicate why we were using that piece of state as the indicator. Do you have a suggestion for an alternate comment?

client/src/containers/DownloadListContainer.jsx
lowellrex: Is there some benefit to using a component in place of a traditional function?

nickheiner-usds:
Here's my suggestion for some cleanup: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/860.

client/src/containers/DownloadContainer.jsx
nickheiner-usds: That won't work because returning from the function passed to `forEach` won't do anything. But perhaps:

```js
const manifestFetchInProgress = ({sources}) => 
    _.some(sources, ({status}) => ['initialized', 'pending'].includes(status))
```

client/src/containers/DownloadContainer.jsx
nickheiner-usds: Either of the following would be consistent:

```js
    // Every manifest we fetch should have at least 2 sources so that is our indicator that manifest fetch is complete.
    if (this.props.documentSources.length === 2) {
      return <DownloadListContainer />;
    }
```

```js
    // Every manifest we fetch should have at least 1 source so that is our indicator that manifest fetch is complete.
    if (this.props.documentSources.length) {
      return <DownloadListContainer />;
    }
```

client/src/containers/DownloadListContainer.jsx
nickheiner-usds: In this instance, it's readability / whatever feels best to us.

client/src/containers/DownloadListContainer.jsx
nickheiner-usds:
```js
const totalDocumentsCount = docSources.reduce((cnt, src) => {cnt + src.number_of_documents}, 0);
```

Ha, I actually like the `reduce` option better than `forEach`, but I appreciate the effort :smile:.

I think that adding Lodash would be beneficial, and I'll state why here. However, we don't need to go back and forth on it (except as needed just to understand each other's position). You and @mdbenjam can have a discussion and I'll go along with whatever the two of you decide.

Lodash's functionality falls well within the scope of the standard library in a "batteries-included" style language. Compared to those languages, JS' standard library is modest. Saying that we won't use Lodash feels like a Rubyist opting out of `Enumerable`. 

We agree that we want to build this with maintenance in mind. To me, maintenance is less expensive for simpler code, and that's what Lodash gives us. With Lodash, we have less code, and the code we have is more obviously correct. For instance, the version of the `reduce` option I posted above is broken, but I could see someone scanning over the line and not catching the error.

Adding third party dependencies does have downsides, of course. Here's what I see them as being for a maintenance-mode project:

#### Third party code changes over time
This is not a concern, because we use Yarn. Yarn locks down the package to a specific version, and verifies package integrity with a checksum. The npm registry does not allow widely-used packages to be unpublished. If our tests pass today, they will continue to pass tomorrow, no matter what new versions of Lodash are released.

#### Third party code is obscure and takes more effort to work with
Lodash is heavily used in caseflow, caseflow-frontend-toolkit, and our AWS Lambdas. Anyone who is able to maintain those projects will already have familiarity with Lodash. And Lodash is, as you note, widely used and well documented.

Please let me know if you have concerns that I didn't address. Otherwise, I'll leave it to you and @mdbenjam to decide.

client/src/containers/DownloadContainer.jsx
lowellrex: I've touched the function up a bit to make it a little more readable. Let me know what you think.

client/src/containers/DownloadContainer.jsx
lowellrex: Clarified the comment a little.

client/src/containers/DownloadListContainer.jsx
lowellrex: Cool cool. I think I'll stick with what we've got now.

client/src/components/DownloadPageFooter.jsx
lowellrex: Think this needs to be `<Link to="/"...` instead of `href`. Fixing now

client/src/components/DownloadPageFooter.jsx
```diff
@@ -11,7 +11,7 @@ export default class DownloadPageFooter extends React.PureComponent {
         type="submit"
         value={this.props.label}
       />
-      <Link href="/" className="ee-button-align">Start over</Link>
+      <Link to="/" className="ee-button-align">Start over</Link>
```
nickheiner-usds: :+1: Ah nice, that makes sense.

client/src/containers/UnauthorizedContainer.jsx
nickheiner-usds: Thanks!

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,132 @@
...
+} from '../actions';
+import DownloadListContainer from './DownloadListContainer';
+import DownloadSpinnerContainer from './DownloadSpinnerContainer';
+
+// Reader polls every second for a maximum of 20 seconds. Match that here.
+const MANIFEST_FETCH_SLEEP_TIMEOUT_SECONDS = 1;
+const MAX_MANIFEST_FETCH_RETRIES = 20;
+
+const manifestFetchInProgress = (sources) => {
+  for (const src of sources) {
```
mdbenjam: You could use the some function here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

client/src/containers/DownloadContainer.jsx
```diff
@@ -0,0 +1,132 @@
...
+    if (['initialized', 'pending'].includes(src.status)) {
+      return true;
+    }
+  }
+
+  return false;
+};
+
+const buildErrorMessageFromResponse = (resp) => {
```
mdbenjam: I like this function. I wished we did something similar in Caseflow.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -35,9 +37,9 @@ class WelcomeContainer extends React.PureComponent {
       use(nocache).
       then(
         (resp) => {
-          // TODO: We will need to be able to query for the status of the manifest fetch based on the
-          // manifest ID.
-          // https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/docs/v2/endpoints.md
+          const responseAttrs = resp.body.data.attributes;
+
+          this.props.setVeteranName(`${responseAttrs.veteran_first_name} ${responseAttrs.veteran_last_name}`);
```
mdbenjam: Curious, why do we set veteran name here, but not other fields? It might be useful to set everything here, so that in the case where we've already downloaded the file, we don't see the loading spinner at all!

nickheiner-usds:
@lowellrex did you see this PR? I'm happy to fix the merge errors if you're interested. https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/860

lowellrex:
@nickheiner-usds I saw that but am a little hesitant to do that refactor until the final two views on the downloads page are built out. That is next up on my list of things to do, so I will go ahead and revisit that PR to see if it makes sense to do that consolidation when those last views are written.

client/src/containers/WelcomeContainer.jsx
lowellrex: Great call! This is a POST request to a different endpoint but they both return the same thing, so we might as well go ahead and set those other fields here as well.

PR #868, opened 2018-02-09
Add download progress page

lowellrex:
Connects #806.

This pull request contains the majority of the code changes necessary to make the manifest download page function properly. Getting this into code review now to get feedback before doing the remaining touch-up work which will not change the general approach or overall organization of this code.


client/src/apiActions.jsx
```diff
@@ -0,0 +1,99 @@
...
+        }
+      },
+      (err) => dispatch(setErrorMessage(buildErrorMessageFromResponse(err.response)))
+    );
+};
+
+export const startDocumentDownload = (options = {}) => (dispatch) => {
+  postRequest(`/api/v2/manifests/${options.manifestId}/files_downloads`, options.csrfToken).
+    then(
+      (resp) => dispatch(setStateFromResponse(resp)),
```
lowellrex: I'm sure there is a way to kick off `pollManifestFetchEndpoint()` here (to avoid doing it in `DownloadProgressContainer.componentDidMount()` in addition to `DownloadContainer.componentDidMount()`), however I am having trouble coming up with a way because the options passed to `pollManifestFetchEndpoint()` are slightly different (DownloadProgressContainer has a different condition to stop polling and does not set an error when it is done retrying).

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  render() {
+    const documents = {
+      progress: this.props.documents.filter((doc) => doc.status === DOCUMENT_DOWNLOAD_IN_PROGRESS_STATUS),
+      success: this.props.documents.filter((doc) => doc.status === DOCUMENT_DOWNLOAD_SUCCESS_STATUS),
+      failed: this.props.documents.filter((doc) => doc.status === DOCUMENT_DOWNLOAD_FAILED_STATUS)
+    };
+
+    return <React.Fragment>
```
lowellrex: A lot of this component's methods could/should probably be factored out to their own components (button in particular, but also the banners). Thoughts?

client/src/Constants.js
```diff
@@ -1 +1,19 @@
-export const START_DOWNLOAD_BUTTON_LABEL = 'Start retrieving efolder';
+export const EFOLDER_RETENTION_TIME_HOURS = '72';
+
+export const ERRORS_TAB = 'ERRORS_TAB';
+export const IN_PROGRESS_TAB = 'IN_PROGRESS_TAB';
+export const SUCCESS_TAB = 'SUCCESS_TAB';
+
+export const DOCUMENT_DOWNLOAD_IN_PROGRESS_STATUS = 'initialized';
```
nickheiner-usds: In Intake, we have just a single `REQUEST_STATUS` object that we share for all requests:

```js
export const REQUEST_STATE = {
  NOT_STARTED: 'NOT_STARTED',
  IN_PROGRESS: 'IN_PROGRESS',
  SUCCEEDED: 'SUCCEEDED',
  FAILED: 'FAILED'
};
```

That approach worked well for us, and it may simplify things here. It's your call.

client/src/containers/DownloadContainer.jsx
```diff
@@ -28,109 +26,104 @@ const MAX_MANIFEST_FETCH_RETRIES = 20;
 // for VVA and VBMS). If either of those document sources have anything other than a finished state, the entire
 // manifest fetch is incomplete.
 const manifestFetchComplete = (sources) => {
-  if (!sources) {
+  if (!sources.length) {
```
nickheiner-usds: I would just remove this `if` entirely, because the `for` loop below won't run if `sources` is empty.

client/src/containers/DownloadContainer.jsx
```diff
@@ -28,109 +26,104 @@ const MAX_MANIFEST_FETCH_RETRIES = 20;
...
 class DownloadContainer extends React.PureComponent {
-  componentDidMount() {
-    if (!manifestFetchComplete(this.props.documentSources)) {
-      this.pollManifestFetchEndpoint(0);
+  async componentDidMount() {
+    // Clear all previous error messages. The only errors we care about will happen after this component has mounted.
+    this.props.setErrorMessage('');
+
+    // Do not attempt to fetch the manifest until we have set the global manifest ID.
+    const totalRetryCount = await this.setGlobalManifestId(0);
```
nickheiner-usds: NICE let's get that `async` `await` going !

client/src/apiActions.jsx
```diff
@@ -0,0 +1,99 @@
...
+        dispatch(setStateFromResponse(resp));
+
+        if (options.stopPollingFunction(resp, dispatch)) {
+          return true;
+        }
+
+        if (retryCount < options.maxRetryCount) {
+          setTimeout(() => {
+            dispatch(pollManifestFetchEndpoint(retryCount + 1, options));
+          }, options.retrySleepSeconds * 1000);
```
mdbenjam: It seems like it should be called `retrySleepMillieconds` or shouldn't be multiplied by 1000.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -1,37 +1,41 @@
...
           {this.props.veteranName} #{this.props.veteranId}. Verify the Veteran ID and click the&nbsp;
-          {START_DOWNLOAD_BUTTON_LABEL} button below to start retrieving the eFolder.
+          {startDownloadButtonLabel} button below to start retrieving the eFolder.
         </p>
 
         <p>
-          <input className="cf-submit cf-retrieve-button" type="submit" value={START_DOWNLOAD_BUTTON_LABEL} />
+          <input
+            className="cf-submit cf-retrieve-button"
+            type="submit"
```
mdbenjam: Does this have to be of type submit? I think if you remove this you won't need to prevent default above.

client/src/containers/DownloadListContainer.jsx
```diff
@@ -1,37 +1,41 @@
...
+            value={startDownloadButtonLabel}
+            onClick={this.startDownload}
```
nickheiner-usds: If the user activates the button with the keyboard (e.g. `enter` key), will the `onClick` handler fire?

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+
+class DownloadProgressContainer extends React.PureComponent {
+  componentDidMount() {
+    this.props.setActiveDownloadProgressTab(IN_PROGRESS_TAB);
+
+    const pollOptions = {
+      csrfToken: this.props.csrfToken,
+      hideErrorAfterRetryComplete: true,
+      manifestId: this.props.manifestId,
+      maxRetryCount: 2 * 60 / 5,
```
nickheiner-usds: It would be more readable to name these numbers so future people know the significance of them. For instance, is the value `60` there because it's `SECONDS_PER_MINUTE`, or is that a coincidence?

nickheiner-usds: And is it a coincidence that the `5` here is the same as the value for `retrySleepSeconds` below?

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+    this.props.pollManifestFetchEndpoint(0, pollOptions);
+  }
+
+  // TODO: Add some request failure handling in here.
+  wrapInDownloadZipForm(element) {
+    return <form action={`/api/v2/manifests/${this.props.manifestId}/zip`} method="GET">{element}</form>;
+  }
+
+  inProgressBanner(docs) {
+    const pctComplete = 100 * (this.props.documents.length - docs.progress.length) / this.props.documents.length;
```
nickheiner-usds: Nit: please spell out `percent`. Our convention is to fully spell words out instead of using abbreviations.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+          <p className="usa-alert-text">
+            You can close this page at any time and eFolder Express will continue retrieving files in the&nbsp;
+            background. View progress and download eFolders from the History on the&nbsp;
+            <Link to="/">eFolder Express home page.</Link>
+          </p>
+          <p>Note: eFolders remain in your History for { EFOLDER_RETENTION_TIME_HOURS } hours.</p>
+        </div>
+      </div>
+
+      <h1 style={{ marginTop: '2rem',
```
nickheiner-usds: We should use glamor instead of inline styles.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+        textAlign: 'center' }}>Retrieving Files ...</h1>
```
nickheiner-usds: Is it intentional to have a space before the `...`? I feel like that's inconsistent with our other copy.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+      maxRetryCount: 2 * 60 / 5,
+      retrySleepSeconds: 5,
+      stopPollingFunction
+    };
+
+    this.props.pollManifestFetchEndpoint(0, pollOptions);
+  }
+
+  // TODO: Add some request failure handling in here.
+  wrapInDownloadZipForm(element) {
```
mdbenjam: I'm curious if you could do this with a Link instead a form and button. Not sure, but something about this feels a little strange to me.

client/src/containers/DownloadProgressContainer.jsx
mdbenjam: Where did you get this wait time. I think this could take several minutes (maybe an hour) or so. I don't actually have numbers, but I know people start this page and leave it over the weekend.

client/src/apiActions.jsx
mdbenjam: Can't the `startDocumentDownload` and `startManifestFetch` functions just pass different options to `pollManifestFetchEndpoint`?

mdbenjam: Though I'm fine with the current approach.

client/src/containers/DownloadContainer.jsx
```diff
@@ -28,109 +26,84 @@ const MAX_MANIFEST_FETCH_RETRIES = 20;
...
 
 // TODO: Add modal for confirming that the user wants to download even when the zip does not contain the entire
 // list of all documents.
 class DownloadContainer extends React.PureComponent {
   componentDidMount() {
+    // Clear all previous error messages. The only errors we care about will happen after this component has mounted.
+    this.props.setErrorMessage('');
+
+    const manifestId = this.props.match.params.manifestId;
+    this.props.setManifestId(manifestId);
```
mdbenjam: Do we need to set this in state? Can we always get it from the URL?

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+        <span style={{ width: `${pctComplete}%` }}>Progress: {pctComplete}%</span>
+      </div>
+    </React.Fragment>;
+  }
+
+  // TODO: Add caution alert to the download anyway button.
+  // TODO: Add action that will kick off the post request again for the "Try retrieving efolder again" button.
+  completeBanner(docs) {
+    if (docs.failed.length) {
+      return <div className="usa-alert usa-alert-error" role="alert">
```
nickheiner-usds: It would be nice to extract a component to share this alert between the two places where it's used in this function.

client/src/containers/DownloadProgressContainer.jsx
mdbenjam: Yeah I think moving the banners to their own components might be good. Seems a little cleaner to me.

client/src/containers/DownloadProgressTab.jsx
```diff
@@ -0,0 +1,36 @@
...
+      this.props.setActiveDownloadProgressTab(this.props.name);
+    }
+    event.preventDefault();
+  }
+
+  render() {
+    let classNames = 'cf-tab';
+
+    if (this.props.name === this.props.activeDownloadProgressTab) {
+      classNames = `${classNames} cf-active`;
```
nickheiner-usds: Please use `classNames` on npm to construct dynamic classnames instead of manual string formatting.

client/src/containers/DownloadProgressTab.jsx
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    return <button className={classNames} onClick={this.setActiveTab}>
+      <span>
```
nickheiner-usds: Why do we need to two `span`s here?

client/src/index.jsx
```diff
@@ -18,11 +21,9 @@ const initState = {
 
 module.exports = {
   init(props) {
-    const store = createStore(reducer, { ...initState,
-      ...props }, applyMiddleware(thunk));
-
-    render(<Provider store={store}>
+    render(<ReduxBase reducer={reducer} initialState={{ ...initState,
```
nickheiner-usds: 👍 Thanks!

client/src/containers/DownloadListContainer.jsx
lowellrex: We use the submit type to set a bunch of styles. We could inline all of those styles and remove the type element, but there appear to be nearly 20 styles set because `[type="submit"]` including some `-webkit-...` styles. I think it would be a non-trivial amount, but may be worth doing when/if we factor buttons out to their own component.

client/src/containers/DownloadListContainer.jsx
lowellrex: Yes

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,211 @@
...
+  DOCUMENT_DOWNLOAD_SUCCESS_STATUS,
+  DOCUMENT_DOWNLOAD_FAILED_STATUS
+} from '../Constants';
+import DownloadProgressTab from './DownloadProgressTab';
+import DownloadProgressTable from '../components/DownloadProgressTable';
+import { aliasForSource } from '../Utils';
+
+// TODO: We don't currently account for the case where there are no documents. This is probably fine.
+const documentDownloadComplete = (docs) => {
+  for (const doc of docs) {
```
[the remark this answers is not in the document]
lowellrex: Refactored this to use the Manifest's `fetched_files_status` field in place of this loop.

client/src/containers/DownloadProgressContainer.jsx
nickheiner-usds: image

Agreed that it feels odd to have a `form` with an `action` and `method` in an SPA, since those things are about doing a full-page refresh. But it may be fine in this case if all it does is open a new tab that triggers the download.

client/src/containers/DownloadProgressContainer.jsx
lowellrex: Makes sense. Would something like the following work?
```
const maxRetryMinutes = 2;
const retrySleepSeconds = 5;
const retryOptions = {
  ...options,
  maxRetryCount: maxRetryMinutes * 60 / retrySleepSeconds,
  retrySleepMilliseconds: retrySleepSeconds * 1000
};
```

client/src/containers/DownloadProgressContainer.jsx
lowellrex: I just copied what we currently have in efolder express. I can remove the space before the ellipsis if you think it best.

client/src/containers/DownloadProgressContainer.jsx
lowellrex: Currently we check the status indefinitely every 2 seconds (https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/84ecbc1e0c4aeef033bfcc25e221c6fd89afe66c/app/assets/javascripts/application/download_progress.js#L62). I wrote this code before I looked up those exact parameters, but I think it is worth considering whether we want to poll indefinitely. I figure we don't want to hit our server with unnecessary traffic, but this is probably minuscule so it may not make a difference.

With the code as written we will poll every 5 seconds and stop polling after two minutes. After those two minutes are up we will stop refreshing the status, but the document download will continue in the background.

lowellrex: After discussion with @mdbenjam we decided to poll every 2 seconds for a day.

client/src/containers/DownloadProgressTab.jsx
lowellrex: We didn't! Good catch!

client/src/containers/DownloadListContainer.jsx
mdbenjam: Testing it out, I think we can make it `type=button` and get all the right styles.

client/src/apiActions.jsx
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  return false;
+};
+
+const pollDocumentDownload = (retryCount = 0, resp, options = {}) => (dispatch) => {
+  const retryOptions = {
+    ...options,
+    // Poll every 2 seconds for 1 day
+    maxRetryCount: 1 * 24 * 60 * 60 / 2,
```
mdbenjam: Thinking more about this, I feel like we should up the time more than every 2 seconds. I like every 10 seconds would be reasonable. For a process that we expect to take a while, an extra couple seconds longer doesn't seem like a big deal. Not sure what the shortest time this might take would be though.

client/src/apiActions.jsx
```diff
@@ -0,0 +1,144 @@
...
+      (err) => dispatch(setErrorMessage(buildErrorMessageFromResponse(err.response)))
+    );
+};
+
+export const startDocumentDownload = (options = {}) => (dispatch) => {
+  postRequest(`/api/v2/manifests/${options.manifestId}/files_downloads`, options.csrfToken).
+    then(
+      (resp) => {
+        dispatch(setStateFromResponse(resp));
+        dispatch(pollManifestFetchEndpoint(0, options));
```
mdbenjam: Something seems weird about this to me. Why are we calling dispatch on functions that don't actually return an action? It seems we should just pass dispatch along?

client/src/apiActions.jsx
```diff
@@ -0,0 +1,144 @@
...
+    dispatch(setErrorMessage(errMsg));
+  }
+};
+
+export const pollManifestFetchEndpoint = (retryCount = 0, options = {}) => (dispatch) => {
+  getRequest(`/api/v2/manifests/${options.manifestId}`, options.csrfToken).
+    then(
+      (resp) => {
+        dispatch(setStateFromResponse(resp));
+        if (documentDownloadStarted(resp.body.data.attributes.fetched_files_status)) {
```
mdbenjam: This seems brittle to me. What if two users are referencing this case at the same time. One is downloading the file, the other is viewing the manifest. It seems we should pass in which of these we want to run in our options.

client/src/apiActions.jsx
```diff
@@ -0,0 +1,144 @@
...
+    retrySleepMilliseconds: 2 * 1000
+  };
+
+  if (!documentDownloadComplete(resp.body.data.attributes.fetched_files_status)) {
+    dispatch(retryPollManifestFetchEndpoint(retryCount, retryOptions));
+  }
+};
+
+const pollUntilFetchComplete = (retryCount = 0, resp, options = {}) => (dispatch) => {
+  if (manifestFetchComplete(resp.body.data.attributes.sources)) {
```
mdbenjam: This method is basically the same as the one above, but it has a different structure. This one starts with an if statement and then returns immediately the other doesn't. I think we should run them as parallel as possible.

mdbenjam: Could they be the same method? Could we pass in the options somewhere else?

client/src/apiActions.jsx
lowellrex: I'm not sure I totally understand the scenario you're describing. Are you saying that if Amy and Barb are both looking at Chuck's manifest on the manifest view page (or search results page), and then Amy starts a document download (and is sent to the download progress page), Barb will not also be redirected? Or is the scenario you're describing something different?

## Manifest view page (or search results page)
image

## Download progress page
image

client/src/apiActions.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+          dispatch(setErrorMessage(errMsg));
+        };
+
+        if (documentDownloadStarted(response.body.data.attributes.fetched_files_status)) {
+          // Poll every 10 seconds for 1 day
+          maxRetryCount = 1 * 24 * 60 * 60 / 10;
+          retrySleepMilliseconds = 10 * 1000;
+          donePollingFunction = (resp) => documentDownloadComplete(resp.body.data.attributes.fetched_files_status);
+          onRetriesExhaustedFunction = () => {}; // eslint-disable-line no-empty-function
```
mdbenjam: Don't we still want to set some error message here?

client/src/apiActions.jsx
```diff
@@ -0,0 +1,125 @@
...
+      (err) => dispatch(setErrorMessage(buildErrorMessageFromResponse(err.response)))
+    );
+};
+
+export const startDocumentDownload = (manifestId, csrfToken) => (dispatch) => {
+  postRequest(`/api/v2/manifests/${manifestId}/files_downloads`, csrfToken).
+    then(
+      (resp) => {
+        setStateFromResponse(dispatch, resp);
+        dispatch(pollManifestFetchEndpoint(0, manifestId, csrfToken));
```
mdbenjam: I'd still probably pass this dispatch as a param, instead of calling the function with dispatch. I don't even understand how this works as it is (though clearly it does). I think it's confusing when we try to dispatch things that don't immediately change state.

client/src/components/DownloadProgressTable.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
```
mdbenjam: Is the only difference between this table and the table on the previous page, the icons?

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -0,0 +1,168 @@
...
+      return this.wrapInDownloadLink(btn);
+    }
+
+    const btn = <button className="usa-button ee-right-button ee-download-button">Download efolder</button>;
+
+    return this.wrapInDownloadLink(btn);
+  }
+
+  render() {
+    const documents = {
```
mdbenjam: Not sure if this is cleaner. You could move this into your mapStateToProps so that you can access the document types as props.

client/src/apiActions.jsx
```diff
@@ -0,0 +1,125 @@
...
+      },
+      (err) => dispatch(setErrorMessage(buildErrorMessageFromResponse(err.response)))
+    );
+};
+
+export const startManifestFetch = (options = {}) => (dispatch) => {
```
mdbenjam: Does it make sense to get rid of the options object here as well?

client/src/apiActions.jsx
lowellrex: Chatted in person. Recap: we are somewhat constrained because we call `pollManifestFetchEndpoint` through `bindActionCreators` in DownloadContainer, so it is easier to allow that action to be dispatched itself instead of passing the dispatch function to that pollManifestFetchEndpoint as an argument.

PR #869, opened 2018-02-12
Enable webpacker watching in local dev

lowellrex:
Connects #812.

From root efolder project directory you can now run `foreman start` to start the rails server, build the initial webpack file, fetch the latest webpacked js file on reload, and have webpack rebuild the js file whenever the source files change.


Procfile
```diff
@@ -0,0 +1,3 @@
+js_server: ruby -run -e httpd app/assets/javascripts/react-app.js -p 3500
+rails: JAVASCRIPT_URL=http://localhost:3500 rails s
+client: HOT_RAILS_PORT=3500 cd client && yarn run dev:hot
```
mdbenjam: Is this actually hot reloading? Or just watching? We shouldn't conflate the two.

PR #870, opened 2018-02-12
Enable automatic routing for react app testing

lowellrex:
Connects #812.

Automatically route efolder express users to the react single page app if they have the `efolder_react_app` feature toggle enabled.


lowellrex:
Before this is merged, I will test in UAT to make sure the feature toggle is working (and importantly, isn't working when it isn't supposed to be).

config/routes.rb
```diff
@@ -53,6 +51,7 @@
     get code, :to => "errors#show", :status_code => code
   end
 
+  match '(*path)' => 'application#serve_single_page_app', via: [:get]
```
mdbenjam: Can't people get to the react app now if the page they enter doesn't exist?

app/controllers/application_controller.rb
```diff
@@ -33,8 +38,58 @@ def authorize
...
+  def initial_react_data
+    {
+      csrfToken: form_authenticity_token,
+      dropdownUrls: dropdown_urls,
+      efolderAccessImagePath: ActionController::Base.helpers.image_path("help/efolder-access.png"),
+      feedbackUrl: feedback_url,
+      recentDownloads: recent_downloads.sort_by(&:created_at).reverse,
+      referenceGuidePath: ActionController::Base.helpers.asset_path("reference_guide.pdf"),
+      trainingGuidePath: ActionController::Base.helpers.asset_path("training_guide.pdf"),
+      userDisplayName: current_user.display_name
```
mdbenjam: current_user may be nil, in which case this breaks.

config/routes.rb
lowellrex: Redirected to the v1 unauthorized page in this case and added logic to display the unauthorized banner in the single page app. I think the unauthorized message is a little imprecise and we should really be displaying a 404 or equivalent. I'll whip up a ticket to do just that if you think it worthwhile.

client/src/containers/InitContainer.jsx
```diff
@@ -25,9 +26,10 @@ class InitContainer extends React.PureComponent {
           defaultUrl="/">
           <main className="usa-grid">
             <Route exact path="/" component={WelcomeContainer} />
-            <Route exact path="/out-of-service" component={OutOfServiceContainer} />
-            <Route exact path="/help" component={HelpContainer} />
             <Route exact path="/downloads/:manifestId" component={DownloadContainer} />
+            <Route exact path="/help" component={HelpContainer} />
+            <Route exact path="/out-of-service" component={OutOfServiceContainer} />
+            <Route path="*" component={UnauthorizedContainer} />
```
mdbenjam: Interesting, I'd be curious to chat about this offline, seems like we should say page doesn't exist if they have access to the frontend router, but try to access a page that isn't there.

app/controllers/application_controller.rb
```diff
@@ -8,6 +8,7 @@ class ApplicationController < BaseController
   before_action :check_v2_app_access
 
   def serve_single_page_app
+    redirect_to("/unauthorized") && return unless can_access_react_app?
```
mdbenjam: This is good, though I feel like it is redundant with check_v2_app_access, we're doing the same check twice. Can we consolidate?

client/src/containers/InitContainer.jsx
lowellrex: I like that solution as well

lowellrex: Added a `NotFoundMessage` component that looks like this:
image

@Chingujo do you have any thoughts on how we can improve the messaging on the above banner that will display when a user navigates to `http://efolder.cf.ds.va.gov/bad_pathname` or some other page that does not exist (and would otherwise return a 404).

app/controllers/application_controller.rb
lowellrex: Moved this logic to the front-end using the authorization technique suggested by the react-router docs: https://reacttraining.com/react-router/web/example/auth-workflow

app/controllers/application_controller.rb
```diff
@@ -30,11 +35,66 @@ def authorize_system_admin
...
+        title: "Sign out",
+        link: url_for(controller: "/sessions", action: "destroy")
+      }
+    ]
   end
 
   private
 
+  def user_is_authorized?
+    current_user.can?("Download efolder") || Rails.env.development?
```
mdbenjam: When logging in current_user can be nil.

app/controllers/application_controller.rb
```diff
@@ -30,11 +35,66 @@ def authorize_system_admin
...
+        link: url_for(controller: "/help", action: "show")
+      },
+      {
+        title: "Send Feedback",
+        link: feedback_url,
+        target: "_blank"
+      },
+      {
+        title: "Sign out",
+        link: url_for(controller: "/sessions", action: "destroy")
```
mdbenjam: Clicking the sign out button redirects to a `cannot find` page now.

Issue #876, opened 2018-02-14
[FE] Display message for document list without documents


lowellrex:
In re-writing the efolder express UI in react, we left a few thing behind. One of those things was the text that displays in a list of documents if there are no documents in this view (for reference, how we [currently implement this](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/9b107e3af836c27016f2376842bc6f8af1e9ad74/app/views/downloads/_progress.html.erb#L101) in html templates). This is a rare case, but we still need to cover it.

## Acceptance criteria:
* Display the "There are no documents here." message in a document list if there are no documents

PR #877, opened 2018-02-14
Display message for document list without documents

lowellrex:
Connects #876.

Covering another edge case here. We want to display a message if we expect a list of documents and there are none.

### DIFF
--- a/client/src/components/ManifestDocumentsTable.jsx
+++ b/client/src/components/ManifestDocumentsTable.jsx
@@ -7,12 +7,18 @@ const documentTypeStyle = css({ width: '100%' });
 
 export default class DownloadProgressTable extends React.PureComponent {
   render() {
+    if (!this.props.documents.length) {
+      return <div className="ee-document-list">
+        <p className="cf-txt-c">There are no documents here.</p>
+      </div>;
+    }
+
     return <div className="ee-document-list">
       <table className="usa-table-borderless ee-documents-table" summary={this.props.summary}>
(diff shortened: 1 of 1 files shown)


client/src/components/ManifestDocumentsTable.jsx
```diff
@@ -7,12 +7,18 @@ const documentTypeStyle = css({ width: '100%' });
 
 export default class DownloadProgressTable extends React.PureComponent {
   render() {
+    if (!this.props.documents.length) {
+      return <div className="ee-document-list">
+        <p className="cf-txt-c">There are no documents here.</p>
+      </div>;
```
lowellrex: We duplicate a little code here by not having this condition in the return block below, but I find it a little more readable. Definitely open to moving it into the other return block if that is desired.

lowellrex:
Agreed that it looks odd, but that style matches what we currently have:
image

Can get it to look a little better to my eye by adding the following styles to `.cf-txt-c`:
```
border-bottom: 1px solid #d6d7d9;
padding-bottom: 2.4rem;
```
image

I'm happy to add those style changes via glamor, if you think that makes sense.

lowellrex:
Used glamor to add the above styles and the style that was being applied from the current class.

lowellrex:
Screencap of what it looks like (on this branch) on the search results page
image

lowellrex:
Ahhhh, are you suggesting no border at all on any of the pages?

PR #883, opened 2018-02-14
Removing documents from eFolder when no longer in VBMS

mdbenjam:
A number of documents were uploaded to a veteran's file that were intended for another veteran. When those files were removed from VBMS, we never remove them in eFolder. Therefore you can still view them in Reader. This isn't a problem in the new eFolder API, but we're back-porting the change to v1.

**Test Plan**
- [ ] Run automated tests to ensure documents that are not returned from VBMS are not returned by eFolder.

### DIFF
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -469,7 +469,6 @@ RSpec.feature "Downloads" do
     end
 
     download_documents = DownloadDocuments.new(download: @download, vbms_service: FakeVBMSService)
-    download_documents.create_documents
     download_documents.download_and_package
 
     visit download_path(@download)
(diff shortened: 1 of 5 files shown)


spec/features/download_spec.rb
```diff
@@ -469,7 +469,6 @@ def self.fetch_document_file(_document)
     end
 
     download_documents = DownloadDocuments.new(download: @download, vbms_service: FakeVBMSService)
-    download_documents.create_documents
```
mdbenjam: This shouldn't be called here. There are no external documents passed to the DownloadDocuments object.

spec/services/download_documents_spec.rb
```diff
@@ -169,7 +184,7 @@ def generate_id
 
         download.documents.each_with_index do |document, _i|
           document.save!
-          expect(document).to be_success
+          expect(document.reload).to be_success
```
mdbenjam: Unclear why the reload is now necessary, but breaks without it.

Issue #882, opened 2018-02-14
[FE] Display footer on error and loading pages


lowellrex:
We do not show the footer that has a link to the landing page when the search results page is loading or when any download page displays an error. We should.

### Search results page with footer:
![image](https://user-images.githubusercontent.com/32683958/36228248-bd097a90-11a1-11e8-8213-617eee7289c0.png)

### Loading spinner without footer:
![image](https://user-images.githubusercontent.com/32683958/36228348-ff20e198-11a1-11e8-9d43-21130ff5e5aa.png)

### Error message without footer:
![image](https://user-images.githubusercontent.com/32683958/36228319-ebd59a66-11a1-11e8-88da-f3394736de18.png)

## Acceptance criteria:
* Show the footer when displaying the loading spinner or error message

PR #885, opened 2018-02-15
Add footer on error and loading screens

lowellrex:
Connects #882.

Show footer when the search results page is loading or when any download page displays an error. Adds two new components: DownloadPageFooter and Button. We can probably convert some of the existing `<button>` tags to the new Button component (looking at you DownloadListContainer), but I will leave that for a later refactor (DownloadListContainer has plenty of cleaning up left to do).


client/src/components/Button.jsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+
+export const BUTTON_ALIGN_RIGHT = 'BUTTON_ALIGN_RIGHT';
+
+export default class Button extends React.PureComponent {
```
mdbenjam: If we're going to make a button component, I would suggest we only build styles of button into it. Things like right alignment can easily be achieved by wrapping the button in a span or div. Happy to chat more about this, I have very strong feelings about it since I think the Caseflow button never worked well.

lowellrex: Removed the Button component.

mdbenjam: To be clear, I'm fine with a button component, I think it's good, it just should have button styling rather than positioning styling. Taking it out is okay here too.

Issue #884, opened 2018-02-14
No error on welcome page when invalid input


mdbenjam:
When a bad input is provided on the welcome page, there is no error state to alert the user.

![badinput](https://user-images.githubusercontent.com/3885236/36231431-fb33be66-11ab-11e8-8c65-e79df5b0234a.gif)

PR #887, opened 2018-02-15
Display error banner on search page if there is an error

lowellrex:
Connects #884.


lowellrex:
I'm not enamored with applying these styles using glamor, but I did it in this case to match the current look of the alert we see in the v1 front-end. Open to literally any other solution to make this look right.

client/src/containers/WelcomeContainer.jsx
```diff
@@ -7,13 +7,32 @@ import AppSegment from '@department-of-veterans-affairs/caseflow-frontend-toolki
...
   fontStyle: 'italic',
   textAlign: 'center'
 });
 
+const userFriendlyErrorFor = (errorStr) => {
+  let err = { title: 'Could not complete search for input Veteran ID',
+    body: errorStr };
+
+  switch (errorStr) {
+  case '400 (Bad Request) File Number is invalid, must be 8 or 9 digits':
```
mdbenjam: This seems brittle. It might be better if we store the status in the store, so we can just check against that?

client/src/components/AlertBanner.jsx
```diff
@@ -17,7 +18,7 @@ export default class DownloadProgressBanner extends React.PureComponent {
       break;
     }
 
-    return <div className={`usa-alert ${alertTypeClass}`} role="alert">
+    return <div className={`usa-alert ${alertTypeClass}`} role="alert" {...css({ marginBottom: '2.4rem' })}>
```
mdbenjam: Do the templates have this css? How are they applying it? Via a class?

lowellrex: The [current html templates](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/7f2ec26fe4975d87102f7bde68af5a0e990529c2/app/views/downloads/new.html.erb#L15) accomplish the spacing by wrapping this banner in a `<p>` tag. Our linter complains about divs being child elements of p tags if we do that in our current setup (though I am now having trouble replicating that complaint).

mdbenjam: I think this should be declared outside of the render function. Otherwise you'll have to regenerate the css on every render. That could be expensive.

client/src/components/AlertBanner.jsx
```diff
@@ -1,6 +1,10 @@
 import { css } from 'glamor';
 import React from 'react';
 
+const childTextStlying = css({
```
mdbenjam: This creates a lot of coupling between this component and it's children. You're expecting specific children. Why not just not pass in a <p> tag?

client/src/components/AlertBanner.jsx
mdbenjam: Also, I'd move this css outside this component. You might want alerts that don't have this space.

lowellrex:
Removed the css from the AlertBanner component

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -40,6 +40,6 @@ def bgs_service
   end
 
   def invalid_file_number
-    render json: { status: "File Number is invalid, must be 8 or 9 digits" }, status: 400
+    render json: { status: "File number is invalid. Veteran IDs must be 8 or more characters and contain only numbers" }, status: 400
```
mdbenjam: Period at the end.

Issue #888, opened 2018-02-15
Populate the recent downloads list with v2 entities


lowellrex:
The last v1 piece in the react app is the recent documents list on the welcome/search page. This ticket encompasses all of the work it takes to fully migrate that view to v2 models.

![image](https://user-images.githubusercontent.com/32683958/36274566-1dce9676-1256-11e8-80a3-86997b0b3bee.png)

## Acceptance criteria:
* The "History" list displays history of v2 manifest document downloads and not v1 downloads

PR #889, opened 2018-02-15
Move recent downloads list to v2 api

lowellrex:
Connects #888.

Was previously using the v1 models to populate the list in the react single page app with some dummy data. This code change makes it so that we use the v2 endpoint to populate that list.


client/src/containers/RecentDownloadsContainer.jsx
```diff
@@ -19,13 +36,34 @@ class RecentDownloadsContainer extends React.PureComponent {
...
-          { this.props.recentDownloads.map((download) => <RecentDownloadRow download={download} key={download.id} />) }
+          { this.props.recentDownloads.map((download) => (
+            <tr id={`download-${download.id}`} key={download.id}>
+              <td>
+                {`${download.attributes.veteran_first_name} ${download.attributes.veteran_last_name}`}
+                <span className="cf-subtext"> ({download.attributes.file_number}) - Expires on&nbsp;
+                  {download.attributes.zip_expiration_date} </span>
+              </td>
+              <td className="ee-actions-cell">
+                <Link to={`/downloads/${download.id}`} onClick={this.props.resetDefaultManifestState}>
```
mdbenjam: What was the reason again we can't do this on the the `downloads/:id` page? I'm not confident what the order is here. Will this be guaranteed to run before the page transition?

client/src/containers/RecentDownloadsContainer.jsx
```diff
@@ -1,9 +1,26 @@
...
-import RecentDownloadRow from '../components/RecentDownloadRow';
+import Link from '@department-of-veterans-affairs/caseflow-frontend-toolkit/components/Link';
+
+import { MANIFEST_DOWNLOAD_STATE } from '../Constants';
+import { resetDefaultManifestState } from '../actions';
+import { getDownloadHistory } from '../apiActions';
+import AlertIcon from '../components/AlertIcon';
+
+const linkTextForStatus = (status) => {
+  const icon = status === MANIFEST_DOWNLOAD_STATE.FAILED ? <AlertIcon /> : null;
```
mdbenjam: I can't get the alert icon to show up :/

client/src/containers/RecentDownloadsContainer.jsx
lowellrex: Moved this logic over to the [downloads page](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/889/commits/0ebac37b9b5579bf1f0e945e582a7c0bb2c16e08#diff-d2b483870579c30dd3240dd7054c3169R30)

Issue #892, opened 2018-02-16
Add confirmation message when user attempts to download incomplete efolder zip


lowellrex:
When a user clicks on the button to download an efolder that does not contain all of the documents in that efolder (say VVA is unavailable so it only contains VBMS docs), we[ currently present a modal alert](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/736c41580f2ae9a35551243c4bfdf181e5cbb8bc/app/views/downloads/_confirm_download_anyway_modal.html.erb) to ensure they know the efolder is incomplete. We need to preserve that behaviour in v2.

## Acceptance criteria
* User is presented with a modal message when attempting to download an efolder zip that is incomplete

PR #896, opened 2018-02-22
Add confirmation message when downloading incomplete zip

lowellrex:
Connects #892.

Clicking the "Download anyway" button now presents the user with the following modal:

image


client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -25,6 +26,11 @@ import ManifestDocumentsTable from '../components/ManifestDocumentsTable';
 import { aliasForSource, documentDownloadComplete } from '../Utils';
 
 class DownloadProgressContainer extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    this.state = { confirmDownloadModalIsVisible: false };
```
mdbenjam: In Caseflow, we had a rule to put all app state in Redux. I think you've been doing that as well in eFolder, so it's a little confusing to have separate local state for this one property.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -125,16 +132,65 @@ class DownloadProgressContainer extends React.PureComponent {
...
+          <button
+            type="button"
+            className="usa-button-outline cf-action-closemodal cf-push-left"
+            data-controls="#confirm-download-anyway"
+            onClick={this.hideConfirmDownloadModal}
+          >
+            Go back
+          </button>
+          { this.wrapInDownloadLink(
+            <button
```
mdbenjam: A button inside a link seems like a bad pattern to me. Link has an onclick prop we could use instead.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -173,19 +167,19 @@ class DownloadProgressContainer extends React.PureComponent {
...
-              className="cf-push-right usa-button usa-button-secondary"
-              onClick={this.hideConfirmDownloadModal}
-            >
-                Download anyway
-            </button>
-          )
-          }
+          <button
+            className="cf-push-right usa-button usa-button-secondary"
+            onClick={() => {
```
mdbenjam: In general we should avoid inline defining functions in react, as we have to redefine the function each time it is rendered, and it could trigger re-renders of the element it's passed to. We should just make it a method.

Issue #898, opened 2018-02-26
Change error page button to secondary


Chingujo:
As a eFolder Express user, I want to know quickly what elements in the UI I can interact with. 

### Acceptance criteria
- Verify that all buttons that use the "gray styling" in eFolder express in relation to another button is changed to 'Default Secondary'
- Verify that the copy of this new styled button is 'Retry missing files'

### Mock 
Other parts of this mock may be out of date, please just use the buttons portion. 
![05_errors](https://user-images.githubusercontent.com/23316107/36693736-38e113b4-1b0a-11e8-8f4e-bcc037fab6ea.png)

Chingujo:
@lowellrex 

The button change from early. Not urgent please prioritize as you see fit =D

PR #899, opened 2018-02-27
Update button text and styling

lowellrex:
Connects #898. Purely a cosmetic change here.

image


Chingujo:
@lowellrex 

This looks great! Only comment would be to add space between the two buttons (around 20 px)

lowellrex:
Updated the image to show what the added spacing looks like

Issue #678, opened 2017-11-03
[FE] eFolder Express VVA Not Included Final Document Page


Chingujo:
### Story
As a FOIA and Privacy and eFolder Express user, I want to know if my efolder is complete and contains both VBMS and VVA documents.  

### Acceptance Criteria 
1. Verify that in the complete download page a VVA banner alert appears if the completed folder does not contain VVA documents 
1. Verify that there is an alert banner with icon per style guide above the table
1. Verify that this alert banner has the following header in h2 bold styling per style guide "We could not retrieve VVA documents at this time" 
1. Verify that paragraph one in the banner has the following copy in body styling (17px): "To continue downloading VBMS files without VVA documents, click the ‘Download VBMS documents’ button below."
1. Verify that paragraph two has the following copy "You can also wait try again by clicking the ‘Retry download’ button below or search for another efolder."
1. Verify that "search for another efolder" is a link that takes you back to the home page 
1. Verify that the button in the alert has the copy 'Download VBMS documents' that downloads the retrieved files 
1. Verify that there is a button to the left of 'Download VBMS documents' with the copy 'Search for another efolder'
1. Verify that clicking that button takes you back to the home page
1. Verify that button copy at the bottom of the table is 'Download VBMS documents' 

### Mocks

![vva documents not found button version](https://user-images.githubusercontent.com/23316107/32904354-32a49be6-cac5-11e7-92d1-60a887a4a9c1.jpg)

### Alternative copy 

![vva documents not found button version v2](https://user-images.githubusercontent.com/23316107/32904280-0deea184-cac5-11e7-8329-afc4332e8fc2.jpg)

Chingujo:
@lakohl 

Mocks on the download complete page based on Invision feedback and our sessions:
- Blue alt secondary button with search for efolder text versus a blue link
- Clearer messaging around button actions


 Only difference in the two is the copy if you prefer one.

PR #904, opened 2018-02-27
Display specific banner on downloads page if manifest source unavailable

lowellrex:
Connects #678. Show a more specific banner on the downloads progress page if a manifest source is unavailable.

image


client/src/Constants.js
```diff
@@ -20,6 +20,6 @@ export const MANIFEST_DOWNLOAD_STATE = {
 export const MANIFEST_SOURCE_FETCH_STATE = {
   NOT_STARTED: 'initialized',
   IN_PROGRESS: 'pending',
-  SUCCEEDED: 'finished',
+  SUCCEEDED: 'success',
```
lowellrex: This was mapped incorrectly previously

lowellrex:
I thought we would need to add a new icon for this change so I went ahead and refactored all of the different Icons into a single file. Even though we didn't end up needing to add a new icon I decided to leave the Icons refactor in this PR because I think it is a good refactor. Happy to remove if that change appears to overload this PR

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -66,6 +69,44 @@ class DownloadProgressContainer extends React.PureComponent {
...
+            <button
+              className="usa-button-outline"
+              onClick={this.restartDocumentDownload}
+              {...css({ marginLeft: '2rem' })}
+            >
+              Retry download
+            </button>
+          </li>
+        </ul>
+      </AlertBanner>;
```
lowellrex: We can return this AlertBanner any of three places in this function which seems a little fishy to me. However, I'm struggling for a way to factor them out without harming the readability of the code as it is now. Open to suggestions.

mdbenjam: I think it's fine.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -66,6 +69,44 @@ class DownloadProgressContainer extends React.PureComponent {
   restartDocumentDownload = () => this.props.startDocumentDownload(this.props.manifestId, this.props.csrfToken);
 
   completeBanner() {
+    const sourcesForStatus = {};
+
+    for (const status of Object.values(MANIFEST_SOURCE_FETCH_STATE)) {
```
mdbenjam: This section seems overly complex to me. It seems doing two filters to get the SUCCEEDED and FAILED would be sufficient here. And I think it'd improve readability.

Issue #900, opened 2018-02-27
Change error message copy


Chingujo:
As a user, I want to clearly know my actionable steps when missing files are rendered. 

### Acceptance criteria
1. Verify that the title of this error message is Some files could not be retrieved
1. Verify that the first line of the message is:
   - Files that could not be retrieved are shown in the errors tab. 
1. Verify that the second line of the message is: 
    - You can still download the completed files by clicking the ‘Download anyway’ button or try to download the missing files using the “Retry missing files” button.  

### Mock
![05_errors](https://user-images.githubusercontent.com/23316107/36736532-de9516b0-1ba6-11e8-87d9-f75972a5239e.png)

PR #906, opened 2018-02-27
Change download results page copy

lowellrex:
Connects #900.

Before:
image

After:
image


client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -67,9 +67,10 @@ class DownloadProgressContainer extends React.PureComponent {
 
   completeBanner() {
     if (this.props.documentsForStatus.failed.length) {
-      return <AlertBanner title="Some files couldn't be added to eFolder" alertType="error">
-        <p>eFolder Express wasn't able to retrieve some files. Click on the 'Errors' tab below to view them</p>
-        <p>You can still download the rest of the files by clicking the 'Download anyway' button below.</p>
+      return <AlertBanner title="Some files could not be retrieved" alertType="error">
+        <p>Files that could not be retrieved are shown in the errors tab.</p>
+        <p>You can still download the completed files by clicking the ‘Download anyway’ button or&nbsp;
```
Chingujo: I think there is an extra space here but other than that looks good to go @lowellrex!

lowellrex: Good catch! Corrected!

PR #909, opened 2018-02-28
Default userDisplayName to Menu

lowellrex:
Set the userDisplayName that will display as the text on the dropdown menu to "Menu" if current_user is not set. Such a case can occur on the unauthorized and out-of-service pages. This behaviour mirrors [what we do](https://github.com/department-of-veterans-affairs/caseflow/search?utf8=%E2%9C%93&q=userDisplayName%3D%22Menu%22) in the main caseflow repo.


lowellrex:
I tested the display properties by just changing the code to:
```
userDisplayName: nil || "Menu"
```

As for the behaviour part, that should be easier to test once #870 is merged (because of the Route/PrivateRoute distinction contained in that changeset)... Though I'll have to think about how we would do that test... and if I can ever figure out how to get those tests to run...

pkarman:
react obsoletes this

PR #915, opened 2018-03-02
Use chromedriver instead of phantomjs

lowellrex:
As part of the efolder react frontend rewrite, we need some way to run tests against the frontend. We decided to use chromedriver to test the rendered pages instead of phantomjs because that is what the main caseflow apps use due to shortcomings with phantomjs.

This PR started as part of #870, but I separated it out to its own PR so we can more easily isolate the problems we've had getting these tests to run in Travis.


spec/features/dashboard_spec.rb
```diff
@@ -111,18 +111,6 @@
...
-    User.authenticate!(roles: ["System Admin"])
-
-    visit "/stats"
-    click_on "Daily"
-
-    find('*[role="button"]', text: "Time to Manifest").trigger("click")
-    expect(page).to have_content("Time to Manifest (95th percentile) 30.37 sec")
-    find('*[role="button"]', text: "Time to Manifest").trigger("click")
-    expect(page).to have_content("Time to Manifest (median) 16.37 sec")
-  end
```
lowellrex: Removed this test entirely because I couldn't get it to work after several hours of fighting with it. Seems like the root of the issue was that we couldn't find the element we needed to hover over (the first child of the svg element). Also we needed to change `...trigger("click")` to `...click()` to satisfy a difference between PhantomJS and Chromedriver, but that was not causing our issue.

I think removing this subtest is fine because the stats page will be removed after we fully transition to the react app.

spec/features/download_spec.rb
```diff
@@ -443,7 +443,7 @@ def assert_coachmark_does_not_exist
     expect(Download.where(id: fake_id)).to be_empty
 
     visit download_download_path(fake_id)
-    expect(page.status_code).to be(404)
+    expect(page).to have_content("Something went wrong...")
```
lowellrex: Chromedriver does not expose the status code of an HTTP response so we have to use some text on the page as our signal that we received a 404.

spec/features/download_spec.rb
```diff
@@ -494,9 +494,11 @@ def expect_page_to_not_have_coachmarks
 
     expect_page_to_have_coachmarks
 
+    DownloadHelpers.clear_downloads
     first(:link, "Download efolder").click
-    expect(page.response_headers["Content-Type"]).to eq("application/zip")
-    expect(page.response_headers["Content-Length"]).to eq(File.size(download_documents.zip_path).to_s)
+    DownloadHelpers.wait_for_download
+    expect(DownloadHelpers.filesize).to eq(File.size(download_documents.zip_path))
+    DownloadHelpers.clear_downloads
```
lowellrex: Another workaround for a difference between Chromedriver and PhantomJS' capabilities.

spec/rails_helper.rb
```diff
@@ -22,13 +23,49 @@
...
+    options: chrome_options
+  }
+
+  Sniffybara::Driver.current_driver = Sniffybara::Driver.new(app, options)
+end
+
+Capybara.default_driver = :parallel_sniffybara
+# the default default_max_wait_time is 2 seconds
+Capybara.default_max_wait_time = 20
+
```
lowellrex: Mirror the settings that caseflow uses.

Gemfile
```diff
@@ -118,13 +118,12 @@ group :development, :test do
...
 group :test do
+  gem "capybara", "2.6.2"
   gem "rspec"
   gem "rspec-rails"
-  gem "timecop"
-  # gem 'guard-rspec'
-  gem "capybara", "2.6.2"
   gem "simplecov"
-  gem "sniffybara", git: "https://github.com/department-of-veterans-affairs/sniffybara.git", branch: "axe"
+  gem "sniffybara", git: "https://github.com/department-of-veterans-affairs/sniffybara.git"
```
lowellrex: Moved sniffybara from the `axe` branch to master because I think it's probably better practice to point to master when we can (and we can here).

lowellrex:
Finally ready for review!

For reference, I pulled a lot of the configs directly from caseflow's [.travis.yml](https://github.com/department-of-veterans-affairs/caseflow/blob/a5e496d53bbee6c1a137f960369e1068e30a4613/.travis.yml#L35), [rails_helper.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/a5e496d53bbee6c1a137f960369e1068e30a4613/spec/rails_helper.rb#L43), and [download_helper.rb](https://github.com/department-of-veterans-affairs/caseflow/blob/master/spec/support/download_helper.rb). The rest of the changes were to accommodate those core changes.

spec/rails_helper.rb
```diff
@@ -22,13 +23,47 @@
...
+  options = {
+    port: 51_674,
+    browser: :chrome,
+    options: chrome_options
+  }
+
+  Sniffybara::Driver.current_driver = Sniffybara::Driver.new(app, options)
+end
+
+Capybara.default_driver = :virtual_framebuffer_chrome
```
mdbenjam: what does this do?

spec/rails_helper.rb
```diff
@@ -22,13 +23,47 @@
...
+# the default default_max_wait_time is 2 seconds
+Capybara.default_max_wait_time = 20
```
mdbenjam: I'm not sure we should copy this over from caseflow. This can make tests run a lot longer.

spec/rails_helper.rb
lowellrex: This tells Capybara to use the driver we set up in the block above as the default driver (I renamed it from Caseflow's `parallel_sniffybara` to `virtual_framebuffer_chrome` because I don't think we run these tests in parallel for efolder).

lowellrex: Running the tests without this line results in a bunch of `Capybara::NotSupportedByDriverError: Capybara::Driver::Base#execute_script` errors. I'm guessing the default default_driver (Capybara::RackTest::Driver?) does not execute javascript.

Issue #916, opened 2018-03-02
Retry 503 Response from VVA (Savon::HTTPErrorSidekiq)


lowellrex:
We see a fair number of the following errrs and generally just ignore them.
> Savon::HTTPErrorSidekiq
> errorHTTP error (503): upstream connect error or disconnect/reset before headers
(https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/667/ for example)

In an effort to make our alerts more meaningful and actionable (as well as making our applications more robust and responsive), we should either handle this error or silence. I believe that we can use Envoy to retry these errors.

PR #917, opened 2018-03-05
Upgrade to latest version of connect_vva

lowellrex:
Connects #916.

Envoy is already handling these retries, and they are just all failing. With [the change to connect_vva](https://github.com/department-of-veterans-affairs/connect_vva/pull/23), we rethrow this error as a VVA::ClientError and will catch it in DownloadManifestJob. The ExceptionLogger will bypass sending this particular error to sentry, and we will set the download as errored in [DownloadAllManifestJob](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/864287fd3d818db516bb899f9dccf402ac642459/app/jobs/download_all_manifest_job.rb#L13).


lowellrex:
Still TODO: Point connect_vva dependency to the newly updated version whenever it is merged to master.

PR #925, opened 2018-03-09
[WIP] Nemo/rails upgrade

nemodjuric:
Connects to #834 
This is part of Rails 5 upgrade.
Similar to https://github.com/department-of-veterans-affairs/caseflow-feedback/pull/171


app/jobs/application_job.rb
```diff
@@ -0,0 +1,2 @@
+class ApplicationJob < ActiveJob::Base
```
lowellrex: How come we need to do this now? Are there issues with jobs inheriting directly from `ActiveJob::Base` in Rails 5?

I quickly glanced at the upgrade docs and couldn't see anything obvious to my eyes http://guides.rubyonrails.org/5_0_release_notes.html#active-job

app/models/application_record.rb
```diff
@@ -0,0 +1,3 @@
+class ApplicationRecord < ActiveRecord::Base
+  self.abstract_class = true
```
lowellrex: I'm not totally sure what this abstract class is doing? Is this necessitated by the Rails 5 upgrade somehow?

config/initializers/assets.rb
```diff
@@ -8,6 +8,7 @@
 
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
+# Rails.application.config.assets.precompile += %w( search.js )
```
lowellrex: I'm a little confused about this comment here. Are we anticipating having a javascript filed named search.js in efolder? Or is this line for some other purpose?

app/jobs/application_job.rb
nemodjuric: This is done for this reason: https://github.com/rails/rails/issues/16976
You can find details about Rails 5 upgrade here: http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#activejob-now-inherits-from-applicationjob-by-default
Also, you can check out conversation we had about feedback update, you might find it useful. 
https://github.com/department-of-veterans-affairs/caseflow-feedback/pull/171

app/models/application_record.rb
nemodjuric: abstract_class is telling Rails here that we don't necessarily have table named ApplicationRecord in database, so other activerecord models can extend this class and have their own database table names.

PR #926, opened 2018-03-09
Add new tests for react version of eX

mdbenjam:
With the react version of the app starting to be rolled out, we need to update our feature. This PR creates a new file of feature tests for the react version of eX.

Our old feature tests depended heavily on the backend, the new tests try to be more agnostic. Instead of setting up complicated data structures we go through the application flow to set things up. In some places this is not possible but in most it is.

Some tests that were not recreated in this file are:
```
- Test that demo mode works.
- Download with manifest_fetch_error status. (We removed that status)
- Delete cache tests. (Not sure what these are...)
```


spec/features/download_spec.rb
```diff
@@ -36,110 +55,61 @@
     expect(page).to have_content("Test VA Saml")
   end
 
-  scenario "Download coachmarks" do
```
mdbenjam: We removed coachmarks from v2.

spec/features/download_spec.rb
```diff
@@ -36,110 +55,61 @@
...
     expect(page).to have_content "We are gathering the list of files in the eFolder now"
-    expect(page).to have_current_path(download_path(@download))
-    expect(page.evaluate_script("window.DownloadStatus.intervalID")).to be_truthy
-    expect(DownloadAllManifestJob).to have_received(:perform_later)
 
-    search = Search.where(user: @user).first
-    expect(search).to be_download_created
-  end
-
-  scenario ": initial wait" do
```
mdbenjam: Moved into above test.

spec/features/download_spec.rb
```diff
@@ -36,110 +55,61 @@
...
-    fill_in "Search for a Veteran ID number below to get started.", with: "1234"
-    click_button "Search"
-    expect(page).to have_css(".cf-loading-indicator .with-icon")
+    expect(manifest).to_not be_nil
+    expect(manifest.veteran_first_name).to eq("Stan")
+    expect(manifest.veteran_last_name).to eq("Lee")
+    expect(manifest.veteran_last_four_ssn).to eq("2222")
   end
 
-  scenario "Searching for an errored download tries again" do
```
mdbenjam: @aroltsch In v1 we would apparently retry on subsequent downloads if there are no documents. Curious why we did that, and not sure if we should continue that behavior. @Chingujo if you have any thoughts.

spec/features/download_spec.rb
```diff
@@ -1,13 +1,65 @@
...
 
-    allow(DownloadAllManifestJob).to receive(:perform_later)
-    allow(DownloadFilesJob).to receive(:perform_later)
+    FeatureToggle.enable!(:efolder_react_app)
+
     User.authenticate!
+
+    allow_any_instance_of(Fakes::BGSService).to receive(:fetch_veteran_info).with(veteran_id).and_return(veteran_info)
+    allow_any_instance_of(Fakes::BGSService).to receive(:valid_file_number?).with(veteran_id).and_return(true)
+    allow_any_instance_of(Fakes::BGSService).to receive(:valid_file_number?).with(invalid_veteran_id).and_return(false)
```
lowellrex: Clever!

spec/features/download_spec.rb
```diff
@@ -20,7 +72,7 @@
     click_on "Sign In"
 
     puts page.current_path
-    expect(page).to have_current_path(root_path)
+    expect(page).to have_current_path("/")
```
lowellrex: Doesn't really matter, but I think this can be changed back to `root_path` if you'd like since we declare it on line 52.

lowellrex:
I think we should test the scenario when somebody attempts to view an expired download URL similar to [what we have for v1](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/6c3f400832512a7878294e70ccec50e4c9840c2f/spec/features/download_spec.rb#L257). I was suspicious that we don't currently handle this scenario properly so I whipped up the following test (which fails... that fails?). Will right up an issue to resolve this bug in a minute.
```
  context "When document manifests have expired" do
    let!(:manifest) { Manifest.find_or_create_by_user(file_number: veteran_id, user: @user) }
    let!(:source) do
      [
        manifest.sources.create(status: :success, name: "VBMS"),
        manifest.sources.create(status: :success, name: "VVA")
      ]
    end
    let!(:records) do
      [
        source[0].records.create(
          version_id: "1",
          series_id: "101",
          mime_type: "application/pdf",
          type_id: Caseflow::DocumentTypes::TYPES.keys.sample
        )
      ]
    end

    before do
      manifest.sources.each do |src|
        src.update!(fetched_at: Time.zone.now - (Manifest::UI_HOURS_UNTIL_EXPIRY + 1).hours )
      end
    end

    scenario "Visiting URL of expired download restarts download" do
      visit "/downloads/#{manifest.id}"
      expect(page).to have_css(".ee-page-loading .ee-page-loading-icon")
    end
  end
```

spec/features/react_download_spec.rb
```diff
@@ -0,0 +1,495 @@
+require "rails_helper"
+
+RSpec.feature "Downloads" do
```
aroltsch: Mark, would it be possible to separate this file into multiple files?

spec/features/react_download_spec.rb
```diff
@@ -0,0 +1,495 @@
...
+      expect(page).to have_content("forbidden: sensitive record")
+    end
+  end
+
+  context "When veteran case folder has no documents" do
+    before do
+      allow(Fakes::DocumentService).to receive(:v2_fetch_documents_for).and_return([])
+    end
+
+    scenario "Download with no documents" do
```
aroltsch: Is it a duplicate of https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/926/files#diff-c96c0379f6eb6a0058e35f18456b4aabR126?

spec/features/react_download_spec.rb
```diff
@@ -0,0 +1,495 @@
...
+      allow(Fakes::VBMSService).to receive(:v2_fetch_documents_for).and_raise(VBMS::ClientError)
+    end
+
+    scenario "Download with VBMS connection error" do
+      perform_enqueued_jobs do
+        visit "/"
+        fill_in "Search for a Veteran ID number below to get started.", with: veteran_id
+        click_button "Search"
+
+        expect(page).to have_css ".usa-alert-heading", text: "We are having trouble connecting to VBMS"
```
aroltsch: Would it display VVA documents successfully?

spec/features/react_download_spec.rb
```diff
@@ -0,0 +1,495 @@
...
+      allow(Fakes::VVAService).to receive(:v2_fetch_documents_for).and_raise(VVA::ClientError)
+    end
+
+    scenario "Download with VVA connection error" do
+      perform_enqueued_jobs do
+        visit "/"
+        fill_in "Search for a Veteran ID number below to get started.", with: veteran_id
+        click_button "Search"
+
+        expect(page).to have_css ".usa-alert-heading", text: "We are having trouble connecting to VVA"
```
aroltsch: Same question as above.

spec/features/react_download_spec.rb
mdbenjam: Yes I should totally do this.

lowellrex:
I am pretty sure this behaviour changed for the v2 react app, but I think we should add a test similar to [this one for v1](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/2fa4abcb5da1ff2e6b6fabaf4cfbe60233fe971d/spec/features/download_spec.rb#L245) (but the inverse) to confirm that if Amy (for example) searches for a given file number then Billie will be able to view that same efolder when searching for the same file number.

lowellrex:
Looks good! No show stoppers here for me!

spec/features/react_download_spec.rb
mdbenjam: Removed the duplicate

mdbenjam:
Addressed all of the comments. I did not create a test for "the scenario when somebody attempts to view an expired download URL". I think that should wait until we've fixed that scenario. I added a segment to the happy path test to ensure a different user would go straight to the success screen if they're downloading a case that someone else has already downloaded.

I also split the test into four files:
react_download_spec - to be renamed download_spec when we remove the original. This has happy path tests and one test I didn't know where to put (the "Download progress shows correct information" test) :D
authentication_spec - has login tests
backend_error_flows_spec - has tests for when VVA or VBMS are down
user_error_flows_spec - has tests for when user input causes an error

I don't love these file names, and am open to other ways of splitting them.

Issue #930, opened 2018-03-12
Limit number of simultaneous request when polling efolder API


lowellrex:
In the v1 version of the app we limit the number of simultaneous requests we make to the backend when polling for the status of a download (https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/assets/javascripts/application/download_status.js#L5). We should do the same thing for v2.

PR #931, opened 2018-03-12
Limit simultaneous poll requests

lowellrex:
Connects #930.

### DIFF
--- a/client/src/apiActions.jsx
+++ b/client/src/apiActions.jsx
@@ -74,10 +74,21 @@ const buildErrorMessageFromResponse = (resp) => {
   return description;
 };
 
+const MAX_SIMULTANEOUS_POLL_REQUESTS = 2;
+let outstandingPollRequestCount = 0;
+
 export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken) => (dispatch) => {
+  if (outstandingPollRequestCount >= MAX_SIMULTANEOUS_POLL_REQUESTS) {
+    setTimeout(() => {
+      dispatch(pollManifestFetchEndpoint(retryCount + 1, manifestId, csrfToken));
+    }, 1000);
+  }
+
+  outstandingPollRequestCount += 1;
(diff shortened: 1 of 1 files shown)


client/src/apiActions.jsx
```diff
@@ -74,10 +74,21 @@ const buildErrorMessageFromResponse = (resp) => {
...
 
+const MAX_SIMULTANEOUS_POLL_REQUESTS = 2;
+let outstandingPollRequestCount = 0;
+
 export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken) => (dispatch) => {
+  if (outstandingPollRequestCount >= MAX_SIMULTANEOUS_POLL_REQUESTS) {
+    setTimeout(() => {
+      dispatch(pollManifestFetchEndpoint(retryCount + 1, manifestId, csrfToken));
+    }, 1000);
+  }
```
lowellrex: I'm 95% sure this approach works. I'm roughly 5% sure this is the correct approach to use. Still trying to get a hang of javascript idioms, so please let me know if there are better ways to do this!

client/src/apiActions.jsx
```diff
@@ -74,10 +74,21 @@ const buildErrorMessageFromResponse = (resp) => {
...
 };
 
+const MAX_SIMULTANEOUS_POLL_REQUESTS = 2;
+let outstandingPollRequestCount = 0;
+
 export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken) => (dispatch) => {
+  if (outstandingPollRequestCount >= MAX_SIMULTANEOUS_POLL_REQUESTS) {
+    setTimeout(() => {
+      dispatch(pollManifestFetchEndpoint(retryCount + 1, manifestId, csrfToken));
+    }, 1000);
```
mdbenjam: This doubles the rate of querying if this happens while we're querying for the document download rather than the manifest getting. I'm wondering if we should just standardize this value at 2 seconds and factor it out into a constant to make things simpler.

lowellrex:
Haha works for me!

Issue #928, opened 2018-03-12
React Design Pass


Chingujo:
Looked through react update for eFolder Express. Things look pretty Ace and great job @lowellrex. 
Here are a few nits I found and happy to make these as individual issues. Lmk if I can help, provide better mocks, or if I am missing something =D. 

### Font
- I may be going crazy but the kerning seems different between our font and the ones we have on apps even though both is Source Sans. 

#### Current
<img width="590" alt="screen shot 2018-03-12 at 1 32 21 pm" src="https://user-images.githubusercontent.com/23316107/37299487-d25af18c-25f9-11e8-8996-b2925c4e5872.png">

#### Reference 
<img width="676" alt="screen shot 2018-03-12 at 1 29 52 pm" src="https://user-images.githubusercontent.com/23316107/37299445-b65a1a80-25f9-11e8-8c2d-5fc4727813ca.png">

### Banner
Creating consistent banner messages and quick notes on styling. 
#### Font of title
- Inspection shows Source Sans. However, title is a serif (probably Meriwether) 
<img width="134" alt="screen shot 2018-03-12 at 12 24 57 pm" src="https://user-images.githubusercontent.com/23316107/37297959-6f429ffe-25f5-11e8-998f-ee3a0765817d.png">

- No periods at the end of major alerts
<img width="1063" alt="screen shot 2018-03-12 at 12 21 03 pm" src="https://user-images.githubusercontent.com/23316107/37298019-9b95bd52-25f5-11e8-8831-5cc680121c5b.png">

- Spacing should be tighter between the lines (usually not a crazy big deal but this is about 2x more is what I see) 
<img width="994" alt="screen shot 2018-03-12 at 1 02 59 pm" src="https://user-images.githubusercontent.com/23316107/37298057-b5bd7242-25f5-11e8-9cb2-3671cdbfecec.png">

_Reference_ 
<img width="659" alt="screen shot 2018-03-12 at 1 03 35 pm" src="https://user-images.githubusercontent.com/23316107/37298092-ced679c2-25f5-11e8-896f-32c545585171.png">

### Header and veteran ID 
Looks like our fonts are bigger and the Veteran ID is bolded. Can we change styling to match certification and our other apps?
#### Current
<img width="1148" alt="screen shot 2018-03-12 at 1 14 47 pm" src="https://user-images.githubusercontent.com/23316107/37298699-61229ac6-25f7-11e8-9258-d593071547b6.png">

#### Reference mock
[...]

PR #933, opened 2018-03-12
Reduce spacing between banner title and description

lowellrex:
Connects #928.

Tagging @Chingujo for visual design review, and tagging @mdbenjam for code review.

## Before
image

## After
image

## efolder v1 UI
image

### DIFF
--- a/client/src/components/AlertBanner.jsx
+++ b/client/src/components/AlertBanner.jsx
@@ -1,3 +1,4 @@
+import { css } from 'glamor';
 import React from 'react';
 
 export default class AlertBanner extends React.PureComponent {
@@ -23,7 +24,7 @@ export default class AlertBanner extends React.PureComponent {
     return <div className={`usa-alert ${alertTypeClass}`} role="alert">
       <div className="usa-alert-body">
         <h2 className="usa-alert-heading">{this.props.title}</h2>
-        <div className="usa-alert-text">{this.props.children}</div>
+        <div {...css({ marginTop: '-2rem !important' })} className="usa-alert-text">{this.props.children}</div>
       </div>
     </div>;
   }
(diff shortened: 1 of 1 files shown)


client/src/components/AlertBanner.jsx
```diff
@@ -23,7 +24,7 @@ export default class AlertBanner extends React.PureComponent {
     return <div className={`usa-alert ${alertTypeClass}`} role="alert">
       <div className="usa-alert-body">
         <h2 className="usa-alert-heading">{this.props.title}</h2>
-        <div className="usa-alert-text">{this.props.children}</div>
+        <div {...css({ marginTop: '-2rem !important' })} className="usa-alert-text">{this.props.children}</div>
```
lowellrex: We have ended up with a doubled top margin (compared to efolder UI v1) because we have a div wrapping a paragraph element (instead of only the paragraph element in v1) and both have top margins.

The reason we decided to add this wrapping div is so we could simply pass in paragraphs of text as child elements without forcing those child elements to also have the "usa-alert-text" class. There are a few examples of how this works in this file: https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/bd04b811eef545988955db99b47bb7cef5a950e6/client/src/containers/DownloadProgressContainer.jsx#L145

Though applying an inline margin to negate a css class property is hack-y I still think it is more desirable applying the "usa-alert-text" class to all paragraph elements passed to AlertBanner. I am willing to be convinced otherwise, and SUPER willing to hear alternative solutions that are better than both of those alternatives!

mdbenjam: Hmmm, yeah `!importants` are pretty bad. Also negative margins are pretty hacky. If the stated purpose is not to have children elements need the "usa-alert-text" class, we've ended up making the children and parents even more coupled. Now the component only works correctly if you pass in <p> as children.

I'm not sure what the best thing to do is here... One option is to reach into the children components and add the "usa-alert-text" class to all of them. That might be overkill though.

lowellrex: Just pushed up an alternative approach. Let me know what you think.

PR #935, opened 2018-03-13
Prevent error when visiting non-specified routes

lowellrex:
Whenever a user visits an efolder URL that is not specified in `config/routes.rb`, they will be presented with a rails error ([example sentry error](https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/1022/)). This change fixes that by redirecting non-specified routes to the unauthorized page if the user does not have access to the react single page app.

Tested this by altering the code to have `can_access_react_app?()` return false and navigating to *http://localhost:3000/browserconfig.xml* (triggers rails error) before this fix, as well as after (redirects to the react version of */unauthorized* or the html templates v1 version of */unauthorized* if `can_access_react_app?()` returns false).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -39,7 +39,7 @@ class ApplicationController < BaseController
   end
 
   def check_v2_app_access
-    serve_single_page_app if can_access_react_app?
+    can_access_react_app? ? serve_single_page_app : redirect_to("/unauthorized")
   end
 
   def initial_react_data
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -8,7 +8,7 @@ class ApplicationController < BaseController
   before_action :check_v2_app_access
 
   def serve_single_page_app
-    render "gui/single_page_app", layout: false
+    can_access_react_app? ? render("gui/single_page_app", layout: false) : redirect_to("/unauthorized")
   end
 
   def authenticate
@@ -39,7 +39,7 @@ class ApplicationController < BaseController
   end
 
   def check_v2_app_access
-    can_access_react_app? ? serve_single_page_app : redirect_to("/unauthorized")
+    serve_single_page_app if can_access_react_app?
   end
 
   def initial_react_data
--- a/app/controllers/health_checks_controller.rb
+++ b/app/controllers/health_checks_controller.rb
@@ -1,5 +1,6 @@
 class HealthChecksController < ApplicationController
   skip_before_action :authenticate
+  skip_before_action :check_v2_app_access
   newrelic_ignore_apdex
 
   def show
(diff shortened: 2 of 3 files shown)


app/controllers/application_controller.rb
```diff
@@ -39,7 +39,7 @@ def authorize
   end
 
   def check_v2_app_access
-    serve_single_page_app if can_access_react_app?
+    can_access_react_app? ? serve_single_page_app : redirect_to("/unauthorized")
```
lowellrex: We need to separate `check_v2_app_access()` and `check_v2_app_access()` into two different functions in order to avoid an infinite loop in the `ApplicationController.before_action()` callback.

app/controllers/application_controller.rb
```diff
@@ -8,7 +8,7 @@ class ApplicationController < BaseController
   before_action :check_v2_app_access
 
   def serve_single_page_app
-    render "gui/single_page_app", layout: false
+    can_access_react_app? ? render("gui/single_page_app", layout: false) : redirect_to("/unauthorized")
```
mdbenjam: Dont we only call `serve_single_page_app` if `can_access_react_app?` is true?

lowellrex: We also call it if we fall through all of the [defined routes in the config file](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/9296698d846cc45381f7b0227440e58dd4153e8c/config/routes.rb#L54). The idea there being that we can decouple the frontend code from the backend code even further by having the frontend routes [defined in only one place](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/9296698d846cc45381f7b0227440e58dd4153e8c/client/src/containers/InitContainer.jsx#L31), allowing the backend to be even more ignorant of the frontend.

app/controllers/health_checks_controller.rb
```diff
@@ -1,5 +1,6 @@
 class HealthChecksController < ApplicationController
   skip_before_action :authenticate
+  skip_before_action :check_v2_app_access
```
lowellrex: Need this to prevent requests from `/health-check` from failing. We already do this in [SessionsController](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/9296698d846cc45381f7b0227440e58dd4153e8c/app/controllers/sessions_controller.rb#L4), and after a deeper dive into [routes.rb](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/config/routes.rb), I think this is the last place we need to add this hook (maybe StatsController just because we haven't replicated that behaviour in the react version of the app, and visiting `/stats` now will result in a "page not found" message. This is probably fine for now).

## Request to `/health-check` before this change
image

## After
image

## stats page in react app
image

Issue #943, opened 2018-03-16
Display message on search page when no efolder exists for file number


lowellrex:
If a file number is valid (8 or 9 digits), we attempt a sensitivity check to determine if the user is allowed to access that file. That sensitivity check could fail because the file number does not have an associated efolder. Currently, in that case we present the user with a banner that reads "forbidden: sensitive record". We should present the user with a more precise banner in these cases.

## Acceptance Criteria:
* Display a banner on the search page notifying users that the file number they searched for does not have an associated efolder (if that is the case)

PR #944, opened 2018-03-16
Return veteran efolder not found message

lowellrex:
Connects #943.

Efolder Express UI v1's [DowloadsController](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/1284734617e65d66d2e4a4a216b80ec44c399f9b/app/controllers/downloads_controller.rb#L13) validates that an [efolder exists for a given file number](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/1284734617e65d66d2e4a4a216b80ec44c399f9b/app/models/search.rb#L73) (by way of `[Search.perform()](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/1284734617e65d66d2e4a4a216b80ec44c399f9b/app/models/search.rb#L30)`and displays a message on the [search results page](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/1284734617e65d66d2e4a4a216b80ec44c399f9b/app/views/downloads/new.html.erb#L2) if one does not. Similar behaviour for v2 except that we present the message on the search/landing page, before ever redirecting to the search results page.

## Before
image

## After
image

### DIFF
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -49,9 +49,6 @@ RSpec.feature "Downloads" do
       allow_any_instance_of(Fakes::BGSService).to receive(:fetch_veteran_info).and_return(nil)
 
       visit "/"
-
-      # Allow page time to load
-      sleep 1
       fill_in "Search for a Veteran ID number below to get started.", with: "DEMO1901"
       click_button "Search"
       expect(page).to have_content("could not find an eFolder with the Veteran ID")
--- a/spec/features/download_spec.rb
+++ b/spec/features/download_spec.rb
@@ -51,6 +51,9 @@ RSpec.feature "Downloads" do
       visit "/"
       fill_in "Search for a Veteran ID number below to get started.", with: "DEMO1901"
       click_button "Search"
+
+      # For debugging
+      it {puts page.body}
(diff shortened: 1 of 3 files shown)


spec/features/download_spec.rb
```diff
@@ -40,10 +40,19 @@
...
+    scenario "coachmarks are not displayed indicating that we are viewing the react app" do
       visit "/"
       expect(page).to_not have_content("See what's new!")
     end
+
+    scenario "requesting veteran that does not exist results in veteran not found error" do
+      allow_any_instance_of(Fakes::BGSService).to receive(:fetch_veteran_info).and_return(nil)
+
+      visit "/"
+      fill_in "Search for a Veteran ID number below to get started.", with: "DEMO1901"
```
mdbenjam: Can we change this to a non-demo number. DEMO numbers can go down different paths and is something I would like to eventually deprecate.

lowellrex: I went with `DEMO1901` in order to avoid mocking `Fakes::BGSService.valid_file_number()` (or `Fakes::BGSService.demo?()`). I'm happy to swap out this approach for that one though!

Issue #946, opened 2018-03-19
Ignore Logon ID Not Found errors


lowellrex:
Occasionally BGS raises an error because a user's login information was not found.

## [Example](https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/1036/)
> **Savon::SOAPFault**
> (S:Client) ID: {{UUID}}: Logon ID {{CSS_ID}} Not Found

Since the caseflow application cannot do anything to resolve this issue (nor can the caseflow team members), we should be ignoring these errors and returning a sensible message to the user.

# Acceptance Criteria:
* Application does not error when it encounters this type of error
* Sensible message is presented to user

PR #947, opened 2018-03-19
Return error to calling code if error is a PublicError

lowellrex:
Connects #946.

Since the caseflow application cannot do anything to resolve the "Savon::SOAPFault: (S:Client) ID: {{UUID}}: Logon ID {{CSS_ID}} Not Found" errors, we will pass along those errors to the calling client code.

This code change relies on changes in [ruby-bgs PR#51](https://github.com/department-of-veterans-affairs/ruby-bgs/pull/51).

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -3,15 +3,18 @@ class Api::V1::ApplicationController < BaseController
   before_action :authenticate_or_authorize
 
   rescue_from StandardError do |error|
-    ExceptionLogger.capture(error)
-
-    render json: {
-      "errors": [
-        "status": "500",
-        "title": "Unknown error occured",
-        "detail": "#{error} (Sentry event id: #{Raven.last_event_id})"
-      ]
-    }, status: 500
+    if error.respond_to?(:public_message)
+      forbidden(error.public_message)
+    else
+      ExceptionLogger.capture(error)
+      render json: {
+        "errors": [
+          "status": "500",
+          "title": "Unknown error occured",
+          "detail": "#{error} (Sentry event id: #{Raven.last_event_id})"
+        ]
+      }, status: 500
+    end
   end
 
   private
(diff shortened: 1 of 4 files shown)


spec/requests/api/v2/manifests_spec.rb
```diff
@@ -196,4 +196,18 @@
       expect(body["status"]).to match(/sensitive/)
     end
   end
+
+  context "When user does not exist in BGS" do
+    let(:error_string) { "Logon ID VACOHSOLO Not Found in the Benefits Gateway Service (BGS). Contact your ISO if you need assistance gaining access to BGS." }
+    before do
+      allow_any_instance_of(Fakes::BGSService).to receive(:check_sensitivity).and_raise(BGS::PublicError.new(error_string))
```
lowellrex: `BGS::PublicError` wasn't introduced until [ruby-bgs PR#51](https://github.com/department-of-veterans-affairs/ruby-bgs/pull/51) so this test will fail until that PR is merged and the Gemfile is updated to point to the latest ruby-bgs version.

app/controllers/api/v1/application_controller.rb
```diff
@@ -3,15 +3,18 @@ class Api::V1::ApplicationController < BaseController
...
-
-    render json: {
-      "errors": [
-        "status": "500",
-        "title": "Unknown error occured",
-        "detail": "#{error} (Sentry event id: #{Raven.last_event_id})"
-      ]
-    }, status: 500
+    if error.respond_to?(:public_message)
+      forbidden(error.public_message)
```
lowellrex: Decided to filter by behaviour instead of type here to maintain flexibility on both sides and not couple this implementation too tightly to the ruby-bgs code. This allows ruby-bgs to change the PublicError class that is thrown in these instances to not have a `public_message` method and this condition will no longer be met. This also allows us to have `VVA::ClientError`s with `public_message` methods in the future without having to rely on the type matching.

aroltsch: I personally prefer rescuing from the `BGS::PublicError` instead of relying on it having a `public_message` method
```
rescue_from BGS::PublicError do |error|
....
end
```

Issue #945, opened 2018-03-19
[Bug] Manifest sources are stuck in a pending state.


aroltsch:
I have noticed there is a manifest source that is stuck in a pending state. Theoretically, this should never happen because we of https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/jobs/v2/download_manifest_job.rb#L16. 

```
[#<ManifestSource id: 176, manifest_id: 88, status: "pending", name: "VBMS", fetched_at: "2018-03-12 18:01:24", created_at: "2018-03-08 17:38:43", updated_at: "2018-03-13 14:11:21">, #<ManifestSource id: 177, manifest_id: 88, status: "success", name: "VVA", fetched_at: "2018-03-18 21:45:32", created_at: "2018-03-08 17:38:43", updated_at: "2018-03-18 21:45:32">]
```

Because the status is pending, VBMS manifest was never attempted to fetch on 03/18 while VVA was successful. 

## AC:
1. Track down the issue 
2. Manifest sources can never be stuck in a pending state. 
3. Idea: can we have a timing mechanism on the duration of the job and if it lasts longer than a day (?) for example, set the manifest source to failed.

lowellrex:
Ran the cronjob ([it worked!](https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/1057/)) and changed the two manifest sources' (IDs 176, 272) statuses to failed.
```
ManifestSource.where(status: 1).where("updated_at < ?", 1.day.ago).map { |m| m.update!(status: 3) }
```

PR #951, opened 2018-03-20
Add cron job to check for stuck ManifestSources

lowellrex:
Connects #945.

Add this cron job so we can investigate the issue of ManifestSources getting stuck in the "pending" state closer to the first occurrence of this happening when (if?!) it happens next.

Have not yet reset the two ManifestSources that are stuck in this state because I want to confirm this runs by using those two ManifestSources as our guinea pigs.

### DIFF
--- a/app/jobs/alert_stuck_manifest_source_job.rb
+++ b/app/jobs/alert_stuck_manifest_source_job.rb
@@ -1,0 +1,16 @@
+class AlertStuckManifestSourceJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    stuck_manifest_ids = ManifestSource.find_by_sql("select * from manifest_sources where status = 1 and created_at + interval '1 day' < current_timestamp").map(&:id)
+
+    unless stuck_manifest_ids.empty?
+      msg = format(
+        "%<count>d ManifestSources with the following IDs have been stuck in the pending state for more than 1 day: %<ids>s",
+        count: stuck_manifest_ids.length,
+        ids: stuck_manifest_ids.join(", ")
+      )
+      ExceptionLogger.capture(msg)
+    end
+  end
+end
--- a/app/jobs/alert_stuck_manifest_source_job.rb
+++ b/app/jobs/alert_stuck_manifest_source_job.rb
@@ -3,11 +3,11 @@ class AlertStuckManifestSourceJob < ApplicationJob
 
   def perform
     if Date.current >= Date.new(2018, 5, 31)
-      # Remove the associate entry in config/sidekiq_cron.yml as well.
+      # Remove the associated entry in config/sidekiq_cron.yml as well.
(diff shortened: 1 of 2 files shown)


app/jobs/alert_stuck_manifest_source_job.rb
```diff
@@ -0,0 +1,22 @@
...
+    stuck_manifest_ids = ManifestSource.find_by_sql("select * from manifest_sources where status = 1 and created_at + interval '1 day' < current_timestamp").map(&:id)
+
+    unless stuck_manifest_ids.empty?
+      msg = format(
+        "%<count>d ManifestSources with the following IDs have been stuck in the pending state for more than 1 day: %<ids>s. Investigate! %<url>s",
+        count: stuck_manifest_ids.length,
+        ids: stuck_manifest_ids.join(", "),
+        url: "https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/945"
+      )
+      ExceptionLogger.capture(msg)
```
aroltsch: Instead of going through Sentry, we can just output the message to slack. The same way we do for establishClaim jobs in Caseflow. Also, I think ExceptionLogger.capture expects an object and not a string.

app/jobs/alert_stuck_manifest_source_job.rb
```diff
@@ -0,0 +1,22 @@
+class AlertStuckManifestSourceJob < ApplicationJob
+  queue_as :default
+
+  def perform
+    if Date.current >= Date.new(2018, 5, 31)
+      # Remove the associate entry in config/sidekiq_cron.yml as well.
+      ExceptionLogger.capture("Consider removing the AlertStuckManifestSourceJob if we haven't seen a stuck ManifestSource in a while")
+    end
+
+    stuck_manifest_ids = ManifestSource.find_by_sql("select * from manifest_sources where status = 1 and created_at + interval '1 day' < current_timestamp").map(&:id)
```
aroltsch: I think we should use updated_at. And I think AR is so much prettier :)
```
ManifestSource.where(status: 1).where("updated_at < ?", 1.day.ago).pluck(:id)
```

app/jobs/alert_stuck_manifest_source_job.rb
lowellrex: Hmmm, sending this directly to Slack would require us to add a SlackService [like Caseflow has](https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/slack_service.rb) as well as making changes to appeals-deployment to [make the required environment variables](https://github.com/department-of-veterans-affairs/appeals-deployment/blob/49297a706c6ba7391afef3bde1051205785f7716/ansible/roles/caseflow-certification/defaults/main.yml#L43) available to the SlackService.

Additionally, one nice thing about having this information in Sentry is that it works as an archive of historical information much better than Slack, and I think retaining this information might end up being useful.

aroltsch: Sounds good!

PR #952, opened 2018-03-20
Improve sensitive record message

lowellrex:
Added more expressive copy for sensitive records alert to match what [currently exists in efolder html](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/3a55456ea222374c7e7ee5217ae3621d7ce5b2e7/app/views/downloads/new.html.erb#L31) templates UI. Changing this copy should have no effect on the caseflow code that depends on this response since [that code](https://github.com/department-of-veterans-affairs/caseflow/blob/db575e1162ddf169074980fc97dca1ae826a4d55/app/services/external_api/efolder_service.rb#L19) only checks the HTTP response code and not the text of the message.

## Before
image

## After
image

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -27,8 +27,8 @@ class Api::V1::ApplicationController < BaseController
     forbidden("It looks like you don't have permission to view this eFolder. This usually happens if the eFolder contains sensitive information.")
   end
 
-  def forbidden(reason = "unspecified")
-    render json: { status: "forbidden: #{reason}" }, status: 403
+  def forbidden(reason = "Forbidden: unspecified")
+    render json: { status: reason }, status: 403
   end
 
   def missing_header(header)
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -23,6 +23,10 @@ class Api::V1::ApplicationController < BaseController
     render json: { status: "unauthorized" }, status: 401
   end
 
+  def sensitive_record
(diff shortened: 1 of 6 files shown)


app/controllers/api/v1/application_controller.rb
```diff
@@ -23,6 +23,10 @@ def unauthorized
     render json: { status: "unauthorized" }, status: 401
   end
 
+  def sensitive_record
+    forbidden("It looks like you don't have permission to view this eFolder. This usually happens if the eFolder contains sensitive information.")
+  end
+
   def forbidden(reason = "unspecified")
     render json: { status: "forbidden: #{reason}" }, status: 403
```
mdbenjam: I'm wondering if we should change this to `Forbidden:` I think the lowercase looks weird.

mdbenjam: Or maybe even just have the standard text be `Forbidden: unspecified` and have the reason parameter override the entire thing.

lowellrex: Should have relatively little impact on the rest of the code. https://github.com/department-of-veterans-affairs/caseflow-efolder/search?utf8=%E2%9C%93&q=forbidden&type=

Chingujo:
Request for small copy change T_T (ugh designers are the worst)

Title: We could not complete the search for this Veteran ID 

Body: This efolder contains sensitive information you do not have permission to view. Please contact your supervisor. 

Making the copy a bit more user friendly and providing actionable steps. When a user does not have permission to access they usually have to get special elevated access from their supervisor and it's not something that we can do (assuming here?). Trying to give actionable steps while reducing potential support tickets we can't help with regarding sensitivity.

PR #956, opened 2018-03-21
Recreating index with unique constraint

mdbenjam:
We want to have a DB level validation to ensure our manifests are unique per file_number. This PR recreates the index on the Manifest table to make it unique.

**Test plan**
- [ ] Ran the query in the first migration to determine how many manifests we'll need to delete in production. Only one, id: 464.
- [ ] Ran the migrations locally where I artificially created duplicate manifests, with associated manifest sources and records. Ensured it delete those extras before creating the unique index.
- [ ] Rolled back all the migrations to ensure we can go down as well as up.

### DIFF
--- a/db/migrate/20180321175607_add_manifest_index.rb
+++ b/db/migrate/20180321175607_add_manifest_index.rb
@@ -1,0 +1,7 @@
+class AddManifestIndex < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :manifests, :file_number, unique: true, algorithm: :concurrently
+  end
+end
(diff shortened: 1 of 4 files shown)


db/schema.rb
```diff
@@ -1,4 +1,3 @@
-# encoding: UTF-8
```
mdbenjam: First rails 5 migration so a lot of changes to the schema file.

db/migrate/20180321175607_add_manifest_index.rb
```diff
@@ -0,0 +1,23 @@
...
+  def up
+    manifest_exists_for_file_number = {}
+    manifests_to_remove = []
+
+    Manifest.find_each do |manifest|
+      if manifest_exists_for_file_number[manifest[:file_number]]
+        manifests_to_remove.push(manifest.id)
+      end
+      manifest_exists_for_file_number[manifest[:file_number]] = true
+    end
```
mdbenjam: I've run this block of code up until right before the destroy_all in prod and seen that we'll remove only one manifest (id: 464). I've run the whole migration locally with manifests to destroy and it works.

Issue #957, opened 2018-03-23
Make record version_id's index unique


mdbenjam:
We currently have a validation in Rails to ensure record's version_ids are unique. However, rails unique validations are subject to race conditions. We should recreate the version_id index to be unique so that we can guarantee only one record for each version_id will exist in the DB. This will be easier to do now while the DB is small, and should be safely able to be run outside of business hours.

PR #958, opened 2018-03-23
Adding unique constraint to version_id column

mdbenjam:
connects: #957 

Adds a unique constraint to the records table's version_id column.

### DIFF
--- a/db/migrate/20180323021917_add_unique_version_id_index_to_record.rb
+++ b/db/migrate/20180323021917_add_unique_version_id_index_to_record.rb
@@ -1,0 +1,7 @@
+class AddUniqueVersionIdIndexToRecord < ActiveRecord::Migration[5.1]
+  disable_ddl_transaction!
+
+  def change
+    add_index :records, :version_id, unique: true, algorithm: :concurrently
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180321175607) do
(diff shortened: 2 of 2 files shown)


app/models/record.rb
```diff
@@ -3,8 +3,7 @@ class Record < ApplicationRecord
 
   belongs_to :manifest_source
 
-  validates :manifest_source, :version_id, :series_id, presence: true
-  validates :version_id, uniqueness: true
+  validates :manifest_source, :version_id, :series_id, :version_id, presence: true
```
aroltsch: version_id is twice here

app/models/record.rb
```diff
@@ -68,7 +67,7 @@ def accessible_by?(user)
   end
 
   def self.create_from_external_document(manifest_source, document)
-    create(
+    create!(
```
aroltsch: create! will raise an exception that record is invalid, was it your intention?

aroltsch:
```since documents are deleted and recreated in a transaction block the old versions of the documents are not removed before the new version goes through a rails validation for uniqueness```
Nice! Can we just move the removal of the documents out of the transaction block?

app/models/record.rb
mdbenjam: awkward

app/models/record.rb
mdbenjam: Yes, if we don't raise an exception we could delete all the documents, fail to save all of our new documents, but not raise any exceptions, so we'd not rollback the transaction.

mdbenjam:
I don't think that is necessary. Now with this validation being performed at a DB level we'll still catch the uniqueness constraint. Having it in a transaction avoids us deleting but not recreating documents.

PR #962, opened 2018-03-27
Adding shoryuken gem

mdbenjam:
We need to add the Shoryuken gem before we can start running Shoryuken on our boxes.


Gemfile.lock
```diff
@@ -292,8 +292,8 @@ GEM
     rails-dom-testing (2.0.3)
       activesupport (>= 4.2.0)
       nokogiri (>= 1.6)
-    rails-html-sanitizer (1.0.3)
-      loofah (~> 2.0)
+    rails-html-sanitizer (1.0.4)
```
aroltsch: I wonder why this thing is being updated.

mdbenjam: I did it manually because `rails-html-sanitizer` had a vulnerability.

PR #969, opened 2018-03-30
Removing Sidekiq

mdbenjam:
Removing any remaining remnants of Sidekiq.


app/middleware/metrics_collector.rb
```diff
@@ -32,13 +31,4 @@ def collect_postgres_metrics
     PrometheusService.postgres_db_connections.set({ type: "dead" }, dead)
     PrometheusService.postgres_db_connections.set({ type: "idle" }, idle)
   end
-
-  def collect_background_jobs_metrics
```
mdbenjam: We should be able to do this from SQS and datadog now.

config/secrets.yml
```diff
@@ -18,7 +18,7 @@
 development:
   secret_key_base: 8c6b64e47fa8d523a9ea3d7e7658dbd5ee916003785039b14ffdd076fae644afaedc04eb683fc705c47f6f07020184c8cf3a30e79f2258521a45b861ab8e7862
   redis_url_cache: <%= ENV["REDIS_URL_CACHE"] || "redis://localhost:6379/0/cache/" %>
-  redis_url_sidekiq: <%= ENV["REDIS_URL_SIDEKIQ"] || "redis://localhost:6379" %>
+  redis_url: <%= ENV["REDIS_URL_SIDEKIQ"] || "redis://localhost:6379" %>
```
mdbenjam: We don't use this url for sidekiq anymore, so just changing it to redis_url. Probably not worth changing the ENV name.

Issue #971, opened 2018-04-01
Visiting URL for manifest user did not create returns "record not found"


lowellrex:
In the course of working on #954, I discovered the following edge case that could cause problems (and complicates the solution to #954 significantly).

Anna searches for file number 123456789, sends the URL to coworker Billie. Billie pastes the URL into their browser but is presented with a record not found message. Billie should be able to view the page without issue.

## Acceptance Criteria:
* Billie should be able to view the efolder by navigating directly to the URL

lowellrex:
Since I don't see a clear way to [cache requests to BGS cleanly](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/972#discussion_r178555301) and no users have explicitly requested this feature (and this change is not required for #954), I am going to close this PR out.

Issue #954, opened 2018-03-20
Do not display expired efolders in react app UI


lowellrex:
We consider efolder fetches to be expired (in the web UI) if they are [more than 3 days old](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/a7854aaedb9e45dc7e60f999adda949ecc734918/app/models/manifest.rb#L22). In the efolder v1 html templates UI, we currently display a technical issues page (below). If we consider an efolder that is more than 3 days old to be expired we should have a similar solution for the efolder app. I *think* we currently handle the case when somebody searches for the file number of an expired efolder properly (that is, we create a new manifest--manifest sources?--instead of presenting the expired efolder).

I think we can do better than showing a technical difficulties message for v2. Maybe we could kick off a new manifest fetch for the given efolder file number and redirect the user to that URL. 

I drafted a test to confirm v2 does not handle this properly [here](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/926#issuecomment-374721902).

### Current (v1) UI behaviour
![image](https://user-images.githubusercontent.com/32683958/37680505-7d499592-2c5a-11e8-8315-ccb14babb186.png)

## Acceptance Criteria
* Do not display an expired efolder manifest (or download results) when accessing an expired efolder directly via URL
* Confirm that we do not present an expired efolder when searching for the file number of an expired efolder

alisan16:
Closing as this hasn't been updated in 2 years

PR #972, opened 2018-04-01
Allow users to visit efolder pages for manifests they did not kick off

lowellrex:
Connects #971.

In the course of working on #954, I discovered the following edge case that could cause problems (and complicates the solution to #954 significantly).

Anna searches for file number 123456789, sends the URL to coworker Billie. Billie pastes the URL into their browser but is presented with a record not found message. Billie should be able to view the page without issue.

### DIFF
--- a/app/controllers/api/v2/manifests_controller.rb
+++ b/app/controllers/api/v2/manifests_controller.rb
@@ -8,7 +8,7 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
     return veteran_not_found(file_number) if bgs_service.fetch_veteran_info(file_number).nil?
     return sensitive_record unless bgs_service.check_sensitivity(file_number)
 
-    manifest = Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
+    manifest = Manifest.find_or_create_by(file_number: file_number)
     manifest.start!
     render json: json_manifests(manifest)
   end
@@ -16,7 +16,7 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
   def progress
     manifest = Manifest.find(params[:id])
     return record_not_found unless manifest
-    return sensitive_record unless bgs_service.check_sensitivity(file_number)
+    return sensitive_record unless bgs_service.check_sensitivity(manifest.file_number)
 
     render json: json_manifests(manifest)
   end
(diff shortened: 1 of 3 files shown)


app/controllers/api/v2/manifests_controller.rb
```diff
@@ -8,16 +8,17 @@ def start
     return veteran_not_found(file_number) if bgs_service.fetch_veteran_info(file_number).nil?
     return sensitive_record unless bgs_service.check_sensitivity(file_number)
 
-    manifest = Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
+    manifest = Manifest.find_or_create_by(file_number: file_number)
```
lowellrex: Since we are now creating the FilesDownload object when we first call it we do not to create it when we create the Manifest object. The FilesDownload object will be created as a side effect of the render function at the end of this function.
```render > json_manifests > ManifestSerializer > zip_expiration_date > Manifest.zip_expiration_date > Manifest.requested_zip_at > file_download```

File: app/models/manifest.rb [hunk 2]
```diff
@@ -102,16 +102,10 @@ def veteran
...
-    manifest = Manifest.find_or_create_by(file_number: file_number)
-    manifest.files_downloads.find_or_create_by(user: user)
-    manifest
-  end
-
   private
 
   def file_download
-    files_downloads.find_by(user: RequestStore[:current_user])
+    files_downloads.find_or_create_by(user: RequestStore[:current_user])
```
lowellrex: Previously we were creating a FilesDownload object when we started the manifest fetch. Since this PR allows users who did not start the manifest fetch to visit the URL of a given manifest, we need to make this change so we create a FilesDownload object for the user requesting this page.

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -8,16 +8,17 @@ def start
...
     manifest.start!
     render json: json_manifests(manifest)
   end
 
   def progress
-    files_download ||= FilesDownload.includes(:manifest).find_by(manifest_id: params[:id], user_id: current_user.id)
-    return record_not_found unless files_download
+    manifest = Manifest.find(params[:id])
+    return record_not_found unless manifest
+    return sensitive_record unless bgs_service.check_sensitivity(manifest.file_number)
```
mdbenjam: This concerns me, we'll greatly increase our calls to BGS.

lowellrex: Absolutely agree. We need to find some way to only make this call to BGS once per user/file. Since I don't see a clear way to do that cleanly and no users have explicitly requested this feature (and this change is not required for #954), I am going to close this PR out.

PR #974, opened 2018-04-02
Reset fetched_files_status for manifests with stale zipfiles

lowellrex:
Connects #954.


app/controllers/api/v2/manifests_controller.rb
```diff
@@ -9,6 +9,7 @@ def start
     return sensitive_record unless bgs_service.check_sensitivity(file_number)
 
     manifest = Manifest.find_or_create_by_user(user: current_user, file_number: file_number)
+    manifest.reset_stale_fetched_files_status!
```
mdbenjam: We do not need this here. We already have code to do this check. https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/manifest_source.rb#L36

app/models/manifest.rb
```diff
@@ -27,6 +27,13 @@ def start!
     vva_source.start!
   end
 
+  def reset_stale_fetched_files_status!
```
mdbenjam: The problem here is we want to check each source individually, and this checks the manifest `fetched_files_at` rather than the manifest source versions.

app/models/manifest_source.rb
```diff
@@ -20,6 +20,7 @@ class ManifestSource < ApplicationRecord
   def start!
     return if current? || pending?
     update(status: :pending)
+    manifest.update!(fetched_files_status: :initialized)
```
mdbenjam: This should probably go into `manifest.rb`

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -13,6 +13,14 @@ def start
     render json: json_manifests(manifest)
   end
 
+  def refresh
+    manifest = Manifest.find(params[:id])
```
mdbenjam: We either need a sensitivity check here, or a check to see if a FilesDownload object was made for this user. Another option is to use the start method for this route.

app/models/manifest_source.rb
lowellrex: For sure for sure. I struggled to figure out how to write this in `Manifest.start!`. The best I could come up with was something like:
```
def start!
  if !sources.empty? && sources.map { |s| s.current? || s.pending? }.all?
    update!(fetched_files_status: :initialized)
  end
  vbms_source.start!
  vva_source.start!
end
```

I figured this disobeys the idea that children shouldn't manipulate parent objects, but I thought this way was cleaner (and we don't really adhere to that idea other places anyway). I'm open to suggestions on how this could be improved or if you like this snippet above better than this.

mdbenjam: I'd rather put this in a new method in manifest like `restart!`, it could reset the `fetched_files_status` and then call start!. I worry about two different manifest sources changing their parent's `fetched_files_status`

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -13,6 +13,15 @@ def start
...
+    return record_not_found unless manifest
+    return sensitive_record unless manifest.files_downloads.find_by(user: current_user)
```
aroltsch: I think you can re-use this method to check access:
```
manifest.downloaded_by?(user)
```

app/models/manifest.rb
```diff
@@ -23,6 +23,9 @@ class Manifest < ApplicationRecord
   API_HOURS_UNTIL_EXPIRY = 3
 
   def start!
+    # Reset stale manifests.
+    update!(fetched_files_status: :initialized) if requested_zip_at && requested_zip_at < UI_HOURS_UNTIL_EXPIRY.hours.ago
```
aroltsch: Do we want to use `requested_zip_at` variable? https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/5dd3163efacb5dc9955082c9d6e1c7b52b08a6f2/app/models/manifest.rb#L120 - it is on per user basis. The variable that is on `manifest` is `fetched_files_at`.

lowellrex: I think that is fine. When we `start!()` the ManifestSources, they will either re-fetch or not based on their staleness across all users, so we should really only be redoing the step of downloading the files for zip and packaging them all up... I think.

Issue #976, opened 2018-04-02
Progress tab is selected whenever page updates


mdbenjam:
The progress tab is selected on the download page whenever the list of documents is updated. If someone clicks on `completed` we should keep that tab highlighted.

![completedtab](https://user-images.githubusercontent.com/3885236/38210975-fce8995c-3686-11e8-94de-9f0883dc972b.gif)

PR #977, opened 2018-04-03
Add manifest failed page and fix active tab

mdbenjam:
Fixes: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/976

This PR address two issues:
1) Stops the download page from always making the progress tab active whenever the list of files changes.
2) Adds a failed message when the manifest status is set to failed. This means there was some error on our end when processing the download. The user should try again or send feedback if that fails.

**Test Plan**
- [ ] Start app, search for demo case then change the manifests' fetched_files_status to failed in the rails console. Ensure the new error message displays and the retry button successfully retries the download.
- [ ] When downloading a case, ensure the progress tab starts highlighted, and when completed either the completed or if there are any errors the error tab are highlighted. While in progress is the completed tab is selected, it should not revert to the progress tab.

### DIFF
--- a/client/src/apiActions.jsx
+++ b/client/src/apiActions.jsx
@@ -26,6 +26,10 @@ const setActiveTab = (dispatch, respAttrs) => {
   const pendingRecords = respAttrs.records.some((record) => record.status === DOCUMENT_DOWNLOAD_STATE.IN_PROGRESS);
   const allPendingRecords = respAttrs.records.every((record) => record.status === DOCUMENT_DOWNLOAD_STATE.IN_PROGRESS);
 
+  // When all the records are pending, the pending tab will be selected. Once a single document is completed or
+  // failed the user can choose any tab. Once the download is complete either the completed tab (if there are no errors)
+  // or the error tab (if there are any errors) is automatically selected for the user. Then the user can switch
+  // to any tab with documents.
   if (allPendingRecords) {
     dispatch(setActiveDownloadProgressTab(IN_PROGRESS_TAB));
   } else if (!pendingRecords) {
--- a/client/src/apiActions.jsx
+++ b/client/src/apiActions.jsx
@@ -31,20 +31,11 @@ const setStateFromResponse = (dispatch, resp) => {
   dispatch(setVeteranId(respAttrs.file_number));
   dispatch(setVeteranName(`${respAttrs.veteran_first_name} ${respAttrs.veteran_last_name}`));
 
-  let activeTab = IN_PROGRESS_TAB;
-
-  switch (respAttrs.fetched_files_status) {
-  case MANIFEST_DOWNLOAD_STATE.SUCCEEDED:
-    activeTab = SUCCESS_TAB;
-    break;
-  case MANIFEST_DOWNLOAD_STATE.FAILED:
-    activeTab = ERRORS_TAB;
-    break;
-  default:
-    activeTab = IN_PROGRESS_TAB;
-    break;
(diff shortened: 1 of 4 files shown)


lib/tasks/spec_browsers.rake
```diff
@@ -4,6 +4,8 @@ begin
   namespace :spec do
     desc "Run the feature specs on supported browsers"
 
+    Rake::Task["assets:precompile"].execute
```
mdbenjam: I think this will make sure we always have an updated bundle when testing.

client/src/apiActions.jsx
```diff
@@ -14,13 +14,29 @@ import {
...
 
+const setActiveTab = (dispatch, respAttrs) => {
+  const failedRecords = respAttrs.records.some((record) => record.status === DOCUMENT_DOWNLOAD_STATE.FAILED);
+  const pendingRecords = respAttrs.records.some((record) => record.status === DOCUMENT_DOWNLOAD_STATE.IN_PROGRESS);
+  const allPendingRecords = respAttrs.records.every((record) => record.status === DOCUMENT_DOWNLOAD_STATE.IN_PROGRESS);
+
+  if (allPendingRecords) {
+    dispatch(setActiveDownloadProgressTab(IN_PROGRESS_TAB));
+  } else if (!pendingRecords) {
+    if (failedRecords) {
```
lowellrex: Might be worth adding a comment here that indicates that if there are no pending records we will activate the errors tab even if the currently active tab is the success tab.

client/src/containers/DownloadProgressContainer.jsx
```diff
@@ -233,27 +234,60 @@ class DownloadProgressContainer extends React.PureComponent {
...
 
-  render() {
-    return <React.Fragment>
-      <AppSegment filledBackground>
+  manifestFailedBanner = () => {
+    return <AlertBanner
+      title="There was an error downloading this efolder"
+      alertType="error">
+      <p>You can try to download this efolder again using the ‘Retry Download' button below.
+      If the problem persists, please send in feedback.</p>
```
lowellrex: Forgot to mention this earlier, but I think it might be helpful to add a link to the feedback page in this help text.
```
...If the problem persists, please <Link href={this.props.feedbackUrl}>send in feedback</Link>.</p>
```

client/src/apiActions.jsx
mdbenjam: Where does that happen? Sorry I'm not sure I follow that.

lowellrex: Consider the following truth table. ✅ indicates that a tab has one or more records and ❌ indicates that the tab has no records. The final column describes which tab we expect this function will activate.

Progress | Completed | Errors | Activated Tab
-------- | --------- | ------ | -------------
✅ | ❌ | ❌ | Progress
✅ | ✅ | ✅ | No change
✅ | ✅ | ❌ | No change
✅ | ❌ | ✅ | No change
❌ | ✅ | ❌ | Completed
❌ | ❌ | ✅ | Errors
❌ | ✅ | ✅ | Errors

If Beth has the Completed tab while we are in row 2 state when the final record finishes downloading, her currently active tab will change from Completed to Errors. I think this is a fine behaviour (and am not all that sure about how to avoid it... adding logic to the SET_ACTIVE_DOWNLOAD_PROGRESS_TAB case in the reducer?), but Beth might be not expect it and it may be worth a code comment.

mdbenjam: Thanks for the truth table :)

lowellrex: Love truth tables!

PR #978, opened 2018-04-04
Lowell/hacky promise flutter fix

lowellrex:
https://github.com/department-of-veterans-affairs/appeals-support/issues/1761 and https://github.com/department-of-veterans-affairs/appeals-support/issues/1762 identified an bug that caused the downloading page to flutter between several different cases. This was caused by outstanding javascript promises setting state back and forth.

This fix is not very elegant, but it does solve the problem by only allowing one outstanding promise at a time and only continuing the polling of the current promise if the manifest ID of the request it is trying to make matches the manifest ID of the outstanding promise.

### DIFF
--- a/client/src/apiActions.jsx
+++ b/client/src/apiActions.jsx
@@ -158,4 +158,4 @@ export const getDownloadHistory = (csrfToken) => (dispatch) => {
     );
 };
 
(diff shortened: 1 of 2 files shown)


lowellrex:
Closing out this PR in favor of #979.

PR #979, opened 2018-04-04
Alternative solution

mdbenjam:
https://github.com/department-of-veterans-affairs/appeals-support/issues/1762
https://github.com/department-of-veterans-affairs/appeals-support/issues/1761

When a user attempts to download multiple case files, we end up in a state where we have multiple case files polling at the same time. We then alternate updating the UI between the multiple case files. This code checks to see if this is the first call to polling, if it is, then we set the currentManifestId. If in a future poll we find that the currentManifestId is set to a new manifestId then we know a user has moved on to a new case file and we cancel this manifest's polling.

### DIFF
--- a/client/src/apiActions.jsx
+++ b/client/src/apiActions.jsx
@@ -21,6 +21,8 @@ import {
 } from './Constants';
 import { documentDownloadComplete, documentDownloadStarted, manifestFetchComplete } from './Utils';
 
+let currentManifestId = null;
+
 const setStateFromResponse = (dispatch, resp) => {
   const respAttrs = resp.body.data.attributes;
 
@@ -74,7 +76,14 @@ const buildErrorMessageFromResponse = (resp) => {
   return description;
 };
 
-export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken) => (dispatch) => {
+export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken, recursionIndex = 0) => (dispatch) => {
+
+  if (recursionIndex > 0 && currentManifestId !== manifestId) {
+    return;
+  }
+
+  currentManifestId = manifestId;
+
   getRequest(`/api/v2/manifests/${manifestId}`, csrfToken).
     then(
       (response) => { // eslint-disable-line max-statements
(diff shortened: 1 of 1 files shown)


client/src/apiActions.jsx
```diff
@@ -74,7 +76,14 @@ const buildErrorMessageFromResponse = (resp) => {
   return description;
 };
 
-export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken) => (dispatch) => {
+export const pollManifestFetchEndpoint = (retryCount, manifestId, csrfToken, recursionIndex = 0) => (dispatch) => {
+
+  if (recursionIndex > 0 && currentManifestId !== manifestId) {
```
lowellrex: I think we can reuse `retryCount` here instead of creating the new `recursionIndex` argument.

lowellrex: Also, strongly in favor of adding a code comment to explain why we are doing this (maybe just one that points back to this PR) since it is a little bizarre.

lowellrex:
This PR resolves the bug identified in department-of-veterans-affairs/appeals-support#1761 and department-of-veterans-affairs/appeals-support#1762.

This issue could crop up when Gaia searched for file number 111223333, is redirected to the search results page for that manifest, and while that manifest is still being fetched (and Gaia is seeing the spinner on the search results page) Gaia returns to the homepage (without refreshing the page, importantly) and searches for file number 444556666. When she does this, Gaia will once again be redirected to the search results, however the file number will flicker between 111223333 and 444556666 (and the associated Veteran names will flicker as well) on that page.

The cause of this bug also caused Gaia to see the search results page for files she recently searched for instead of the download progress page for any recent downloads she clicked on.

We were able to replicate this in our local test environments by increasing the timeout for demo manifest fetches (https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/428bec2cc69814da6abadcba9eeb5b4e20b44ee3/lib/fakes/document_service.rb#L65) so we could simulate the occasion when more than one manifest fetch was happening at a given time.

This bug was caused by multiple API requests changing the state that determines what information is displayed on Gaia's screen. The efolder express application makes API requests and takes actions based on responses to those requests asynchronously, which allows the application to continue doing things; drawing HTML elements on the page, manipulating data in the browser's memory, or making additional API requests.

Whenever Gaia visits the search results page or the download progress page (any page with a URL that ends in `/downloads/###`), we make an API request to determine the status of the file Gaia is trying to view. If that file is in progress (manifest fetch is incomplete and we see the spinner, or the zipfile has not finished being created and we see the progress bar) the application waits a few seconds (either 1 or 2 seconds depending on the screen) and makes the same API request again until the manifest is done being fetched or the zipfile has been created.

In addition to retrying the API request, we update the visual appearance of the webpage (by way of updating the redux state). Gaia created two API requests by searching two times in rapid succession, and since neither of those manifest fetches had completed, both of those API requests were being retried and updating the visual appearance of the webpage whenever the application received a response to either of the API requests. This resulted in the flicker of Veteran name and file number Gaia observed (and also prevented the download progress page from being displayed).

We fixed this bug by allowing a maximum of one single API request to be outstanding at any given time.

PR #983, opened 2018-04-12
Dealing with VVA's non-unique document ids

mdbenjam:
This is to deal (hopefully once and for all) with the duplicate document ids we receive from VVA. We will stop excluding documents. Instead, we'll download them all to eX. However, we change around our uniqueness constraints. This PR does three things:

1) Remove the unique index on version_id. Instead we replace that with a uniqueness constraint on [version_id, manifest_source_id]. This means we can only have a document appear once per manifest source. It also means we can still have fast lookups keying in with either `version_id` or `version_id and manifest_source_id`. 
2) We remove the [manifest_source_id, version_id] index since it should be superseded by the above index.
3) We add logic in manifest_fetcher to remove duplicates that appear in a single manifest source.

**Test Plan**
- [ ] Automated tests to ensure we filter out duplicates within a manifest source.

### DIFF
--- a/app/services/manifest_fetcher.rb
+++ b/app/services/manifest_fetcher.rb
@@ -5,14 +5,23 @@ class ManifestFetcher
 
   EXCEPTIONS = [VBMS::ClientError, VVA::ClientError].freeze
 
-  # VVA has a bug where they return these ids in multiple veteran's eFolders. This violates our unique constraint
-  # and since documents should be unique to veterans, we have decided to filter out these documents.
-  VVA_DUPLICATE_ID_LIST = %w[{F291C1BC-FCDE-4C58-8544-B4FCE2E59008} {4F277E03-66ED-4263-9A43-075CFFF72393}].freeze
+  # We've found for some veteran eFolders, VVA returns the same document multiple times. We filter those out here.
+  def remove_duplicates(documents)
+    list_of_ids = {}
+
+    documents.select do |doc|
+      if !list_of_ids[doc.document_id]
+        list_of_ids[doc.document_id] = true
+        true
+      else
+        false
+      end
+    end
+  end
 
   def process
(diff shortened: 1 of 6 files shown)


app/services/manifest_fetcher.rb
```diff
@@ -5,14 +5,23 @@ class ManifestFetcher
...
+        list_of_ids[doc.document_id] = true
+        true
+      else
+        false
+      end
     end
+  end
+
+  def process
+    documents = remove_duplicates(manifest_source.service.v2_fetch_documents_for(manifest_source))
```
aroltsch: I would think we want to handle this action in the DocumentCreator as DocumentCreator also filters out restricted documents. I think ManifestFetcher should only worry about fetching documents.

app/services/manifest_fetcher.rb
```diff
@@ -5,14 +5,23 @@ class ManifestFetcher
...
-  # and since documents should be unique to veterans, we have decided to filter out these documents.
-  VVA_DUPLICATE_ID_LIST = %w[{F291C1BC-FCDE-4C58-8544-B4FCE2E59008} {4F277E03-66ED-4263-9A43-075CFFF72393}].freeze
+  # We've found for some veteran eFolders, VVA returns the same document multiple times. We filter those out here.
+  def remove_duplicates(documents)
+    list_of_ids = {}
 
-  def process
-    documents = manifest_source.service.v2_fetch_documents_for(manifest_source).reject do |document|
-      VVA_DUPLICATE_ID_LIST.include?(document.document_id)
+    documents.select do |doc|
```
aroltsch: Could you use `uniq` with a block to filter out duplicate document IDs?

```
a = [ OpenStruct.new(id: 1), OpenStruct.new(id: 1) ]
=> [#<OpenStruct id=1>, #<OpenStruct id=1>]
[5] pry(main)> a.uniq { |i| i.id }
=> [#<OpenStruct id=1>]
```

Issue #975, opened 2018-04-02
Reduce the time of the visibility timeout and lock for the package files job


mdbenjam:
Since deploys kill all running jobs, our package files job poses a problem. It can take a long time to run, so we've set a long visibility timeout. However, this means that if the job is killed it won't be retried for 12 hours. This is a really bad user experience. We could retry more often, but having multiple of these jobs running seems wasteful (but maybe not really that bad). There are two possible options for us.

1) Reduce the visibility timeout to only a few hours.
1) Spin up lots of jobs to cache the files in S3, and then run one final job to build the zip file. This allows us to restart any one of the jobs which should be quick rather than one long job which is slow.

alisan16:
Closing as this hasn't been updated in 2 years

PR #986, opened 2018-04-13
Release message after SIGTERM

mdbenjam:
Connects: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/975

Lots of interesting things to report about Ruby, SQS and Shoryuken.

1) Ensure blocks are run even when a process is shutting down. When shoryuken receives a SIGTERM it goes into a shutdown cycle. It somehow terminates any running jobs (I can't figure out how since I cannot seem to catch a relevant exception). But since ensure blocks still run, the locks we acquire on each job start are released.
2) Setting visibility timeout of a message to 0 allows another worker to pick it up immediately.
3) Shoryuken allows you to write middleware where you can modify the SQS visibility timeout.

Putting these altogether gives us the ability to use ensure blocks to make visibility timeouts 0 when Shoryuken is shutting down.

I've also reduced the timeout on our package files job lock. Over the past 10 days the longest we've waited for a job is 5,850 seconds. The average is around 1,000 seconds. So our current lock time of 12 hours seems too long. I've moved it down to a still very comfortable 6 hours (21,200 seconds).

Data can be found here: https://app.datadoghq.com/dash/422343/efolder-express-overview?live=true&page=0&is_auto=false&from_ts=1521065479628&to_ts=1523657479628&tile_size=m&fullscreen=false

Note I'm not actually sure all of this is a good idea, so I welcome y'alls thoughts on it!

### DIFF
--- a/app/jobs/middleware/job_sigterm_middleware.rb
+++ b/app/jobs/middleware/job_sigterm_middleware.rb
@@ -1,0 +1,16 @@
+class JobSigtermMiddleware
+  # When Shoryuken receives a call to shutdown (i.e. SIGTERM), it stops jobs that
+  # are not complete. Since some of our jobs have long visibility timeouts they
+  # may not be restarted for a long time. This gets around that problem by
+  # forcing the visibility timeout on a job that did not complete to be 0, as is
+  # suggested here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html#terminating-message-visibility-timeout
+  def call(_worker, _queue, msg, _body)
+    job_finished = false
+    yield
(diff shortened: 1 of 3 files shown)


sunil-sadasivan:
Awesome Mark!  Agreed with Anya, is there a way to validate this, send a sigterm?  Are you certain the job is able to finish the execution of setting the visibility timeout to 0?

mdbenjam:
I tested locally against localstack. Started a job, sent a sigterm, started up shoryuken again, the job started running again. Without this code, the job did not start running again. Not sure if there is some way to do automated testing with this. It's probably not easy since we'd likely need to use localstack in testing on Travis. Could also test locally against the real SQS if that seems desirable.

aroltsch:
Ha! Very nice.

Issue #987, opened 2018-04-18
`/api/v2/records/:uuid` is generating a massive number of data series in Prometheus


askldjd:
The root cause is https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1377. Since this feature was turned on, Prometheus has been crashing. At its peak, we observed over 250k series per `/metrics` API call. Since Prometheus polls `/metrics` every 15s, it was not able handle the load.

To fix this, we need to implement a label collapser in eX similar to one in Caseflow.
https://github.com/department-of-veterans-affairs/caseflow/blob/master/config.ru#L25

Here's a sample `/metric` output from efolder server.
```
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.005"} 1.0
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.01"} 1.0
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.025"} 1.0
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.05"} 1.0
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.1"} 1.0
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.25"} 1.0
http_server_request_duration_seconds_bucket{code="200",method="options",host="efolder.cf.ds.va.gov",path="/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30",le="0.5"} 1.0
[...]

PR #988, opened 2018-04-18
Update prometheus url collapser regex to include UUID pattern

lowellrex:
Connects #987.

cc @lisac-usds, @alan-ning-usds

### DIFF
--- a/config.ru
+++ b/config.ru
@@ -24,17 +24,24 @@ use Rack::Deflater,
 # Customized collector for our own metrics
 use MetricsCollector
 
+# Replace ids and id-like values to keep cardinality low.
+# Otherwise Prometheus crashes on 400k+ data series.
+# '/users/1234/comments' -> '/users/:id/comments'
+# '/hearings/dockets/2017-10-15' -> '/hearings/dockets/:id'
+# '/certifications/new/123C' -> '/certifications/new/:id'
+# '/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30' -> '/api/v2/records/:id'
+numeric_id_pattern = '\d+'
+date_pattern = '\d{4}-\d{2}-\d{2}'
+vbms_id_pattern = '\d+[CS]?'
+uuid_pattern = '[A-F0-9]{8}-[A-F0-9]{4}-[A-F0-9]{4}-[A-F0-9]{4}-[A-F0-9]{12}'
+id_matching_regex = %r'/(?:#{numeric_id_pattern}|#{date_pattern}|#{vbms_id_pattern}|#{uuid_pattern})(/|$)'
+
 label_builder = lambda do |env, code|
   {
     code: code,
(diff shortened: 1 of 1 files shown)


config.ru
```diff
@@ -24,17 +24,24 @@ use Rack::Deflater,
...
+# Replace ids and id-like values to keep cardinality low.
+# Otherwise Prometheus crashes on 400k+ data series.
+# '/users/1234/comments' -> '/users/:id/comments'
+# '/hearings/dockets/2017-10-15' -> '/hearings/dockets/:id'
+# '/certifications/new/123C' -> '/certifications/new/:id'
+# '/api/v2/records/7D11AE16-1DFF-49F5-AE31-B9E4675EBC30' -> '/api/v2/records/:id'
+numeric_id_pattern = '\d+'
+date_pattern = '\d{4}-\d{2}-\d{2}'
+vbms_id_pattern = '\d+[CS]?'
+uuid_pattern = '[A-F0-9]{8}-[A-F0-9]{4}-[A-F0-9]{4}-[A-F0-9]{4}-[A-F0-9]{12}'
```
amprokop: How sure are we that this will catch all vbms document ids?

lowellrex: Can you think of examples it wouldn't catch?

lowellrex: Also, looking at the routes, I don't think we pass VBMS IDs as a parameter of the path anywhere (and we probably shouldn't because those VBMS IDs can contain PII). Though this is probably not the time to be making this pattern more restrictive...

amprokop: I'm not too familiar with vbms document ids,  @aroltsch is a better person to double check. I'm seeking some level of assuredness that:

-document ids always 5 alphanumeric groups
-the groups are 8 chars then 4 chars then 4 then 4 then 12
etc

Not sure if you were just going off of that one example or if you have broader knowledge, sorry to not be clear on that

lisac-usds:
accounting for the uuid pattern here should have a HUGE benefit to Prometheus. And thank you @lowellrex for breaking down the regex into less-scary variables! 

One way forward: we could deploy this version and re-check the flow through `/metric` - that might give us a sense of whether there are additional patterns to incorporate (like other document ids).

config.ru
lowellrex: After chatting with Anya, I'm pretty sure that the document ID will always be a [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier). Also, I'll go ahead and get rid of the date and (what I incorrectly called) VBMS ID patterns since we don't use those anywhere for efolder.

aroltsch: I think this regex for uuid_pattern looks good.

lowellrex:
Looks like this is working in my local environment (see image below).
image

@lisac-usds does this look good to you with regards URL collapsing through the push gateway?

amprokop:
Looks good to me, I'd like @lisac-usds to make the final call though 👍

lisac-usds:
Looks good to me! Thank you!!

Issue #995, opened 2018-04-20
Bugs with new SQS setup


mdbenjam:
We've been finding that some manifest downloads are getting stuck in a "pending" state. This means users cannot retry the download, nor will it ever finish. I've been investigating why this happens and present my findings below.

We're pretty sure the only time this happens is when an EC2 instance is stopped. This can happen during a deploy or when AWS decides to rotate our boxes. Ideally to solve this issue we can do two things:

1) Reduce the visibility timeout. This means jobs will get picked up quickly if another server is killed that was processing the job.
1) When a server dies, send a message to SQS to re-enqueue the job.

Problems with reducing the visibility timeout: While we can enqueue the same job multiple times, we like to try to avoid it. This is because if the job still has the lock, then a second thread will end up waiting until the first job has finished before it can proceed (and then will note the first job already completed the work, and will then exit immediately). If the first job has given up the lock, then the second thread will try to create the same zip file at the same time which may make the UI look weird, and is wasteful. We'd like to set this value higher than we expect any job to take. The job usually takes less than 1 hour, but sometimes takes an order of magnitude longer:
<img width="1243" alt="screen shot 2018-04-20 at 6 22 17 pm" src="https://user-images.githubusercontent.com/3885236/39076178-cf31d132-44c7-11e8-9390-9af37f460cc8.png">

It's unclear why this is happening. We should investigate to figure it out. Is it just larger zip files? Or is something more going on. We also don't have very good log tagging for Shoryuken logs, [so the first step is to improve that](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/994).

[...]

PR #998, opened 2018-04-23
Move lock from job to model

lowellrex:
Connects #995.

First half of the fix for #995 here. The ultimate fix is to offload all of the queue handling to SQS by way of Shoryuken and make extensive use of [SQS ChangeMessageVisibility](https://github.com/phstc/shoryuken/wiki/Worker-options#auto_visibility_timeout) to ensure that the job is always in process or soon to be made available on the queue.

In order to allow for that approach, we need to have jobs execute when we fire them off. The Redis/Semaphore lock could prevent that from happening (if an EC2 instance is killed while the job is running we will try to run it again when SQS makes the message available, but will run into the lock once we start executing the job). Since the lock exists to prevent us from doing the work in the job multiple times, we can move that lock out of the job and instead use the lock to make sure we don't create more than one job for the same manifest.

### DIFF
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -22,6 +22,8 @@ class Manifest < ApplicationRecord
   UI_HOURS_UNTIL_EXPIRY = 72
   API_HOURS_UNTIL_EXPIRY = 3
 
+  SECONDS_TO_AUTO_UNLOCK = 5
+
   def start!
     # Reset stale manifests.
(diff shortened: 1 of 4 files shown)


app/models/manifest.rb
```diff
@@ -31,8 +31,12 @@ def start!
   end
 
   def download_and_package_files!
-    return if pending?
-    update(fetched_files_status: :pending)
+    s = Redis::Semaphore.new("package_files_#{id}".to_s, url: Rails.application.secrets.redis_url)
```
aroltsch: We still need an expiration to expire keys and auto-unlock timeout. We will also need an ensure block I believe.

PR #1003, opened 2018-05-02
 conditional connect_vbms depending on env

emvalenzuela:
## Description

Made a condition to depending on the environment install a different gem to connect to VBMS
## Acceptance Criteria
it installs the correct GEM or not.


emvalenzuela:
Going in a different direction, closing this PR

Issue #1019, opened 2018-06-15
Handle Prometheus unavailability 


nemodjuric:
We had an outage of Prometheus which caused outage of eFolder since pushgetaway was not available. 
https://github.com/department-of-veterans-affairs/appeals-deployment/pull/1508/files

We should catch the failure when sending metrics.

https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/jobs/middleware/job_prometheus_metric_middleware.rb#L13

https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/jobs/middleware/job_data_dog_metric_middleware.rb

RobertAlexanderUSDS:
@nemo I think this is complete. Can this be closed?

nemo:
@RobertAlexanderUSDS I'd say it looks good to me, but you misspelt @nemodjuric's handle ;)

PR #1021, opened 2018-06-18
Rescue when sending metrics to Prometheus and DataDog

nemodjuric:
Connects to #1019


app/jobs/middleware/job_prometheus_metric_middleware.rb
```diff
@@ -1,18 +1,25 @@
 class JobPrometheusMetricMiddleware
-  def call(_worker, _queue, _msg, body)
-    job_class = body["job_class"]
+  def call(_worker, queue, _msg, body)
+    @job_class = body["job_class"]
```
mdbenjam: I think it would be more standard to just pass job_class in our call to `record_and_push_metrics` rather than making it an instance variable.

Issue #1029, opened 2018-07-16
When eFolder is out of service, logins should still succeed


amprokop:
Caseflow depends on eFolder for logins. Right now, if we take eFolder down, it means we can't authenticate.

# AC
- After taking eFolder out of service, Caseflow users should still be able to successfully authenticate

alisan16:
Closing as this hasn't been updated in 2 years

PR #1030, opened 2018-07-16
Doesn't prevent login when app is out of service

amprokop:
Connects https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/1029

# Testing
- Check out the master branch and start the app connected to Redis as the cache 
- Log out
- Run `Rails.cache.write("out_of_service", true)`
- Attempt to log in
- Log in will fail
- Check out this branch
- Attempt to log in
- User will be successfully logged in, and "Technical Difficulties" page will be displayed

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ In a separate terminal, watch for webpack changes:
 
 > $ cd client && yarn run build --watch
 
-And in another seperate terminal, start a jobs worker
+And in another separate terminal, start a jobs worker
 
 > $ bundle exec shoryuken start -q efolder_development_high_priority efolder_development_low_priority efolder_development_med_priority -R
 
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Now start both the rails server,
 
 > $ rails s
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -68,7 +68,11 @@ Now start both the rails server,
 
 > $ rails s
 
-And in a seperate terminal, start a jobs worker
+In a separate terminal, watch for webpack changes:
+
+> $ cd client && yarn run build --watch
```
amprokop: this command is necessary to create a webpack bundle, we could consolidate the rails & yarn commands at some point but the README was incomplete

config/environments/development.rb
```diff
@@ -16,14 +16,11 @@
   if Rails.root.join('tmp/caching-dev.txt').exist?
     config.action_controller.perform_caching = true
 
-    config.cache_store = :memory_store
```
amprokop: Without these two lines, we'll fall back to the application.rb config in dev mode, which properly uses the Redis instance we have running in a Docker container.

PR #1031, opened 2018-07-16
Forking access check

mdbenjam:
Connects https://github.com/department-of-veterans-affairs/caseflow/issues/6243

### Description
Added a temporary fork to let us try out this new access control method.

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -24,11 +24,13 @@ class ExternalApi::BGSService
   end
 
   def check_sensitivity(file_number)
+    current_user = RequestStore[:current_user]
+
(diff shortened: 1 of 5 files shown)


app/services/external_api/bgs_service.rb
```diff
@@ -24,11 +24,13 @@ def fetch_veteran_info(file_number)
   end
 
   def check_sensitivity(file_number)
+    current_user = RequestStore[:current_user]
+
     @bgs_client ||= init_client
-    MetricsService.record("BGS: can_access? (find_flashes): #{file_number}",
```
shanear: same as with the efolder. The string could also be toggled.

PR #1032, opened 2018-07-17
Optimize queries around record counts

aroltsch:
Closes #6239

I refactored counting of documents to perform in Ruby instead of SQL. 

Before:

image

After:

image

### DIFF
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -71,7 +71,7 @@ class Manifest < ApplicationRecord
   end
 
   def number_successful_documents
-    records.success.count
+    records.select(&:success?).size
   end
 
   def time_to_complete
@@ -88,7 +88,7 @@ class Manifest < ApplicationRecord
   end
 
   def number_failed_documents
-    records.failed.count
+    records.select(&:failed?).size
   end
 
   def s3_filename
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -79,7 +79,7 @@ class Manifest < ApplicationRecord
   end
 
   def seconds_left
-    records.initialized.count * Record::AVERAGE_DOWNLOAD_TIME_IN_SECONDS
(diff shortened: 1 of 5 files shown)


app/models/manifest.rb
```diff
@@ -79,7 +79,7 @@ def time_to_complete
   end
 
   def seconds_left
-    records.initialized.count * Record::AVERAGE_DOWNLOAD_TIME_IN_SECONDS
+    records.select(&:initialized?).count * Record::AVERAGE_DOWNLOAD_TIME_IN_SECONDS
```
aroltsch: Count using Ruby instead of using SQL

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -20,7 +24,7 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
         source: source.name,
         status: source.status,
         fetched_at: source.fetched_at,
-        number_of_documents: source.records.count
+        number_of_documents: source.records.size
```
aroltsch: This will count in Ruby instead of SQL

app/models/serializers/v2/manifest_serializer.rb
```diff
@@ -8,8 +8,12 @@ class Serializers::V2::ManifestSerializer < ActiveModel::Serializer
   attribute :updated_at
   attribute :fetched_files_at
   attribute :fetched_files_status
-  attribute :number_successful_documents
-  attribute :number_failed_documents
+  attribute :number_successful_documents do
+    object.records.select(&:success?).size
```
mdbenjam: why not move this into the definition of the `number_successful_documents` and `number_failed_documents` methods instead of leaving them in the serializer?

aroltsch: egggsellent point! I forgot to move it there.

PR #1035, opened 2018-08-02
Changing veteran endpoint

mdbenjam:
Looks like some veterans aren't in the people service??? We had a report that a user couldn't get to reader for a veteran's file number. When we tried it we saw BGS was saying the veteran didn't exist when hitting the people endpoint, but does exist on the veteran endpoint. I think we should switch to the veteran endpoint, especially since we now use that same endpoint for access control.


app/services/external_api/bgs_service.rb
```diff
@@ -18,7 +18,7 @@ def fetch_veteran_info(file_number)
       MetricsService.record("BGS: fetch veteran info for vbms id: #{file_number}",
                             service: :bgs,
                             name: "veteran.find_by_file_number") do
-        @bgs_client.people.find_by_file_number(file_number)
+        @bgs_client.veteran.find_by_file_number(file_number)
```
aroltsch: Interesting!

Issue #1042, opened 2018-10-04
Remove unnecessary api call for recent downloads


nemodjuric:
We are making unnecessary call to get recent downloads as part of initial react data (and do nothing with it), then RecentDownloadsContainer does the same thing. 

<img width="1428" alt="screen shot 2018-10-04 at 9 57 29 am" src="https://user-images.githubusercontent.com/4960030/46478980-0c5cfd80-c7bc-11e8-9920-db60824b0f81.png">

alisan16:
Closing as this hasn't been updated in 2 years

PR #1048, opened 2018-11-02
Remove recentDownloads from initial react data

nemodjuric:
Connected to #1042 

To test:
1. Run app in dev environment
2. Search for 'DEMO1' and download file
3. Restart the server, clear the cookies and login as the same user from steps 1 and 2
4. Make sure recent download shows on home page
5. Make sure server log does not show loading of FilesDownloads as part of DownloadsController#new step 

screen shot 2018-11-02 at 1 21 02 pm

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -48,7 +48,6 @@ class ApplicationController < BaseController
       dropdownUrls: dropdown_urls,
       efolderAccessImagePath: ActionController::Base.helpers.image_path("help/efolder-access.png"),
       feedbackUrl: feedback_url,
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -48,7 +48,6 @@ def initial_react_data
       dropdownUrls: dropdown_urls,
       efolderAccessImagePath: ActionController::Base.helpers.image_path("help/efolder-access.png"),
       feedbackUrl: feedback_url,
-      recentDownloads: ActiveModelSerializers::SerializableResource.new(current_user.recent_downloads, each_serializer: Serializers::V2::HistorySerializer),
```
nemodjuric: This line was causing time outs, which we fixed here #https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1046. However, it turns out that this line was unnecessary at all. This issue made me realize that, can't see the forest for the trees.

Issue #1052, opened 2018-11-19
Use C file number if SSN doesn't have files


mdbenjam:
Users sometimes requests a Veteran's eFolder with their SSN when the file is actually stored under the C-File-Number. We should add code like this: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/services/external_api/vbms_service.rb#L43 to eFolder so that we can try the C-File-Number if the SSN doesn't work.

Open question: What should we do with VVA? If someone provides the SSN should we also get what's under the C-File-Number?

nemodjuric:
This will cause duplicate records, but I created separate issue for this, as I think we have to answer at some question first before fixing it. We can tackle one issue at the time, it will make things easier to track.

PR #1053, opened 2018-11-20
Search for C file number when SSN has no files 

nemodjuric:
Connects to #1052


Gemfile.lock
```diff
@@ -268,7 +268,7 @@ GEM
     public_suffix (3.0.2)
     puma (3.11.3)
     quantile (0.2.0)
-    rack (2.0.5)
+    rack (2.0.6)
```
nemodjuric: There was a security failure with rack (2.0.5)

Issue #1059, opened 2018-11-23
Fix BGS service response when there is no :ssn element 


nemodjuric:
BGS response has VetBirlsRecod and vetCorpRecord. If latter one is missing, :ssn will be nil
Not sure why do we read from vetCorpRecord, but not VetBirlsRecod record. However we should have a backup  in case one is missing. VetBirlsRecod record has :soc_sec_number that should correspond to :ssn in vetCorpRecord

We currently have 1804 records with empty veteran_last_four_ssn field. We used this for package names, so there will be packages that are missing veteran_last_four_ssn in their name.

Issue #1058, opened 2018-11-23
Fix scenario when no record is found


nemodjuric:
Currently, there is a bug where we don't check correctly if there is no record found in BGS. 

Here:
https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/controllers/api/v2/manifests_controller.rb#L8

Inappropriate use here as well:
https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/veteran.rb#L33

As a result, we have number of Manifest records that were never fetched and have empty veteran info  (1248 at the moment). Also, this causes unnecessary search and calls to dependencies. 
We should check if record exists, if it doesn't we should not proceed. 

Consider using find_by_file_number from People service (vs Veteran service) in ruby-bgs.

Issue #1057, opened 2018-11-23
Standardize what is used for file_number in Manifests creation


nemodjuric:
See #1052 for context. 

Once we fix #1052, we'll have double records in our database since user can search both by C-file number and SSN. 
We should have a standard way what we use for file_number in Manifest creation. Either claim_number or SSN. 
The impression is that we should always use vbms_id as Manifest file_number. 
I believe that vbms_is is the same as claim_number from BGS which is same as slogid in CORRES in VACOLS.

mdbenjam:
Yeah maybe we make the call to BGS first before anything to get the claim number and then always use that. Regardless of what they've entered.

pkarman:
This is how we approach this problem in caseflow: https://github.com/department-of-veterans-affairs/caseflow/blob/master/app/controllers/intakes_controller.rb#L97

PR #1060, opened 2018-12-11
Always use claim number as file_number... and other fixes

nemodjuric:
Connects to #1057 
Connects to #1058 
Connects to #1059 

To test this would be the best either in UAT or locally in staging environment while hooked to production endpoints. 
- Run search by both C file number and S file number and make sure both cases have same results and that there is only one manifest stored and that it has C number as file number.
- Run search for non-existent file number and make sure no Manifest is stored for this search and no-record-found error message was displayed
- Run search for file that has sensitive files and make sure Sensitive files message is displayed. 

Notes:
During work on this PR below is what I experienced:
BGS VeteranService find_by_file_number returns file for both C and S number in most cases I tested, I ran into only one case where it didn't return for S number. 
(VBMS FindDocumentVersionReference seems to do it only for vbms_id. VVA get document list accepts both numbers similar to BGS)
If BGS response for veteran.find_by_file_number is nil that means there was some kind of error (error is raised) including case when file contains sensitive record. 
Whether record exists or not, it is stored in return_message field,(it should have "BPNQ0301" if there is one, "No BIRLS record found” if there's none). This is different behavior than BGS PeopleService find_by_file_number which returned nil instead if there was no record.


app/services/external_api/bgs_service.rb
```diff
@@ -3,27 +3,26 @@
 # Thin interface to all things BGS
 class ExternalApi::BGSService
   def parse_veteran_info(veteran_data)
-    ssn = veteran_data[:ssn]
+    ssn = veteran_data[:ssn] ? veteran_data[:ssn] : veteran_data[:soc_sec_number]
```
nemodjuric: This gives an option to read from VetBirlsRecod in case vetCorpRecord has no ssn.

app/services/external_api/vbms_service.rb
```diff
@@ -38,7 +38,7 @@ def self.v2_fetch_documents_for(source)
     rescue VBMS::HTTPError => e
       raise unless e.body.include?("File Number does not exist within the system.")
 
-      alternative_file_number = ExternalApi::BGSService.new.fetch_veteran_data(veteran_file_number)[:claim_number]
+      alternative_file_number = ExternalApi::BGSService.new.fetch_veteran_info(veteran_file_number)["file_number"]
```
nemodjuric: We probably don't need this anymore, since we always store under C number now. However, we can consider an option searching by S number in VBMS if C number is not available. Not sure if VBMS has vbms_ids that are S numbers and different than C numbers though.

spec/services/bgs_service_spec.rb
```diff
@@ -3,33 +3,34 @@
 
   context "#parse_veteran_info" do
     before do
-      @veteran_info = {
+      @veteran_data = {
```
nemodjuric: Changed this to clear the confusion - veteran_data is used to refer to bgs response throughout the code, whether veteran_info refers to parsed response.

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -1,15 +1,27 @@
...
+# rubocop:disable Metrics/CyclomaticComplexity
+# rubocop:disable Lint/RescueException
 class Api::V2::ManifestsController < Api::V1::ApplicationController
   def start
     file_number = request.headers["HTTP_FILE_NUMBER"]
     return missing_header("File Number") unless file_number
 
     return invalid_file_number unless bgs_service.valid_file_number?(file_number)
-    return veteran_not_found(file_number) if bgs_service.fetch_veteran_info(file_number).nil?
-    return sensitive_record unless bgs_service.check_sensitivity(file_number)
```
nemodjuric: No need for check_sensitivity call since find_by_file_number does it for us (by throwing specific error)

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -1,15 +1,27 @@
...
 
-    manifest = Manifest.includes(:sources, :records)
-                       .find_or_create_by_user(user: current_user, file_number: file_number)
+    begin
+      veteran_info = bgs_service.fetch_veteran_info(file_number)
+
+    # Exception instead of StandardError since we reraise the error
```
mdbenjam: I think rescuing Exception is a bad practice, I guess it's fine if we're re-raising it, but can we just catch StandardError instead. Presumably the sensitive error is a descendent of it? Maybe I'm wrong 🤷‍♂️

mdbenjam: Still re-raise it. Just if it derives from StandardError, might as well scope the rescue statement to that. i.e.

```
rescue StandardError => e
      return sensitive_record if e.message.include?("Sensitive File - Access Violation")
      raise e
    end
```

Issue #1067, opened 2019-03-01
Fix ActiveRecord::RecordNotUnique bug on manifests table


lowellrex:
Related to [this epic in the main Caseflow repo](https://github.com/department-of-veterans-affairs/caseflow/issues/9740).

[Caseflow is returning errors](https://sentry.ds.va.gov/department-of-veterans-affairs/caseflow/issues/3899/) from the `/api/v2/manifests` efolder endpoint due to a [unique key constraint violation](https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/3155/) for the `manifests` table.

## Acceptance criteria
* [ ] Determine why we are encountering this error.
* [ ] Gracefully handle this error.

PR #1068, opened 2019-03-06
Fix not unique error on find_or_create manifest and retry

nemodjuric:
Connects to #1067 

This should fix race condition when using find_or_create. More context here: https://api.rubyonrails.org/classes/ActiveRecord/Relation.html#method-i-find_or_create_by


db/migrate/20190307025848_remove_index_files_downloads_on_manifest_id_and_user_id.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveIndexFilesDownloadsOnManifestIdAndUserId < ActiveRecord::Migration[5.1]
+  def change
+    remove_index :files_downloads, ["manifest_id", "user_id"]
+  end
+end
```
lowellrex: Do these database changes have anything to do with this change in the `Manifest` class?

nemodjuric: This is constraint against duplicate records for FilesDownloads.

PR #1080, opened 2019-04-24
Add BGSError class

nemodjuric:
Adds new class BGSError which creates new exception classes based on existing BGS errors.


app/services/external_api/bgs_service.rb
```diff
@@ -23,6 +23,8 @@ def fetch_veteran_info(file_number)
         @bgs_client.veteran.find_by_file_number(file_number)
       end
     parse_veteran_info(veteran_data) if veteran_data
+  rescue StandardError => e
```
pkarman: is this the only place we want to catch these?

nemodjuric: That's the only BGS call we make, out of two. For the other one, we started reading into the error message instead of calling check_sensitivity/can_access?

PR #1093, opened 2019-05-10
Add DocumentCounter service and controller endpoint

pkarman:
**Why**: Caseflow Queue needs only a document count for display purposes. We should not have to wait for the entire manifest's worth of documents to be downloaded and processed if all we need is the count that will eventually be downloaded.

This API endpoint should provide a very fast count of what the manifest will eventually contain.

### DIFF
--- a/app/controllers/api/v2/manifests_controller.rb
+++ b/app/controllers/api/v2/manifests_controller.rb
@@ -1,24 +1,8 @@
 # TODO: create Api::V2::ApplicationController
-# rubocop:disable Metrics/CyclomaticComplexity
-# rubocop:disable Metrics/PerceivedComplexity
 class Api::V2::ManifestsController < Api::V1::ApplicationController
   def start
-    file_number = request.headers["HTTP_FILE_NUMBER"]
-    return missing_header("File Number") unless file_number
-
-    return invalid_file_number unless bgs_service.valid_file_number?(file_number)
-
-    begin
-      veteran_info = bgs_service.fetch_veteran_info(file_number)
-    rescue StandardError => e
-      return sensitive_record if e.message.include?("Sensitive File - Access Violation")
-      return vso_denied_record if e.message.include?("Power of Attorney of Folder is")
-      raise e
-    end
-
-    return veteran_not_found(file_number) unless bgs_service.record_found?(veteran_info)
-
-    file_number = veteran_info["file_number"] if veteran_info["file_number"]
+    file_number = verify_veteran_file_number
+    return if performed?
 
     manifest = Manifest.includes(:sources, :records).find_or_create_by_user(user: current_user, file_number: file_number)
     manifest.start!
@@ -53,10 +37,10 @@ class Api::V2::ManifestsController < Api::V1::ApplicationController
     doc_count = Rails.cache.fetch(cache_key) do
       doc_counter = DocumentCounter.new(manifest: Manifest.find(manifest_id))
       doc_counter.count
-    rescue ActiveRecord::RecordNotFound
-      return record_not_found
     end
     render json: { documents: doc_count }
+  rescue ActiveRecord::RecordNotFound
+    return record_not_found
   end
 
   private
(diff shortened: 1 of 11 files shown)


app/controllers/api/v2/manifests_controller.rb
```diff
@@ -1,24 +1,8 @@
...
-      return sensitive_record if e.message.include?("Sensitive File - Access Violation")
-      return vso_denied_record if e.message.include?("Power of Attorney of Folder is")
-      raise e
-    end
-
-    return veteran_not_found(file_number) unless bgs_service.record_found?(veteran_info)
-
-    file_number = veteran_info["file_number"] if veteran_info["file_number"]
+    file_number = verify_veteran_file_number
+    return if performed?
```
lowellrex: Where do we define the `performed?` function?

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -75,6 +71,27 @@ def vso_denied_record
   def invalid_file_number
     render json: { status: "File number is invalid. Veteran IDs must be 8 or more characters and contain only numbers." }, status: 400
   end
+
+  def verify_veteran_file_number
+    file_number = request.headers["HTTP_FILE_NUMBER"]
+    return missing_header("File Number") unless file_number
```
lowellrex: I think we're returning `render`s to the calling code and will trigger double render errors if we return early here.

app/services/document_counter.rb
```diff
@@ -0,0 +1,14 @@
...
+  attr_accessor :manifest
+
+  def count
+    total = 0
+    manifest.sources.each do |source|
+      documents = ManifestFetcher.new(manifest_source: source).fetch_documents
+      total += DocumentCreator.new(external_documents: documents).external_documents.count
+    end
+    total
+  end
```
lowellrex: Can we consolidate this [and `V2::DownloadManifestJob`](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/ad7dfe8f793a4a0e5b3e09c8a4dd11c7e45a8698/app/jobs/v2/download_manifest_job.rb#L8)? It looks like they do pretty similar things (though `V2::DownloadManifestJob` triggers an extra job we don't want to wait for).

app/controllers/api/v2/manifests_controller.rb
pkarman: it's a Rails controller method.

https://www.rubydoc.info/docs/rails/4.1.7/ActionController/Metal#performed%3F-instance_method

https://stackoverflow.com/questions/26446375/is-it-possible-in-rails-to-check-whether-a-redirect-or-render-had-already-been-i

app/controllers/api/v2/manifests_controller.rb
pkarman: see `performed?`

app/controllers/api/v2/manifests_controller.rb
lowellrex: Makes sense! Thanks

app/services/document_counter.rb
pkarman: The point of adding a `ManifestFetcher.fetch_documents` method was to *avoid* consolidating them, because of that extra (long-running) job that `DownloadManifestJob` performs. We don't want to download the manifest documents, just the manifest itself.

Are you worried about the additional upstream calls? Those seem pretty cheap (1-3s in my testing) compared to the fetch of the documents themselves (minutes).

Ideally we could split out the filtering that `DocumentCreator` performs into a separate `DocumentFilter` class and then just call that in `DocumentCounter`. But that was more invasive than what I was trying for here.

pkarman: I went ahead and refactored for `DocumentFilter` and was able to cut some dupe code as well.

app/services/download_documents.rb
```diff
@@ -136,15 +110,6 @@ def before_package_contents
     # Test hook for testing race conditions
   end
 
-  def self.filter_documents(external_documents)
```
aroltsch: DownloadsDocuments file is not used anymore. Some of the files are used by efolder v1 api and should have been deleted. We probably never had a chance to get to it.

app/controllers/api/v2/manifests_controller.rb
```diff
@@ -47,6 +31,18 @@ def history
     render json: recent_downloads, each_serializer: Serializers::V2::HistorySerializer
   end
 
+  def document_count
+    manifest_id = params[:id]
+    cache_key = "manifest-doc-count-#{manifest_id}"
+    doc_count = Rails.cache.fetch(cache_key) do
```
aroltsch: Where is the expiration for the cache?

app/services/document_counter.rb
```diff
@@ -0,0 +1,14 @@
+class DocumentCounter
```
aroltsch: Nice!

app/controllers/api/v2/manifests_controller.rb
pkarman: ha! only in my mind! good catch.

app/services/download_documents.rb
pkarman: ok I will open ticket to remove API v1 files.

PR #1098, opened 2019-05-14
DocumentCounter can use Veteran or Manifest

pkarman:
**Description**

Adds a new API endpoint `GET /api/v2/veterans` which will return the document count for the Veteran.

**Why**

No manifest required. Returns the total documents across all services for any Veteran.

### DIFF
--- a/app/services/document_counter.rb
+++ b/app/services/document_counter.rb
@@ -1,14 +1,22 @@
 class DocumentCounter
   include ActiveModel::Model
 
-  attr_accessor :manifest
+  attr_accessor :manifest, :veteran
 
   def count
     total = 0
-    manifest.sources.each do |source|
-      documents = ManifestFetcher.new(manifest_source: source).documents
+    services.each do |service|
+      documents = service.v2_fetch_documents_for(manifest || veteran)
       total += DocumentFilter.new(documents: documents).filter.count
     end
     total
+  end
+
+  private
+
+  def services
+    return manifest.sources.map(&:service) if manifest
+    return [VBMSService, VVAService] if veteran
+    []
   end
 end
--- a/app/services/document_counter.rb
+++ b/app/services/document_counter.rb
@@ -1,22 +1,15 @@
 class DocumentCounter
   include ActiveModel::Model
 
-  attr_accessor :manifest, :veteran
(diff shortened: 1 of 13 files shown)


lib/fakes/document_service.rb
```diff
@@ -132,7 +137,7 @@ def self.create_document(i)
 
     OpenStruct.new(
       vbms_filename: "happy-thursday-#{SecureRandom.hex}.#{type[:ext]}",
-      type_id: Document::TYPES.keys.sample,
+      type_id: (11..20).to_a.sample,
```
pkarman: randomly picking from all types will eventually result in a RESTRICTED type and invalidate specs that assume all types are allowed.

app/services/document_counter.rb
```diff
@@ -1,14 +1,22 @@
...
       total += DocumentFilter.new(documents: documents).filter.count
     end
     total
   end
+
+  private
+
+  def services
+    return manifest.sources.map(&:service) if manifest
+    return [VBMSService, VVAService] if veteran
```
nemodjuric: Sources for any manifest are always both VBMS and VVA, if there's one, there's the other one as well (see here https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/models/manifest.rb#L31). So you can probably simplify this with ternary conditional or putting `if manifest || veteran`

pkarman:
> Why not have an endpoint where we just take the file_number and pull the docs without creating veteran (or manifest)?

good idea!

app/services/document_counter.rb
```diff
@@ -1,12 +1,13 @@
...
-  attr_accessor :manifest
+  attr_accessor :veteran_file_number
 
   def count
     total = 0
-    manifest.sources.each do |source|
-      documents = ManifestFetcher.new(manifest_source: source).documents
+    [VBMSService, VVAService].each do |service|
+      source = OpenStruct.new(file_number: veteran_file_number)
+      documents = service.v2_fetch_documents_for(source)
```
nemodjuric: I would change method v2_fetch_documents_for(source) to v2_fetch_documents_for(file_number), instead of creating an OpenStruct. Beside file_number field, object source is not used anywhere else inside the method.

PR #1101, opened 2019-05-20
Rename API route /veterans -> /document_counts

pkarman:
Renaming the API route for this new feature.


app/controllers/api/v2/document_count_controller.rb
```diff
@@ -1,4 +1,4 @@
-class Api::V2::VeteransController < Api::V2::ApplicationController
+class Api::V2::DocumentCountController < Api::V2::ApplicationController
```
aroltsch: I think we use plural names for controllers: DocumentCountController => DocumentCountsController

aroltsch: I will leave it up to you though

PR #1107, opened 2019-07-16
Resolves extra carriage return in response in connect-vbms

Sjones352:
Connects to https://github.com/department-of-veterans-affairs/connect_vbms/pull/226


pkarman:
the failing Travis test is the security check. That's a known issue I spent some time on yesterday and requires a bigger fix than we want to do here. I'm going to override and merge for now.

PR #1116, opened 2019-08-09
Make error handling consistent with Caseflow

monfresh:


app/exceptions/standard_error.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class StandardError
+  def ignorable?
+    false
```
nemodjuric: So, right now we have no ignorable errors in our code and this is for future implementation?

nemodjuric: I see now that ruby-bgs has this same method implemented.

monfresh: All the ignorable errors are now defined in the BGS, VBMS, and VVA gems. That way, we only have to maintain one set of ignorable errors instead of duplicating them in caseflow and efolder. Here is an example in BGS: https://github.com/department-of-veterans-affairs/ruby-bgs/blob/master/lib/bgs/errors.rb

PR #1118, opened 2019-08-16
Initial CircleCI configuration

nemodjuric:
This is basic CircleCI configuration similar to what we had on Travis with better logging and additional failing test spec/features/backend_error_flows_spec.rb which is marked pending (see #1119 )
After this, we should introduce workflow similar to caseflow-certification with linting and security as parallel tasks.


ci-bin/capture-log
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+# capture-log
```
enriquemanuel: What is this file for?

ci-bin/concatenate-log.rb
```diff
@@ -0,0 +1,22 @@
...
+# frozen_string_literal: true
+
+# concatenate-log.rb
+
+require "json"
+require "open-uri"
+
+artifacts_url = "https://circleci.com/api/v1.1/project"\
+                "/github/department-of-veterans-affairs"\
+                "/caseflow/#{ENV['CIRCLE_BUILD_NUM']}/artifacts"
```
enriquemanuel: Is this right? This is reading from Caseflow not Efolder

enriquemanuel:
@pkarman can you also check please?
And also check if you can remove TravisCi integration?

enriquemanuel:
Just removed TravisCI

ci-bin/capture-log
nemodjuric: We used this to store CircleCI output. We do similar in caseflow-certification repo.

ci-bin/concatenate-log.rb
nemodjuric: It should be caseflow-efolder

PR #1121, opened 2019-09-19
SSO redirect logging, first/last name in session

pkarman:
This PR does 2 things:

* adds more logging for SSO redirect logic
* captures the SAML first/last name attributes as separate fields

### DIFF
--- a/app/models/css_authentication_session.rb
+++ b/app/models/css_authentication_session.rb
@@ -2,7 +2,7 @@ class CssAuthenticationSession
   include ActiveModel::Model
   include ActiveModel::Serializers::JSON
 
-  attr_accessor :id, :name, :roles, :station_id
+  attr_accessor :id, :name, :roles, :station_id, :first_name, :last_name
   attr_reader :css_id, :email
 
   def email=(value)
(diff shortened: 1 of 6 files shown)


app/models/css_authentication_session.rb
```diff
@@ -2,7 +2,7 @@ class CssAuthenticationSession
   include ActiveModel::Model
   include ActiveModel::Serializers::JSON
 
-  attr_accessor :id, :name, :roles, :station_id
+  attr_accessor :id, :name, :roles, :station_id, :first_name, :last_name
```
pkarman: possibly useful to https://github.com/department-of-veterans-affairs/caseflow/issues/12113

lomky: If the name previously was just built out of "$first $last", wouldn't that imply the suffix is coming through how we're grabbing last here?

pkarman: if first name is `John E.` and last name is `Smith, Jr` then presumably it's easier to sort on the last name value if you don't need to parse spaces with a regex?

lomky: good point; it does grab us a valid start of the last name!

PR #1125, opened 2019-10-08
Remove unused env vars

aroltsch:
* Removing sidekiq env variables as we no longer use sidekiq.
* Removing metrics password/username since these were used by Prometheus

Test plan: 
- [x] deploy to UAT, run any job from the rails console, make sure it works

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -31,7 +31,7 @@ gem "redis-namespace"
 gem "redis-rails", "~> 5.0.2"
 gem "redis-semaphore"
 gem "request_store"
-gem "rubyzip", "~> 1.2.2"
+gem "rubyzip", ">= 1.3.0"
 gem "sass-rails", "~> 5.0"
 gem "sdoc", "~> 0.4.0", group: :doc
 gem "sentry-raven"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -368,7 +368,7 @@ GEM
     ruby_dep (1.5.0)
     ruby_parser (3.7.3)
       sexp_processor (~> 4.1)
-    rubyzip (1.2.2)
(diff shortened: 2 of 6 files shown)


Gemfile
```diff
@@ -31,7 +31,7 @@ gem "redis-namespace"
 gem "redis-rails", "~> 5.0.2"
 gem "redis-semaphore"
 gem "request_store"
-gem "rubyzip", "~> 1.2.2"
+gem "rubyzip", ">= 1.3.0"
```
aroltsch: Had to update it due to a security vulnerability

aroltsch:
### Problem: 
Caseflow eFolder could not connect to Redis when I deployed this PR

### Why testing through the rails console was not enough:
1. When I tested my changes in UAT, I made an assumption that ENV[“REDIS_URL_SIDEKIQ”] could only be used with a job so I sshed into the rails console and I ran `V2::DownloadManifestJob.perform_now(Manifest.last)`
2. It turns out ENV[“REDIS_URL_SIDEKIQ”] is used in models to store semaphores  which explains why the job that I tested through the console worked fine
3. Also, in the codebase, it is called `redis_url` instead of `redis_url_sidekiq`: `Rails.application.secrets.redis_url` so when I searched for `redis_url_sidekiq` I made an assumption that it was not used anywhere 

### Why did all the tests pass? 
If you pass `nil` url to the redis semaphore, it will by default initialize redis client pointing to the localhost `redis://127.0.0.1:6379/0` and in our test environment redis is running locally inside the machine. Meanwhile in production redis is not running locally and pointing to the remote host 

### Action items
1. For changes like that, I need a better test plan
2. Fix efolder rollback

PR #1129, opened 2019-11-12
Return Latest Record

lomaxap:
resolves https://github.com/department-of-veterans-affairs/caseflow/issues/12490

There are multiple `Record` rows for a `version_id`, we should return the latest record.

relevant slack thread: https://dsva.slack.com/archives/CHX8FMP28/p1573586741241500


spec/requests/api/v2/records_spec.rb
```diff
@@ -41,11 +44,20 @@
     context "when user has access to the corresponding manifest record" do
       let!(:files_download) { FilesDownload.create(user: current_user, manifest: manifest) }
 
-      it "returns a document" do
+      before do
+        Record.create(
+          record_attrs.merge(created_at: Time.zone.now - 5.days, mime_type: "old/record")
+        )
+      end
```
lomaxap: this test assumes that Record.find_by(version_id: version_id) would return the first created record.

spec/requests/api/v2/records_spec.rb
```diff
@@ -8,16 +8,19 @@
 
     let(:manifest) { Manifest.create(file_number: "1234") }
     let(:source) { ManifestSource.create(name: %w[VBMS VVA].sample, manifest: manifest) }
-
-    let(:record) do
-      Record.create(
-        version_id: "{3333-3333}",
-        series_id: "{4444-4444",
+    let(:record_attrs) do
```
pkarman: the only thing that must be the same is the version_id, and the manifest_source_id must be different, so maybe easier to write it as:

```ruby
let(:version_id) { "{333-333}" }
let(:manifest2) { Manifest.cerate(file_number: "5678") }
let(:source2) { ManifestSource.create(name: %w[VBMS VVA].sample, manifest: manifest2) }
let!(:record) do
  Record.create(
    created_at: Time.zone.now - 1.day,
    version_id: version_id,
    source: source,
    series_id: "{4444-4444}",
   # etc
  )
end
let!(:record2) dp
  Record.create(
    created_at: Time.zone.now,
    version_id: version_id,
    source: source2,
    # set
  )
end
```

and then verify that the record returned is `record2`

spec/requests/api/v2/records_spec.rb
```diff
@@ -41,11 +44,20 @@
     context "when user has access to the corresponding manifest record" do
       let!(:files_download) { FilesDownload.create(user: current_user, manifest: manifest) }
 
-      it "returns a document" do
+      before do
```
pkarman: create `record2` instead of this

spec/requests/api/v2/records_spec.rb
```diff
@@ -41,11 +44,20 @@
...
+        Record.create(
+          record_attrs.merge(created_at: Time.zone.now - 5.days, mime_type: "old/record")
+        )
+      end
+
+      let!(:record) { Record.create(record_attrs) }
```
pkarman: this is already created in the outer scope

spec/requests/api/v2/records_spec.rb
```diff
@@ -41,11 +44,20 @@
...
+
+      it "returns the latest document" do
         allow(S3Service).to receive(:fetch_content).and_return("hello there")
         get "/api/v2/records/#{version_id}"
         expect(response.code).to eq("200")
         expect(response.body).to eq("hello there")
+        expect(response.type).to eq(record.mime_type)
```
pkarman: `record2.mime_type`

pkarman:
gave my 👍 and then saw the failing test and why. some suggestions on how to rewrite the test, then lgtm

PR #1130, opened 2019-11-14
Screenreader not reading table in eFolder fix

ajspotts:
Resolves #12119

The `display:block` styling on the `<table>` element was blocking the screenreader from identifying the inherent semantical information and was unable to recognize the table.  

The solution was to reapply proper semantics to the elements via the ARIA guidelines on the table element and its children.


app/controllers/api/v2/records_controller.rb
```diff
@@ -32,7 +32,11 @@ def enable_caching
   end
 
   def record
-    @record ||= Record.includes(:manifest_source).find_by(version_id: "{" + params[:version_id] + "}")
+    @record ||= Record.includes(:manifest_source).where(version_id: version_id).order(created_at: :desc).first
```
kevmo: Why was this change necessary?

ajspotts: did a git rebase to pull in Andrew's solution because `brakeman` was still throwing errors for me locally when I attempted to implement myself, that bit there is some additional code he had added for his issue

PR #1134, opened 2019-11-20
Security updates

pkarman:
Addresses:

* CVE-2019-15587 (loofah)
* nokogiri
* rack-cors


Gemfile.lock
```diff
@@ -273,7 +273,8 @@ GEM
     public_suffix (3.0.2)
     puma (3.11.3)
     rack (2.0.7)
-    rack-cors (1.0.2)
+    rack-cors (1.1.0)
+      rack (>= 2.0.0)
```
cabeaulac: Why is this `rack (>= 2.0.0)` needed when `rack (2.0.7)` is two lines up? Or does it force the dependent package to use the included 2.0.7 instead of a 1.x? Similar to maven dependencies maybe

pkarman: the child dep listings just show what the dependencies are for the parent

Issue #1132, opened 2019-11-15
.zip files are invalid or too big


pkarman:
User has trouble downloading folder for Veteran with 13k+ documents. PID: 3978641

User has trouble downloading folder with invalid ZIP file. PID: 5052060

Related to INC7523674

alisan16:
Is this related to https://github.com/department-of-veterans-affairs/caseflow/issues/10665 and the VBMS::DocumentTooBig errors? It looks like there's an open task to add a fallback to use the updated VBMS paginated API when this happens: https://github.com/department-of-veterans-affairs/caseflow/issues/10901

pkarman:
> Maximum total size of ZIP archive
> A standard ZIP file can be up to 2^32 bytes in size (4 gigabyte of total size of the archive), ZIP64 standard allows the maximum archive size of 2^64 bytes (16 exabyte of total zip size), exceeding the current needs and limitation of most of current implementations of filesystems.
Please note some old zip archivers supports up to 2 GB for maximum archive size (and for max size of each input file), using a signed integer for representing the 32-bit wide size variable.

msj0nes:
@alisan16 let me know if I should open a VBMS ticket in your name regarding this issue and [other VBMS errors](https://github.com/department-of-veterans-affairs/caseflow/issues/10665#issue-440847078) present.  

**This would help clear:**
https://github.com/department-of-veterans-affairs/caseflow/issues/10732
https://github.com/department-of-veterans-affairs/caseflow/issues/10665
https://github.com/department-of-veterans-affairs/caseflow/issues/6680

**Info I need for ticket creation:**

- Data service associated to error
- Engineering Description of issue
- Case Details URL for Vet (PII Prevention - I can obtain ID from URL) 
- Best method of contact for VBMS to reach out

pkarman:
@msj0nes I don't think we're at the point yet (for this ticket anyway) where there is a clear problem with VBMS per se. Most of these require investigation into the root problem.

pkarman:
Confirmed that #1141 fixes this:
> User has trouble downloading folder with invalid ZIP file. PID: 5052060

This one seems to related to #1137

pkarman:
The 13k+ documents generates `VBMS::DocumentTooBig` like

https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/6744/events/575179/

pkarman:
Closing this as fixed, moving to #10901

PR #1141, opened 2019-12-09
Enable Zip64 support for large files

pkarman:
https://github.com/rubyzip/rubyzip#zip64-support

connects #1132 

## Description

Our Zip config defaults to a max file size of 4G. If there are enough files, or enough large files, compressed in a single .zip, it will error out and return a .zip file that cannot be opened.

This PR fixes that config and adds an optional test to verify. Because the test is slow to run and requires a very large example file, it is off by default and requires the `TEST_LARGE_FILES` env var to run. The intent is that developers should run this locally and not on CI.

```
TEST_LARGE_FILES=true bundle exec rspec spec/services/zipfile_creator_spec.rb
```

*DO NOT* check in the large file that is created. It will over write a local .pdf file. You need to `git checkout lib/pdfs` before committing any changes after running that test.


config/initializers/rubyzip.rb
```diff
@@ -0,0 +1,11 @@
+# https://github.com/rubyzip/rubyzip#configuration
+
+require "zip"
+
+Zip.setup do |c|
+#  c.on_exists_proc = true
+#  c.continue_on_exists_proc = true
+  c.unicode_names = true
+  c.default_compression = Zlib::BEST_COMPRESSION
+  c.write_zip64_support = true
```
pkarman: this is the important line.

lib/fakes/document_service.rb
```diff
@@ -83,11 +83,21 @@ def self.v2_fetch_document_file(record)
...
-      IO.binread(Rails.root + "lib/tiffs/#{record.id % 5}.tiff")
+    filename = if record.mime_type == "application/pdf"
+                 Rails.root + "lib/pdfs/#{record.id % 5}.pdf"
+               else
+                 Rails.root + "lib/tiffs/#{record.id % 5}.tiff"
+               end
+    fsize = File.size(filename)
+    buf = String.new(capacity: fsize) # pre-allocate to reduce number of mallocs
+    # for large files we must read in chunks.
+    # common page size is 4096 (4K) so we want 10k pages of memory (or about 40MB) per chunk.
```
aroltsch: How did you decide on this number of 40MB per chunk?

pkarman: Manually tested it with different number of pages to find a size that was fast enough. 4MB was too slow, and anything bigger than 40MB was indistinguishable from 40MB.

lib/fakes/document_service.rb
```diff
@@ -83,11 +83,21 @@ def self.v2_fetch_document_file(record)
   end
 
   def self.file_content(record)
-    if record.mime_type == "application/pdf"
-      IO.binread(Rails.root + "lib/pdfs/#{record.id % 5}.pdf")
-    else
-      IO.binread(Rails.root + "lib/tiffs/#{record.id % 5}.tiff")
+    filename = if record.mime_type == "application/pdf"
```
aroltsch: I am a little confused because this is in the fakes.

pkarman: It's in the fakes so we could test the configuration change for zip64.

The actual production (non-fake) code in https://github.com/department-of-veterans-affairs/caseflow-commons/blob/master/app/services/caseflow/s3_service.rb#L40 uses `IO.binread` and it works fine.

In our test we have a single large file (500MB) in order to more quickly create a large .zip file. In production, we have 1000s of smaller files that add up to one large file. So this change to the fakes is to support tests only.

If we find a further error with the `S3Service.fetch_content` we can rewrite in the caseflow-commons repo. But that's outside the scope of this PR (which is mostly just about one line of configuration change).

PR #1142, opened 2019-12-10
Fix test db data for veteran_info by including required return_message

alisan16:
#1060 (note this PR was reverted and eventually re-added in #1063 ) added a required check for `return_message` field on `veteran_info` but did not update the test data on the DEMO# file numbers. Update the test data so that in the development environment, we can actually see the magic of eFolder and fake documents!

Before fixing, querying any of the [DEMO# test file numbers](https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/lib/fakes/document_service.rb#L7) resulted in an error:
Screen Shot 2019-12-10 at 5 18 15 PM

After fixing, using any of the DEMO test files correctly pulls fake files:
Screen Shot 2019-12-10 at 5 16 02 PM

Acceptance criteria
- [x] Following the repo README to test out the DEMO flow (without VBMS connection) now works


pkarman:
I opened #1143 to fix how CI displays lint errors.

alisan16:
Oh no this is the style guidance is the opposite to what I'm used to - we always added trailing commas to better preserve blame

pkarman:
@alisan16 I agree -- feel free to propose a change to the rubocop rules via a PR. I know efolder is out of sync with caseflow, among other things.

PR #1149, opened 2019-12-12
Update setup instructions in eFolder express README 

alisan16:
After going through the development set up process, update the README. The last section of testing in UAT has not been tested.


README.md
```diff
@@ -49,77 +53,73 @@ docker-compose down -v
 ## First Time Development Setup
 
 You'll need Ruby 2.5.3
```
pkarman: Maybe we should make this more evergreen and instead say:

> You'll need the proper version of Ruby:

```
rbenv install `cat .ruby-version`
```

README.md
```diff
@@ -49,77 +53,73 @@ docker-compose down -v
...
 
 If a pending migration exists, you will need to run them against both the development and test database:
-
-> $ rake db:migrate
-
-> $ RAILS_ENV=test rake db:migrate
-
+```
+make migrate
+RAILS_ENV=test bundle exec rake db:migrate
```
pkarman: that second line might not be necessary anymore, based on some experiences in caseflow. So `make migrate` might be sufficient. Doesn't hurt to run the 2nd line though.

PR #1164, opened 2020-01-02
VBMS document pagination

### LINKED ISSUE #1160: Support VBMS paged documents
https://github.com/department-of-veterans-affairs/caseflow/issues/10901 describes the need for paged VBMS requests in order to fetch efolder contents with 5k+ documents.

- [x] update the connect_vbms gem
- [x] add VBMS paged documents service (possibly in gem to share with Caseflow?)
- [x] test with known large efolders

pkarman:
Resolves #1160 

## Description

Updates VBMS and BGS gems.

Adds feature toggle `:vbms_pagination` to control whether we use the new VBMS paged documents service to allow fetching VBMS efolders where there are 5k+ documents.

## Testing Plan
1. Identify a Veteran with 5k+ efolder in UAT (https://dsva.slack.com/archives/CK466V65V/p1577992111002300?thread_ts=1577990211.002000&cid=CK466V65V)
1. In Rails console, toggle the `:vbms_pagination` feature for yourself.
1. In Rails console, call the `ExternalApi::VBMSService.v2_fetch_documents_for` method for the Veteran.

Repeat in production after deployment.

See also https://github.com/department-of-veterans-affairs/caseflow/pull/13084


app/services/external_api/vbms_service.rb
```diff
@@ -16,6 +16,19 @@ def log(event, data)
   end
 end
 
+class RailsVBMSDebugLogger
```
pkarman: useful locally via console. pass it to initialize a VBMS client.

app/services/external_api/vbms_service.rb
```diff
@@ -67,4 +87,13 @@ def self.send_and_log_request(id, request)
       @vbms_client.send_request(request)
     end
   end
+
+  def self.call_and_log_service(service:, vbms_id:)
+    name = service.class.name.split("::").last
+    MetricsService.record("call #{service.class} for #{vbms_id}",
+                          service: :vbms,
```
alisan16: Should this be the `service` that's passed into the method rather than `:vbms`?

pkarman: good question. I think the word "service" is overloaded here.

the existing convention seems to be to log the DD "service" as `:vbms` for everything in this class.

The `VBMS::Service::*` classes are what the `service` object here is named for.

PR #1165, opened 2020-01-02
Fix flakey download tests

pkarman:
We had 2 problems running tests at CircleCI (not locally):

* flakey VBMS and VVA error handling (specs were skipped)
* download tests did not work at all (specs were skipped)

This PR fixes both problems.

* The flakey error handling was due to DataDogService causing a 130-second pause when running background jobs. The solution was to explicitly mock DataDogService in the affected tests so that we do not attempt to send data to DD.
* The download tests failed because CircleCI had out-dated chrome/chromedriver version.

In order to diagnose these 2 problems, I created a new feature test with the simplest of interactions: touch a file on the file system, then download that file via the browser. That exercises both the background jobs via `perform_enqueued_jobs` helper and the browser download functionality.

The chrome/driver version mismatch was diagnosed by ssh'ing to the failed CircleCI instance and manually comparing the installed version vs what I had installed locally. CI has 72.x, I had 79.x.


.circleci/config.yml
```diff
@@ -74,10 +74,21 @@ jobs:
             ~/project/ci-bin/capture-log "RAILS_ENV=test bundle exec rake db:create"
             ~/project/ci-bin/capture-log "RAILS_ENV=test bundle exec rake db:schema:load"
 
+      - run:
+          name: Update Chrome/ChromeDriver
```
pkarman: always test agains the latest stable chrome version.

app/services/metrics_service.rb
```diff
@@ -13,7 +13,7 @@ def self.record(description, service: nil, name: "unknown")
       return_value = yield
     end
 
-    if service
+    if service && Rails.env.production?
```
pkarman: we only want to use DataDog where we have it running.

spec/features/backend_error_flows_spec.rb
```diff
@@ -35,6 +35,8 @@
   end
 
   before do
+    allow(DataDogService).to receive(:emit_gauge) { true } # mock DD during tests
```
pkarman: this fixes problem with 130-second pause

spec/features/active_job_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+    User.authenticate!
+    DownloadHelpers.clear_downloads
+  end
+
+  after do
+    DownloadHelpers.clear_downloads
+  end
+
+  context "when job is queued" do
+    scenario "the job runs" do
```
alisan16: Do we also need `download: true` here?

PR #1166, opened 2020-01-07
Upgrade Rails to 5.2.4.1

pkarman:
Branched from #1165 so merge that first.


Gemfile.lock
```diff
@@ -178,151 +173,149 @@ GEM
...
-    delayed_job (4.1.4)
-      activesupport (>= 3.0, < 5.2)
+    database_cleaner (1.7.0)
+    delayed_job (4.1.8)
+      activesupport (>= 3.0, < 6.1)
     diff-lcs (1.3)
-    distribute_reads (0.2.4)
-      makara
-    docile (1.1.5)
-    dogstatsd-ruby (3.3.0)
```
ferristseng: Noticing some other dependencies here that don't seem related to Rails. This in particular is upgrading a major version, which concerns me a little bit. Would it be better to isolate to just upgrading Rails?

pkarman: re-done to limit to Rails

PR #1175, opened 2020-01-15
Revert to previous streaming response code over send_data

pkarman:
Along with the Rails 5.2 upgrade we changed the way the zip file download HTTP response was handled.  This introduced a regression that our automated testing did not cover and our manual testing did not check.

This PR reverts to the older pattern and adds automated testing.


spec/features/react_download_spec.rb
```diff
@@ -87,6 +90,20 @@
...
+    DownloadHelpers.wait_for_download
+    download = DownloadHelpers.downloaded?
+    expect(download).to be_truthy
+
+    expect(DownloadHelpers.download).to include(filename)
+    expect(DownloadHelpers.filesize).to be > 32 # any smaller is invalid zip
+
+    zip_path = DownloadHelpers.download
+    Zip::File.open(zip_path) do |zip_file|
+      expect(zip_file.size).to eq documents.count
```
jcq: Is it sufficient to just check the number of documents? Might it be worth comparing contents of one of them or something to ensure they haven't somehow become corrupted?

Certainly this is far better than before (and would have caught today's issue), so definitely still worth merging.

Issue #1043, opened 2018-10-04
Replace root path in routes and remove DownloadsController and Download model


nemodjuric:
We don't use download model anymore. However, root path still has downloads#new. 
We should remove all the code related to download(s).

alisan16:
Closing as this hasn't been updated in 2 years

pkarman:
I have an open PR against this one.

pkarman:
closing in favor of #1095

PR #1182, opened 2020-01-24
Refactor the code to not use DownloadsController

pkarman:
opening as a reminder to myself to use this branch as a reference point for the clean up.


pkarman:
see #1043

Issue #1178, opened 2020-01-17
SSOi SAML migration


pkarman:
The existing SAML SSO solution we use is being retired and we must migrate. This ticket captures some technical notes.

### Background links

* https://github.com/department-of-veterans-affairs/appeals-deployment/wiki/CSS-SAML-Workflow
* https://github.com/department-of-veterans-affairs/appeals-deployment/wiki/Certificates#saml-message-signing

### Email from BEP

The new SAML solution does not provide CSS/station information. That must be fetched from VBA API via 2nd step. The new VBA API is under active development.

> The authentication process will be divided in two consecutive steps:
> • IAM Authentication: The User attempts to connect to the protected resource. The request is 
> redirected to IAM according to the pattern decided to challenge the User for credentials which are 
> validated against the VA Active Directory.
> • CSS Authorization: Once the User has been authenticated by IAM and the request is back, the 
> Application should call methods from the BGS CommonSecurityService to determine if the user has 
> authorization to access the application.

per email:

For the getCssUserStationsByApplicationUsername you will have to pass a System Account User/Station. You can pass an empty password.

```xml
<soapenv:Envelope xmlns:soapenv="http://schemas.xmlsoap.org/soap/envelope/" xmlns:v1="http://types.ws.css.vba.va.gov/services/v1">
<soapenv:Header>
<wsse:Security xmlns:wsse="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-secext-1.0.xsd">
<wsse:UsernameToken xmlns:wsu="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-wssecurity-utility-1.0.xsd">
  <wsse:Username>VBMSSYSACCNT</wsse:Username> <!--VBMS SYSTEM ACCOUNT USERNAME -->
  <wsse:Password Type="http://docs.oasis-open.org/wss/2004/01/oasis-200401-wss-username-token-profile-1.0#PasswordText"></wsse:Password>  <!—Can be empty-->
</wsse:UsernameToken>
<vaws:VaServiceHeaders xmlns:vaws="http://vbawebservices.vba.va.gov/vawss">
  <vaws:CLIENT_MACHINE>10.224.104.174</vaws:CLIENT_MACHINE>  <!-- The IP of the user machine -->
  <vaws:STN_ID>317</vaws:STN_ID>  <!—VBMS SYSTEM ACCOUNT STATION -->
[...]

PR #1184, opened 2020-01-29
SSOi IAM SAML

pkarman:
connects #1178 

### Description

New IAM SSO SAML provider.

* Adds feature toggle `:use_ssoi_iam` to allow toggling between different IdPs
* Small refactor of BGS service.
* Upgrades BGS gem to get latest authz endpoints.

Depends on https://github.com/department-of-veterans-affairs/appeals-deployment/pull/2474

### Testing

- [x] deploy branch to UAT and confirm current SSO works
- [x] toggle feature and redeploy to UAT
- [x] confirm new SSO works
- [x] revert toggle (off)


app/services/external_api/bgs_service.rb
```diff
@@ -66,6 +100,7 @@ def init_client
       ssl_cert_file: ENV["BGS_CERT_LOCATION"],
       ssl_ca_cert: ENV["BGS_CA_CERT_LOCATION"],
       forward_proxy_url: forward_proxy_url,
+      jumpbox_url: ENV["RUBY_BGS_JUMPBOX_URL"],
```
pkarman: caseflow has this

pkarman:
tested manually in uat and all ok for now.

feature toggle disabled

```
irb(main):002:0> FeatureToggle.disable!(:use_ssoi_iam)
=> true
irb(main):003:0> FeatureToggle.enabled?(:use_ssoi_iam)
=> false
```

app/services/external_api/bgs_service.rb
```diff
@@ -47,12 +51,42 @@ def record_found?(veteran_info)
     veteran_info["return_message"].include?("No BIRLS record found") ? false : true
   end
 
+  def fetch_user_info(username)
+    resp = client.common_security.get_css_user_stations(username)
+    # example
+    # {:network_login_name=>"CF_Q_283", :user_application=>"CASEFLOW", :user_stations=>{:enabled=>true, :id=>"283", :name=>"Hines SDC", :role=>"User"}}
+    # TODO if we have more than one station
+    css_id = resp[:network_login_name] # probably the same as username but just in case.
+    station_id = Array.wrap(resp[:user_stations]).first[:id]
```
alisan16: Based on the example, do we need to grab the user station where `enabled = true`?

app/services/external_api/bgs_service.rb
```diff
@@ -47,12 +51,42 @@ def record_found?(veteran_info)
...
+    #  ],
+    #  :job_title=>"Example Review Officer", :last_name=>"ONE", :message=>"Success", :middle_name=>nil, :participant_id=>"123"
+    # }
+    {
+      css_id: css_id,
+      station_id: station_id,
+      first_name: resp2[:first_name],
+      last_name: resp2[:last_name],
+      email: resp2[:email_address],
+      roles: Array.wrap(resp2[:functions]).select { |func| func[:assigned_value] == "YES" }.map { |func| func[:name] }
```
alisan16: Similar question here - what does `:disable_ind` represent?

config/initializers/omniauth.rb
```diff
@@ -15,8 +18,29 @@ def ssoi_authentication_enabled?
...
+        # we just care about it during actual app startup during deployment step.
+end
+
 # :nocov:
-if ssoi_authentication_enabled?
+if use_ssoi_iam?
+  # for transition to new IdP. Once fully deployed, we can remove the older ENV vars and certs.
+  Rails.application.config.middleware.use OmniAuth::Builder do
+    provider :samlva,
+      "https://efolder.cf.ds.va.gov", # same in all envs # ENV[ENV_SAML_ID],
```
alisan16: Why is this the same for all environments?

app/services/external_api/bgs_service.rb
pkarman: Yes. I am working on a follow-on PR with a lot more error handling and that filter. This is just a POC for this authz feature (mostly to make sure the feature toggle works as expected).

app/services/external_api/bgs_service.rb
pkarman: I'm not sure. I will check with BGS. Should not block this PR though.

config/initializers/omniauth.rb
pkarman: SSOI set it up that way so that we don't have to change our SAML config between UAT and preprod. It's not a URL in the HTTP sense, it's just a unique string for the SAML id.

PR #1185, opened 2020-02-03
Test SAML IdP

pkarman:
Follow on to #1184 

## Description

This PR implements two things:

* In order to test our SAML flow,  we now have a FakeSamlIdp service that mimics the entire IAM SAML handshake. Certs/keys/config are created, expiring in 10 years. (Future improvement could create these on-demand as part of the test suite.)
* The test-user-id/station-id authorization screen is no longer part of the IdP SAML process, so we implement our own at the front-end of the process. This allows us to capture the intended station id the user wants to use (if they have more than one), and in non-production environments, allows the user to override the CSS ID associated with the PIV to use a test user id. Full flow is now documented at the top of the `sessions_controller` (and will be added to the wiki in beautifully illustrated format).

## Testing

As with #1184 this branch should be tested independently in UAT before merging. Follow the same pattern:

* [x] deploy branch
* [x] validate existing BEP IdP
* [x] toggle feature
* [x] re-deploy branch
* [x] validate new IAM IdP
* [x] toggle feature
* [x] re-deploy branch
* [x] validate existing BEP IdP

## Examples

New /login page (Test User Id field hidden in production)
Screen Shot 2020-02-05 at 8 51 41 AM

In local development only, the next step would be:
Screen Shot 2020-02-05 at 8 51 29 AM

In all AWS environments, the next step is the new SSO IAM IdP:
Screen Shot 2020-02-05 at 10 01 50 AM

Error on log in
Screen Shot 2020-02-03 at 4 38 59 PM


app/exceptions/bgs_errors.rb
```diff
@@ -0,0 +1,8 @@
+module BGS
```
alisan16: Can we change the error message displayed to the user to be more readable for each of these? Is it standard in Caseflow right now to display errors in this format?

Not blocking for this PR

Gemfile
```diff
@@ -21,7 +21,8 @@ gem "mini_magick"
 gem "moment_timezone-rails"
 gem "newrelic_rpm"
 gem "nokogiri", ">=1.10.5"
-gem "omniauth-saml-va", git: "https://github.com/department-of-veterans-affairs/omniauth-saml-va", ref: "fbe2b878c250b14ee996ef6699c42df2c42e41a1"
+gem "omniauth-saml-va", git: "https://github.com/department-of-veterans-affairs/omniauth-saml-va", branch: "pek-iam-ssoi"
```
alisan16: Does this need to be changed back?

app/services/external_api/bgs_service.rb
```diff
@@ -69,14 +78,20 @@ def fetch_user_info(username)
...
-      email: resp2[:email_address],
-      roles: Array.wrap(resp2[:functions]).select { |func| func[:assigned_value] == "YES" }.map { |func| func[:name] }
+      first_name: profile[:first_name].strip,
+      last_name: profile[:last_name].strip,
+      email: profile[:email_address].strip,
+      roles: Array.wrap(profile[:functions]).select { |func| func[:assigned_value] == "YES" }.map { |func| func[:name] }
     }
-  rescue BGS::ShareError
+  rescue BGS::ShareError => error
+    fail BGS::InvalidUsername if error.message =~ /Unable to get user authroization/
```
alisan16: Typo here for "authorization"

lib/fakes/bgs_service.rb
```diff
@@ -1,19 +1,55 @@
+require 'bgs_errors'
+
 class Fakes::BGSService
   include ActiveModel::Model
 
-  def fetch_user_info(username)
+  def fetch_user_info(username, station_id = nil)
+    fail "Must defined current_user" unless RequestStore[:current_user] # mock what real service requires
```
alisan16: Typo: "define"

Gemfile
pkarman: in a follow on PR, yes.

app/exceptions/bgs_errors.rb
pkarman: I wanted to get design folks to help me think about language, so I thought we could start with just the error class name. Most of these reflect error conditions that would only show up in edge cases (multiple stations or in UAT/preprod only, where the username is overrideable).

app/services/external_api/bgs_service.rb
pkarman: agreed, but it's not our typo. It's a known typo in BGS.

Issue #1137, opened 2019-12-05
File number might be SSN or claim number


pkarman:
https://dsva.slack.com/archives/CFBLH2LR4/p1575577328364300

![claims-processing-file-number-ssn-or-claim-number](https://user-images.githubusercontent.com/1205061/70280784-e1cf5180-177e-11ea-80b3-6284448266a7.png)

When fetching documents from VBMS we should try with both the claim number and the SSN because VBMS may store under either one.

pkarman:
Example veteran. Two different BGS records, one using SSN as file_number (participant_id `30212588`) and the other using claim number as file_number (participant_id `11968303`).

pkarman:
confirmed queries work as expected in production, though VBMS has cleaned up this particular veteran already.

Interestingly, VVA reports the same documents regardless of which number is passed (claim or SSN).

PR #1188, opened 2020-02-10
Add VeteranFinder, prefer bgs :file_number

pkarman:
connects #1137 

### Description

Adds `VeteranFinder` service class to search all possible combinations of ssn/claim/file numbers within BGS.

Previously, we assumed BGS `claim_number` was always the `file_number`. That is not true. SSN or claim number may legitimately be used as the file number.

Search for all VBMS efolders by file_number reported for all found BGS records.

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -13,7 +13,7 @@ class ExternalApi::BGSService
     ssn = veteran_data[:ssn] ? veteran_data[:ssn] : veteran_data[:soc_sec_number]
     last_four_ssn = ssn ? ssn[ssn.length - 4..ssn.length] : nil
     {
-      "file_number" => veteran_data[:claim_number],
+      "file_number" => veteran_data[:file_number],
       "veteran_first_name" => veteran_data[:first_name],
       "veteran_last_name" => veteran_data[:last_name],
       "veteran_last_four_ssn" => last_four_ssn,
@@ -21,13 +21,14 @@ class ExternalApi::BGSService
     }
   end
 
-  def fetch_veteran_info(file_number)
+  def fetch_veteran_info(file_number, parsed: true)
     veteran_data =
       MetricsService.record("BGS: fetch veteran info for vbms id: #{file_number}",
                             service: :bgs,
                             name: "veteran.find_by_file_number") do
         client.veteran.find_by_file_number(file_number)
       end
+    return veteran_data unless parsed
     parse_veteran_info(veteran_data) if veteran_data
   end
 
--- a/app/services/veteran_finder.rb
+++ b/app/services/veteran_finder.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+# return all the BGS record data for a file number.
+# this may return multiple BGS records, uniquely identified by participant_id
+
+class VeteranFinder
+  def find(file_number)
+    bgs_rec = bgs_record_for(file_number)
+
+    return unless bgs_rec
+
+    bgs_rec_numbers = bgs_numbers(bgs_rec)
+    bgs_rec2 = find_duplicate_bgs_rec(bgs_rec_numbers)
+
+    return [bgs_rec_numbers, bgs_numbers(bgs_rec2)] if bgs_rec2
+
+    [bgs_rec_numbers]
+  end
+
+  private
+
+  def find_duplicate_bgs_rec(bgs_rec_numbers)
+    if bgs_rec_numbers[:file].to_s == bgs_rec_numbers[:ssn].to_s
+      # look again by claim number
+      bgs_record_for(bgs_rec_numbers[:claim])
+    elsif bgs_rec_numbers[:file].to_s == bgs_rec_numbers[:claim].to_s
+      # look again by ssn
+      bgs_record_for(bgs_rec_numbers[:ssn])
+    else
+      fail "Found BGS file_number not equal to SSN or claim number: #{bgs_rec_numbers[:file]}"
+    end
+  end
+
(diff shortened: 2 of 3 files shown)


app/services/external_api/bgs_service.rb
```diff
@@ -13,21 +13,22 @@ def parse_veteran_info(veteran_data)
     ssn = veteran_data[:ssn] ? veteran_data[:ssn] : veteran_data[:soc_sec_number]
     last_four_ssn = ssn ? ssn[ssn.length - 4..ssn.length] : nil
     {
-      "file_number" => veteran_data[:claim_number],
+      "file_number" => veteran_data[:file_number],
```
leikkisa: Oh, having the claim_number in the original might explain some things!

app/services/veteran_finder.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+  end
+
+  def bgs_numbers(bgs_rec)
+    {
+      ssn: bgs_rec[:soc_sec_number] || bgs_rec[:ssn],
+      claim: bgs_rec[:claim_number],
+      file: bgs_rec[:file_number],
+      participant_id: bgs_rec[:ptcpnt_id]
+    }.merge(bgs.parse_veteran_info(bgs_rec))
```
leikkisa: The method naming for bgs_numbers was a little misleading for me. I thought at first it was just the numbers but see now that it's returning more of an augmented bgs_record with the additional numbers.

app/services/manifest_fetcher.rb
```diff
@@ -16,6 +16,18 @@ def process
...
   def documents
-    manifest_source.service.v2_fetch_documents_for(manifest_source.file_number)
+    @documents ||= fetch_documents
+  end
+
+  def fetch_documents
+    # fetch documents for all the "file numbers" known for this veteran
+    file_numbers.map do |file_number|
+      manifest_source.service.v2_fetch_documents_for(file_number)
+    end.flatten
```
leikkisa: I don't know enough about the data, but if there are two records, would they ever share duplicates documents with each other?

spec/requests/api/v2/manifests_spec.rb
```diff
@@ -59,6 +59,69 @@
...
+      {
+        "veteran_first_name" => "Bob",
+        "veteran_last_name" => "Marley",
+        "veteran_last_four_ssn" => "1234",
+        "return_message" => "BPNQ0301",
+      }
+    end
+    let(:veteran_claim_number) { "12345678" }
+    let(:veteran_ssn) { "666001234" }
+    let(:veteran_info) do
```
leikkisa: This doesn't match the structure of what I would expect bgs.veteran_info to return.

spec/requests/api/v2/manifests_spec.rb
```diff
@@ -59,6 +59,69 @@
     end
   end
 
+  context "When more than one BGS Veteran record matches the file number" do
+    before do
+      allow_any_instance_of(Fakes::BGSService).to receive(:veteran_info).and_return(veteran_info)
+      allow_any_instance_of(VeteranFinder).to receive(:find) { [ { file: veteran_claim_number }, { file: veteran_ssn } ] }
```
leikkisa: It looks like the `find` method is only getting called with a file_number as the argument, which doesn't match what I see here.

app/services/veteran_finder.rb
pkarman: The `merge ..` part is not strictly necessary. I was using it in order to manually/visually confirm that the names returned for both records were the same.

app/services/manifest_fetcher.rb
pkarman: they might have duplicate documents, but we have no way of knowing. VBMS does what it does.

We already de-duplicate them here https://github.com/department-of-veterans-affairs/caseflow-efolder/blob/master/app/services/document_creator.rb#L13

spec/requests/api/v2/manifests_spec.rb
pkarman: I'm not sure I understand your comment. The `allow_any_instance_of` declaration does not specify a `with(...)` argument at all. The block at the end is the equivalent of `.and_return(...)` but evaluated at run time.

spec/requests/api/v2/manifests_spec.rb
pkarman: if you look at `lib/fakes/bgs_service.rb` you'll see that the `def veteran_info` method is a stub intended to be overridden in tests. It should return a Hash, as it is consumed by the `fetch_veteran_info` method in that same file.

What structure would you expect?

pkarman:
Thanks for the review @leikkisa -- hopefully I responded to all your comments/questions.

spec/requests/api/v2/manifests_spec.rb
leikkisa: Here's what I'm seeing:
```
{"12345678"=>{"veteran_first_name"=>"Bob", "veteran_last_name"=>"Marley", "veteran_last_four_ssn"=>"1234", "return_message"=>"BPNQ0301"},
 "666001234"=>{"veteran_first_name"=>"Bob", "veteran_last_name"=>"Marley", "veteran_last_four_ssn"=>"1234", "return_message"=>"BPNQ0301"}}
```

I wasn't expecting it to be multiple hashes keyed by the number, just one hash that might look more like one of the individual veteran records.

leikkisa: Ok I got eFolder installed and was able to look at it! I think I was misunderstanding the test data setup for fetch_veteran_info, but it looks good.

PR #1197, opened 2020-02-17
Fix regressions in bgs client and veteran de-duping

pkarman:
Fixes 2 regressions found in recent changes to how we initialize BGSService and how we de-duplicate Veteran records.


app/services/veteran_finder.rb
```diff
@@ -33,6 +33,8 @@ def find_duplicate_bgs_rec(bgs_rec_numbers)
     # common enough in missing BGS data that we just return nil
     return if bgs_rec_numbers[:file].blank?
 
+    return if bgs_rec_numbers[:ssn].blank?
```
pkarman: apparently there bgs records with no ssn.

PR #1212, opened 2020-03-20
Bundle update, skip CVE-2020-5267

pkarman:
Skipping CVE for now till Rails 5.2.4.2 is released.


Gemfile
```diff
@@ -32,7 +32,6 @@ gem "redis-semaphore"
 gem "request_store"
 gem "rubyzip", ">= 1.3.0"
 gem "sass-rails", "~> 5.0"
-gem "sdoc", "~> 0.4.0", group: :doc
```
pkarman: we do not use this

Issue #1218, opened 2020-05-04
Apply POA rules like Caseflow


pkarman:
EE currently applies a simple POA check that relies on BGS error message to disqualify viewing of VBMS efolder documents.

VBMS itself uses the same BGS POA participant id API calls that Caseflow does, so switch over to those.

See context in https://dsva.slack.com/archives/C3EAF3Q15/p1588359644063900

VBMS POA rules:

<img width="714" alt="Screen Shot 2020-05-05 at 10 31 58 AM" src="https://user-images.githubusercontent.com/1205061/81084606-b54d6c00-8ebb-11ea-889a-80ea56752201.png">

Some POA rules in this older ticket too: https://github.com/department-of-veterans-affairs/caseflow-efolder/issues/410#issuecomment-285137660

### Testing

* [x] Enable feature toggle `:user_authorizer` for known POA user having trouble https://dsva.slack.com/archives/CHX8FMP28/p1588101644494200
* [x] If successful, enable feature toggle for all users
* [x] After 2 weeks, remove feature toggle code.

pkarman:
tested as of https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1229 and working.

PR #1219, opened 2020-05-04
Power of Attorney rules (part 1)

pkarman:
connects #1218 

* Add BGS service endpoints for PID and Person services. This will allow us to begin implementing POA rules based on the same endpoints that Caseflow uses.
* Add `participant_id` to the `User` model so we can look up POA matches more easily for the `current_user`.
* Start re-orienting the authorization model to check permissions at the boundaries for BGS calls.

### DIFF
--- a/app/services/concerns/poa_mapper.rb
+++ b/app/services/concerns/poa_mapper.rb
@@ -1,0 +1,29 @@
+module POAMapper
+  extend ActiveSupport::Concern
+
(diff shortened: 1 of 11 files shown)


app/services/concerns/poa_mapper.rb
```diff
@@ -0,0 +1,29 @@
+module POAMapper
+  extend ActiveSupport::Concern
```
pkarman, line 2: cribbed from caseflow

PR #1221, opened 2020-05-05
POA rules (part 2)

pkarman:
connects #1218 

* use latest bgs gem
* add support for user CSS record on demand
* use BGS claimants service for POA by file number
* adds `deceased` boolean flag to veteran info
* BGS service tests


app/services/external_api/bgs_service.rb
```diff
@@ -65,10 +66,10 @@ def fetch_veteran_info(file_number, parsed: true)
   def fetch_poa_by_file_number(file_number)
     bgs_poa = MetricsService.record("BGS: fetch poa for file number: #{file_number}",
                                     service: :bgs,
-                                    name: "org.find_poas_by_file_number") do
-      client.org.find_poas_by_file_number(file_number)
+                                    name: "claimants.find_poas_by_file_number") do
```
ferristseng, line 69: Small nit to also update the method name here to match

app/services/external_api/bgs_service.rb
```diff
@@ -65,10 +66,10 @@ def fetch_veteran_info(file_number, parsed: true)
...
+      client.claimants.find_poa_by_file_number(file_number)
```
ferristseng, line 70: How does BGS distinguish these two calls? Does the previous version interpret the file number as an org's file number?

app/services/external_api/bgs_service.rb
pkarman: yep fixed in the follow on PR already

app/services/external_api/bgs_service.rb
pkarman: I have no idea. I think they must be implementing some claimants logic (like the claimant has POA but the veteran does not) in the `claimants` service, but not the `org` service.

PR #1223, opened 2020-05-06
POA rules (part 3)

pkarman:
connects #1218 

* adds `UserAuthorizer` class to implement POA rules
* adds `user_authorizer` feature toggle to enable new rules

Previously the POA check was a simple error string match with yes/no.

Now we check explicitly for POA record on the claimant and veteran, independently, and also whether the Veteran is deceased (in which case the claimant's POA has precedence).


spec/features/user_error_flows_spec.rb
```diff
@@ -127,6 +127,81 @@
     end
   end
 
+  context "user_authorizer feature toggle on" do
+    before { FeatureToggle.enable!(:user_authorizer) }
```
pkarman, line 131: this test block is essentially a copy/paste of the block above it, with the feature toggle turned on.

app/services/user_authorizer.rb
```diff
@@ -0,0 +1,132 @@
...
+  rescue StandardError => err
+    self.sensitive_file = true if err.message.include?("Sensitive File - Access Violation")
+    self.poa_denied = true if err.message.include?("Power of Attorney of Folder is")
+    {}
+  end
+
+  def fetch_veteran_record_as_system_user
+    system_bgs.fetch_veteran_info(file_number)
+  rescue StandardError => err
+    # to do: log?
```
alisan16, line 108: Can we create a ticket for this if it's not being handled soon?

app/services/user_authorizer.rb
```diff
@@ -0,0 +1,132 @@
...
+#
+# The rules for POA for deceased veteran that VBMS uses:
+#
+# 1. Find if the vet is represented by the POA
+# 2. Find if the Vet is deceased
+# 3. Pull a list of all claimants for the Vet
+# 4. See if the POA represents any of those claimants
+# 5. Grant access to the efolder
+
+class UserAuthorizer
```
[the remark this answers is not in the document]
pkarman: it is not only used for POA. It is *the* class for determining whether a User is authorized to read an efolder.

app/services/user_authorizer.rb
```diff
@@ -0,0 +1,132 @@
...
+  attr_reader :file_number, :user
+
+  def initialize(file_number:, user:)
+    @file_number = file_number
+    @user = user
+  end
+
+  def can_read_efolder?
```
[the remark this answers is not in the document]
pkarman: good call. done.

app/services/user_authorizer.rb
pkarman: I removed the comment. All errors will be logged anyway and there's nothing else actionable here.

app/controllers/api/v2/application_controller.rb
```diff
@@ -27,6 +27,29 @@ def verify_veteran_file_number
   end
 
   def fetch_veteran_by_file_number(file_number)
```
pkarman, line 29: this method controls all user authorization for the entire app.

We have moved the bulk of the logic into a service class `UserAuthorizer` and use that service class here instead.

app/controllers/api/v2/application_controller.rb
```diff
@@ -27,6 +27,29 @@ def verify_veteran_file_number
...
+    if FeatureToggle.enabled?(:user_authorizer, user: current_user)
+      fetch_veteran_with_user_authorizer(file_number)
+    else
+      fetch_veteran_with_bgs_service(file_number)
+    end
+  end
+
+  def fetch_veteran_with_user_authorizer(file_number)
```
pkarman, line 37: once we are confident in the `UserAuthorizer` the feature toggle can be removed and this becomes the `fetch_veteran_by_file_number` method.

app/controllers/api/v2/application_controller.rb
```diff
@@ -27,6 +27,29 @@ def verify_veteran_file_number
...
+    elsif authorizer.veteran_record_poa_denied?
+      vso_denied_record
+    elsif !authorizer.veteran_record_found?
+      veteran_not_found(file_number)
+    else
+      raise BGS::ShareError, "Cannot access Veteran record"
+    end
+  end
+
+  def fetch_veteran_with_bgs_service(file_number)
```
pkarman, line 52: `UserAuthorizer` replaces this method.

pkarman:
@nanotone good question.

Developing this feature for EE has revealed that we are probably not using the best BGS endpoints in Caseflow. The `claimants` BGS service is more reliable.

The `UserAuthorizer` logic itself is less useful in Caseflow because we do not authorize *users* quite the same way. We associate POA itself with a case, rather than an efolder. So we are not just trying to reproduce VBMS behaviour in Caseflow the way we are in EE.

But definitely +1 to exploring a switchover to these BGS endpoints in Caseflow.

pkarman:
@alisan16 I have added testing details to the original ticket. We have a known user to test with.

Issue #1095, opened 2019-05-14
Remove API v1


pkarman:
eFolder Express was redesigned to use React frontend and released as v2 in 2018. None of the now-unused v1 code was ever removed.

This ticket represents the tech debt of removing all the unused v1 code.

See also #1043 

## AC

- [x] Remove v1 code
- [x] Remove feature toggles that controlled v1 vs v2 code
- [x] Drop old tables

alisan16:
@pkarman Can you add more details to this ticket so we can prioritize it if necessary?

PR #1224, opened 2020-05-12
Remove v1 code

pkarman:
Replaces #1182 and connects #1095  

Still TODO:

- [ ] drop old tables
- [ ] remove feature toggles


alisan16:
Is this related to #1095?

aroltsch:
I left a comment in the slack thread last week that I've noticed there were a few files that could be deleted. Posting it here for visibility:

There are a few things left that we can delete:
1. v1 serializers
2. stats_controller.rb
3. controllers/api/v1/application_controller.rb

Also we should probably point jenkins to your branch and test efolder in uat

pkarman:
> Is this related to #1095?

yes @alisan16

pkarman:
@aroltsch and I tested this in preprod and found no regressions.

PR #1227, opened 2020-05-13
Apply per-user authorization at the Veteran model level

pkarman:
Previously we were applying user authorization at the controller level for initial request, but the serialization level makes a subsequent call to BGS which was not getting the same authorization rules applied.

### DIFF
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -123,10 +123,6 @@ class Manifest < ApplicationRecord
     end
   end
 
-  def downloaded_by?(user)
-    users.include?(user)
-  end
-
   def veteran
     @veteran ||= fetch_veteran
   end
--- a/app/models/manifest.rb
+++ b/app/models/manifest.rb
@@ -24,6 +24,9 @@ class Manifest < ApplicationRecord
 
   SECONDS_TO_AUTO_UNLOCK = 5
 
+  # single user for authorization on an instance - not persisted
+  attr_accessor :user
+
   def start!
     # Reset stale manifests.
     update!(fetched_files_status: :initialized) if stale?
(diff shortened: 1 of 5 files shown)


app/services/user_authorizer.rb
```diff
@@ -70,15 +70,15 @@ def veteran_record_found?
     bgs.record_found?(system_veteran_record)
   end
 
-  private
-
-  attr_accessor :sensitive_file, :poa_denied
-
   def system_veteran_record
```
pkarman, line 73: make this method public because now we need it in Veteran model

spec/requests/api/v2/manifests_spec.rb
```diff
@@ -296,4 +296,153 @@ def documents
       end
     end
   end
+
```
pkarman, line 299: these tests all ported from the `api/v2/application_controller_spec`

app/models/manifest.rb
```diff
@@ -24,6 +24,9 @@ class Manifest < ApplicationRecord
 
   SECONDS_TO_AUTO_UNLOCK = 5
 
+  # single user for authorization on an instance - not persisted
+  attr_accessor :user
```
tomas-nava, line 28: It's a little confusing to see this `user` mixed in the class with `user`s passed to methods. Could we rename it `auth_user` or something like that?

app/models/veteran.rb
```diff
@@ -6,6 +6,7 @@ class Veteran
 
   attr_accessor(*BGS_ATTRIBUTES)
   attr_accessor :file_number
+  attr_accessor :user # for authorization
```
tomas-nava, line 9: See last comment...not confusing here, but it should be named consistently.

PR #1229, opened 2020-05-14
Use the POA org pid to compare, not the POA user pid

pkarman:
We were missing one key piece in the user authorization logic model, which is that the PID of the user is not the same PID referenced by the bgs.claimants POA responses. As the field name suggests the BGS response has `person_org_ptcpnt_id` which means we must look up the `org` participant id and use that to compare.

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -84,6 +84,16 @@ class ExternalApi::BGSService
     get_claimant_poa_from_bgs_claimants_poa(bgs_poa)
   end
 
+  # Fetch the POA User's org record
+  def fetch_poa_user_record(poa_participant_id)
+    bgs_poa = MetricsService.record("BGS: fetch record for POA participant id: #{poa_participant_id}",
+                                    service: :bgs,
+                                    name: "org.find_poa_by_participant_id") do
+      client.org.find_poas_by_ptcpnt_id(poa_participant_id)
+    end
+    get_poa_from_bgs_poa(bgs_poa)
+  end
+
   # The participant IDs here are for Claimants.
   # I.e. returns the list of POAs that represent the Claimants.
   def fetch_poas_by_participant_ids(participant_ids)
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -85,13 +85,13 @@ class ExternalApi::BGSService
   end
 
   # Fetch the POA User's org record
-  def fetch_poa_user_record(poa_participant_id)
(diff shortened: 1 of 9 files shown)


app/services/user_authorizer.rb
```diff
@@ -3,6 +3,13 @@
 # determine authorization rules for whether a User can view
 # an efolder for file number.
 #
+# The POA lookup dance with BGS API follows this pattern:
+#
+# 1. Find the participant id of the current user (user.participant_id)
+# 2. Find the POA org participant id with the user's participant id (org_poa_participant_id).
```
pkarman, line 9: previously we were missing step (2) and trying to compare the pid in (1) to (3).

app/services/user_authorizer.rb
```diff
@@ -113,7 +127,7 @@ def fetch_veteran_record_as_system_user
     raise err
   end
 
-  def poa_participant_id
+  def file_number_poa_participant_id
```
ferristseng, line 130: I'm having a bit of trouble understanding this method name. I think with the comments I can understand it, but maybe `participant_id_for_poa_by_file_number` reads clearer?

app/services/external_api/bgs_service.rb
```diff
@@ -84,6 +84,16 @@ def fetch_poa_by_participant_id(participant_id)
     get_claimant_poa_from_bgs_claimants_poa(bgs_poa)
   end
 
+  # Fetch the POA User's org record
+  def fetch_poa_org_record(participant_id)
+    bgs_poa = MetricsService.record("BGS: fetch record for POA participant id: #{participant_id}",
+                                    service: :bgs,
+                                    name: "org.find_poas_by_ptcpnt_id") do
+      client.org.find_poas_by_ptcpnt_id(participant_id)
```
ferristseng, line 92: Is there one POA org for each user or do we only care about one? I'm wondering because the BGS method name is pluralized.

ferristseng: When I was looking at this for VSOs, the flow was pretty similar, and for that, I discovered you could have a user represent multiple different VSOs (https://github.com/department-of-veterans-affairs/caseflow/wiki/VSO-Permission-Mapping-in-Caseflow#determining-access)

tomas-nava: (I've noticed in other areas that BGS can return an array when there are multiple objects to return, and an object when there's only one (as opposed to an array containing a single object))

pkarman: good catch. I had oversimplified when I ported code from caseflow, based on the specific use case I was testing for. I have refactored to assume array-or-object.

PR #1231, opened 2020-05-14
Retry flaky specs

pkarman:
Ported from Caseflow.


spec/features/react_download_spec.rb
```diff
@@ -42,8 +42,6 @@
     S3Service.files = {}
 
     allow(S3Service).to receive(:stream_content) do |key|
-      return unless key
```
rubaiyat22, line 45: Did you mean to remove this line?

pkarman: yep. it was there from earlier experiment to address a flaky test. that particular spec has been failing repeatedly on #1229

PR #1238, opened 2020-05-27
Revert login check for eFolder, for folks going to other apps

leikkisa:
This is in response to batteam request:
https://dsva.slack.com/archives/CHX8FMP28/p1590603820436400

The login rule was slightly changed in a recent PR, and it seems to be breaking for users that can't download eFolder, so reverting to the previous code.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,9 +7,7 @@ class ApplicationController < BaseController
   before_action :set_raven_user
 
   def serve_single_page_app
-     redirect_to("/unauthorized") unless user_is_authorized?
-
-     render("gui/single_page_app", layout: false)
+    can_access_react_app? ? render("gui/single_page_app", layout: false) : redirect_to("/unauthorized")
   end
 
   def authenticate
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -78,6 +78,10 @@ def dropdown_urls
 
   private
 
+  def can_access_react_app?
+    FeatureToggle.enabled?(:efolder_react_app, user: current_user) || Rails.env.development?
```
alisan16, line 82: Note this will always be true since the feature toggle is set to true https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/ansible/vars/features-config.yml#L188

app/controllers/application_controller.rb
```diff
@@ -7,9 +7,9 @@ class ApplicationController < BaseController
   before_action :set_raven_user
 
   def serve_single_page_app
-     redirect_to("/unauthorized") unless user_is_authorized?
+    redirect_to("/unauthorized") unless can_access_react_app?
```
alisan16, line 10: Looking at the old code, this `/unauthorized` endpoint was never hit since can_access_react_app is always true. What happens when someone tries to login to eFolder but doesn't actually have access? What does the `gui/single_page_app` show in that case?

leikkisa:
Closing this PR - it's no longer relevant.

PR #1241, opened 2020-05-27
Add sign in tests

pkarman:
Adds some test coverage for the SSO behavior, especially when serving Caseflow.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -1,7 +1,7 @@
 class SessionsController < ApplicationController
   skip_before_action :check_out_of_service
   skip_before_action :verify_authenticity_token, only: [:create, :failure, :login, :login_creds]
-  skip_before_action :authenticate, only: [:create, :failure, :login, :login_creds]
+  skip_before_action :authenticate, only: [:create, :failure, :login, :login_creds, :me]
 
   class MissingSAMLRequest < StandardError; end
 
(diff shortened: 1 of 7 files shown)


app/controllers/sessions_controller.rb
```diff
@@ -1,7 +1,7 @@
 class SessionsController < ApplicationController
   skip_before_action :check_out_of_service
   skip_before_action :verify_authenticity_token, only: [:create, :failure, :login, :login_creds]
-  skip_before_action :authenticate, only: [:create, :failure, :login, :login_creds]
+  skip_before_action :authenticate, only: [:create, :failure, :login, :login_creds, :me]
```
pkarman, line 4: safelisting `:me` lets us debug the `return_to` value

app/controllers/sessions_controller.rb
```diff
@@ -63,6 +63,10 @@ def me
 
   protected
 
+  def will_redirect_to_url
+    session.delete("return_to") || url_for_sso_host("/")
```
leikkisa, line 67: I wonder if not deleting return_to here would maintain it instead of setting it to nil when it's back in Caseflow's hands after authentication. I know it's pretty late for you so maybe it makes sense to check that out tomorrow?

pkarman: we want to delete it *somewhere* because otherwise the session is "dirty". But where we delete it depends on what we consider the "terminal" point of the authentication dance.

I think that would require some more explicit cooperation between caseflow and EE and that will require some more extensive UX research.

PR #1242, opened 2020-05-28
Do not over-write session return_to

pkarman:
In some instances users are coming to the efolder /login page with a `session["return_to"]` value already set (as when redirected from Caseflow) so do not over-write it.


pkarman:
Verified in manual testing in UAT that this does not fix Caseflow sign in because Caseflow redirects to the /login endpoint directly, which bypasses this `authenticate` method.

Issue #1226, opened 2020-05-13
Cache BGS responses like Caseflow does


pkarman:
example code from Caseflow `BGSService`

```ruby
  def can_access_cache_key(user, vbms_id)
    "bgs_can_access_#{user.css_id}_#{user.station_id}_#{vbms_id}"
  end

  def fetch_veteran_info_cache_key(vbms_id)
    "bgs_veteran_info_#{vbms_id}"
  end
```

PR #1247, opened 2020-06-01
BGS veteran info caching

pkarman:
connects #1226 

## Description

Caching BGS Veteran info responses per-user to optimize performance.

### DIFF
--- a/app/services/external_api/bgs_service.rb
+++ b/app/services/external_api/bgs_service.rb
@@ -54,13 +54,15 @@ class ExternalApi::BGSService
   end
 
   def fetch_veteran_info(file_number, parsed: true)
-    veteran_data =
+    veteran_data = Rails.cache.fetch(fetch_veteran_info_cache_key(file_number), expires_in: 10.minutes) do
       MetricsService.record("BGS: fetch veteran info for vbms id: #{file_number}",
                             service: :bgs,
                             name: "veteran.find_by_file_number") do
         client.veteran.find_by_file_number(file_number)
       end
+    end
     return veteran_data unless parsed
+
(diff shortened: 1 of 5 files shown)


app/services/external_api/bgs_service.rb
```diff
@@ -239,4 +239,18 @@ def fetch_user_info(username, station_id = nil, application = "CASEFLOW")
     fail BGS::NoCaseflowAccess if error.message =~ /TODO unknown error string/
     {}
   end
+
+  def bust_fetch_veteran_info_cache(file_number)
```
tomas-nava, line 243: Is this only used by tests?

app/services/external_api/bgs_service.rb
```diff
@@ -54,13 +54,15 @@ def parse_veteran_info(veteran_data)
   end
 
   def fetch_veteran_info(file_number, parsed: true)
-    veteran_data =
+    veteran_data = Rails.cache.fetch(fetch_veteran_info_cache_key(file_number), expires_in: 10.minutes) do
```
tomas-nava, line 57: Can you say more about the reasoning behind the 10 minute expiration?

app/services/external_api/bgs_service.rb
pkarman: yes

app/services/external_api/bgs_service.rb
pkarman: Most jobs to download all finish w/in 10 minutes so I wanted to make it long enough to make a difference and not so long that any upstream changes were not reflected "immediately".

PR #1251, opened 2020-06-04
Update Sentry error handling

### LINKED ISSUE #1250: Fix Raven exception handling in background jobs
The version of the `sentry-raven` gem we use is bypassing our custom Raven middleware. We need to rewrite all the `capture_exception`-based logic in our `app/jobs` to correctly handle and report errors to Sentry.

Example is https://sentry.ds.va.gov/department-of-veterans-affairs/efolder/issues/7712/events/a65ca1d51ec44121b6b0706da609b9d6/ where there is no `extra` metadata and the error should be ignored from sentry.

pkarman:
resolves #1250 

## Description

The existing Raven middleware class was being bypassed entirely and instead Raven has been sending exceptions directly to Sentry via its built-in ActiveJob integration.

This PR:
* removes unused code
* updates sentry-raven gem to latest version
* configures sentry to ignore certain exception classes (identified as transient in connect_vbms gem et al)
* tests exception handling


app/jobs/application_job.rb
```diff
@@ -1,6 +1,29 @@
 class ApplicationJob < ActiveJob::Base
+  attr_accessor :start_time
+
   before_perform do |job|
     # setup debug context
     Raven.tags_context(job: job.class.name, queue: job.queue_name)
+
+    job.start_time = Time.zone.now
```
pkarman, line 8: unused presently but ported from Caseflow just for parity for future analytics.

pkarman:
@ferristseng caseflow is "working" in a way that EE is not, so upgrading the gem might be useful but won't "fix" anything afaik. I'm not sure if caseflow would benefit from the same configuration pattern as here, as we have an explicit `capture_exception` logic in caseflow that we do not have here in EE.

So yeah, probably, sort of?

ferristseng:
That's fair. I saw that the sentry upgrade was a major version upgrade, so I might be thinking the changes are bigger than they are.

PR #1253, opened 2020-06-05
Log Puma thread count

pkarman:
connects https://github.com/department-of-veterans-affairs/appeals-deployment/issues/2777

Port identical code from Caseflow changing only the `app_name` reported to Data Dog.


app/controllers/concerns/collect_data_dog_metrics.rb
```diff
@@ -28,7 +28,7 @@ def emit_datadog_point(db_name, type, count)
       metric_group: "database",
       metric_name: "#{type}_connections",
       metric_value: count,
-      app_name: "caseflow",
+      app_name: "eFolder",
```
aroltsch, line 31: one tine request to make it all lower case

PR #1268, opened 2020-12-01
Change temp id to primary key on records

hschallhorn:
## Description
Updates the primary key on the records table to be our new temp_id

## Test:

`make migrate` succeeds
`bin/rails db:rollback` succeeds
`bin/rails db:rollback` succeeds
`bin/rails db:rollback` succeeds

after migrating:
```ruby
 pp Record.columns.first
#<ActiveRecord::ConnectionAdapters::PostgreSQLColumn:0x00007fea27414ba0
...
 @default_function=nil,
 @name="id",
 @null=true,
...

 pp Record.columns.last
#<ActiveRecord::ConnectionAdapters::PostgreSQLColumn:0x00007fea2dc02b18
...
 @default_function="nextval('records_temp_id_seq'::regclass)",
 @name="temp_id",
 @null=false,
...

# Ensure id is not being set
manifest = Manifest.create(file_number: "134")
source = ManifestSource.create(name: %w[VBMS VVA].sample, manifest: manifest)
# Don't be concerned if ID still shows in the active record as not null, it is null in the db.
# id here is the primary key of record (now temp id) NOT id column in the db (confusing, sorry)
Record.create(version_id: "1234",  series_id: "5678", manifest_source: source)
Record.pluck(:id).last
=> nil

# ensure we can store a record over max int
 Record.connection.exec_query("SELECT SETVAL('records_temp_id_seq', 2147483650);")
manifest = Manifest.create(file_number: "1340")
source = ManifestSource.create(name: %w[VBMS VVA].sample, manifest: manifest)
Record.create(version_id: "1234",  series_id: "5678", manifest_source: source)
=> #<Record ..., temp_id: 2147483651>
```

### DIFF
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -25,8 +25,6 @@ class Record < ApplicationRecord
   # then PDFkit will check if PDF is valid
   before_create :adjust_mime_type
 
-  after_save :save_to_temp_columns
-
   delegate :manifest, :service, to: :manifest_source
   delegate :file_number, to: :manifest
 
@@ -111,9 +109,5 @@ class Record < ApplicationRecord
 
   def adjust_mime_type
     self.mime_type = "application/pdf" if mime_type == "application/octet-stream"
-  end
-
-  def save_to_temp_columns
-    self.temp_id = id
   end
 end
--- a/app/models/record.rb
+++ b/app/models/record.rb
@@ -32,6 +32,8 @@ class Record < ApplicationRecord
 
   MAXIMUM_FILENAME_LENGTH = 100
 
+  self.primary_key = "temp_id"
+
   def fetch!
     fetcher.process
   end
--- a/db/migrate/20201201195149_change_temp_id_to_not_null.rb
+++ b/db/migrate/20201201195149_change_temp_id_to_not_null.rb
@@ -1,0 +1,8 @@
+class ChangeTempIdToNotNull < ActiveRecord::Migration[5.2]
+  def up
+    change_column :records, :temp_id, :bigint, null: false
+  end
+  def down
+    change_column :records, :temp_id, :bigint, null: true
+  end
+end
--- a/db/migrate/20201201205105_change_record_id_to_int.rb
+++ b/db/migrate/20201201205105_change_record_id_to_int.rb
@@ -1,0 +1,14 @@
+class ChangeRecordIdToInt < ActiveRecord::Migration[5.2]
+  def up
+    change_column :records, :id, :int, null: true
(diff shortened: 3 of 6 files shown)


app/models/record.rb
```diff
@@ -25,15 +25,15 @@ class Record < ApplicationRecord
   # then PDFkit will check if PDF is valid
   before_create :adjust_mime_type
 
-  after_save :save_to_temp_columns
-
```
hschallhorn, line 29: Remove our callbacks as temp_id should be auto incrementing here: https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1268/files#diff-401c73b7a8075d77e9d9c708d76aadfe648b4d7168335f09440dbe9e0c2dc937R10

app/models/record.rb
```diff
@@ -25,15 +25,15 @@ class Record < ApplicationRecord
...
   delegate :manifest, :service, to: :manifest_source
   delegate :file_number, to: :manifest
 
   AVERAGE_DOWNLOAD_TIME_IN_SECONDS = 2
 
   MAXIMUM_FILENAME_LENGTH = 100
 
+  self.primary_key = "temp_id"
+
```
hschallhorn, line 36: Most likely unneeded as no other associations rely on record_id, but ok to have

db/migrate/20201201195149_change_temp_id_to_not_null.rb
```diff
@@ -0,0 +1,8 @@
+class ChangeTempIdToNotNull < ActiveRecord::Migration[5.2]
+  def up
+    change_column :records, :temp_id, :bigint, null: false
```
hschallhorn, line 3: Postgres requires a not null constraint for primary keys

db/migrate/20201201205105_change_record_id_to_int.rb
```diff
@@ -0,0 +1,14 @@
+class ChangeRecordIdToInt < ActiveRecord::Migration[5.2]
+  def up
+    change_column :records, :id, :int, null: true
```
hschallhorn, line 3: change from serial to int

db/migrate/20201201205105_change_record_id_to_int.rb
```diff
@@ -0,0 +1,14 @@
...
+    execute "ALTER TABLE records ALTER COLUMN id DROP DEFAULT;"
+    execute "DROP SEQUENCE records_id_seq;"
```
hschallhorn, line 5: Remove the auto incrementing function from id

hschallhorn:
Closing to break apart

PR #1272, opened 2020-12-01
Bigint for record ids

meckilabritt-navapbc:


alisan16:
Closing this as we went with another option for resolving the production incident. Full post-mortem here: https://github.com/department-of-veterans-affairs/appeals-deployment/blob/master/docs/post-mortems/2020-12-02-efolder-max-int-records-id-col.md

PR #1283, opened 2021-01-25
Fix cache key for veteran info to prevent data leakages

ferristseng:
Resolves department-of-veterans-affairs/caseflow#15829

### Description

Updates the veteran info cache key to prevent data leakages

### Testing Plan

Copy & paste code in prod and test


nanotone (approved):
Looks good to me!

PR #1334, opened 2022-12-16
Mattray gov/appeals 10276

mikefinneran:
Resolves APPEALS-10276

### Description
Updated manifest fetcher to fetch the documents that have changed since the last time the manifest was fetched. The manifest fetcher's documents method will return the result of either the fetch_documents method or the fetch_delta_documents method based on whether or not the manifest's source is current. If the source is not current it will use the original fetch_documents to get all the documents and delete and recreate all the Record models for the documents. If the manifest source is current it will use a new fetch_delta_documents method in the VBMS service. This method takes in the manifest sources's fetched_at field and builds a VBMS request for document version references within a certain date range. It then sends and logs the request and returns the documents that have changed. The document creator will then delete the old versions of the updated documents and replace them with updated ones.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
Make sure the docker containers are running ```make up```
Run all the tests ```make test```
Run just the Manifest Fetcher tests ```bundle exec rspec spec/services/manifest_fetcher_spec.rb ```
Run just the Document Creator tests ```bundle exec rspec spec/services/document_creator_spec.rb```

### Code Documentation Updates
- [ ] Add or update code comments at the top of the class, module, and/or component.

### Integrations: Adding endpoints for external APIs
* [ ] Check that Caseflow's external API code for the endpoint matches the code in the relevant integration repo
  * [ ] Request: Service name, method name, input field names
  * [ ] Response: Check expected data structure
* [ ] Update Fakes
* [ ] Integrations impacting functionality are tested in Caseflow UAT


docker-compose.yml
```diff
@@ -1,7 +1,7 @@
 version: '3'
 services:
   appeals-redis:
-    image: redis:2.8.23
+    image: redis:3.2.10
```
mikefinneran, line 4: Is this a M1 or local system specific change and do we want this in prod?

.ruby-version
```diff
@@ -1 +1 @@
-2.5.3
+2.7.3
```
mikefinneran, line 1: Is this a M1 or local system specific change and do we want this in prod?

jshields13: @mikefinneran Yeah I would be hesitant to just randomly update the ruby version. That should have its own story with proper testing

app/jobs/v2/download_manifest_job.rb
```diff
@@ -8,14 +8,36 @@ def perform(manifest_source, user = nil)
...
+  end
+
+  def log_message(file_number, docs, zipfile_size)
+    "DownloadManifestJob - " \
+    "User Inspect: (#{current_user.inspect})" \
+    "File Number: (#{file_number}) - " \
+    "Documents Fetched Count: #{docs.count} - " \
+    "Documents Fetched IDs: #{docs.map(&:id)}" \
+    "Manifest Zipfile Size: #{zipfile_size} - "
+  end
```
jshields13, line 42: Not a filing comment but we should be adding tomdoc style comments to new methods

app/services/document_creator.rb
```diff
@@ -29,4 +39,9 @@ def create
   def external_documents
     DocumentFilter.new(documents: @external_documents).filter
   end
+  private
+
+  def current_user
+    RequestStore[:current_user]
+  end
```
jshields13, line 46: Could factor out this into a concern so you don't have to define it in every file

app/services/external_api/vbms_service.rb
```diff
@@ -44,6 +44,14 @@ def self.v2_fetch_documents_for(veteran_file_number)
...
   end
 
+  def self.fetch_delta_documents_for(veteran_file_number, begin_date_range, end_date_range = Time.zone.now)
+    @vbms_client || init_client
+    
+    request = VBMS::Requests::FindDocumentVersionReferenceByDateRange.new(veteran_file_number, begin_date_range, end_date_range)
+    documents = send_and_log_request(veteran_file_number, request)
+    documents
+  end
+
```
jshields13, line 54: This method shoudl really have proper comments following the tom doc style guide

app/services/manifest_fetcher.rb
```diff
@@ -61,4 +61,50 @@ def bgs_client
...
+    docs
+  end
+  
+
+  def documents_from_service_for(file_number)
+    documents = MetricsService.record("ManifestFetcher documents or delta documents for file_number: #{file_number}",
+                                               service: manifest_source.name.downcase.to_sym,
+                                               name: "fetch_documents_or_delta_from_service") do
+      fetch_documents_or_delta_documents_for(file_number)
+  end
```
jshields13, line 90: Method comments be useful

app/services/manifest_fetcher.rb
```diff
@@ -61,4 +61,50 @@ def bgs_client
...
+
+
+  def current_user
+    RequestStore[:current_user]
+  end
+
+  def log_info(file_number, docs, ft_delta_documents, zipfile_size)
+    Rails.logger.info log_message(file_number, docs, ft_delta_documents, zipfile_size)
+  end
```
jshields13, line 99: Looks like this is also repeated across some files. I would consider moving into a concern. May of been worth making a reporting concern or utility. This will make maintaining code a nightmare if we want to change or add to our logging

jshields13 (changes requested):
Test plans need to be jira and linked to the PR description

mattray-gov (approved):
LGTM

zurbergram (approved):

maura-shaughnessy-bah (dismissed):

mgrozinski (approved):
lgtm

PR #1370, opened 2023-03-17
CI Migration

yeedster:
-Added ecr login python script
-Added secrets to login to AWS
-Migrated over existing circleci workflows


AKeyframe (commented):
I only have two comments for these. The first being that on line 64 of the build.yml, the specific version of bundler we are currently using is being locked in. While in the circleci file it's getting the version information from the Gemfile.lock. Getting this to parity would be nice, so that in the future we won't have to update (or remember to update) this file every time we update bundler.

Beyond that, the only other thing is that general readability needs to be improved. Each step should be broken up with empty lines and so should the each section / job. The commands for each step can stay together though.

yeedster:
@AKeyframe 
Addressed your needs for having the bundle version pull from the gemlock file and added more spacing to be readable.

jshields13 (approved):

PR #1543, opened 2024-03-29
AlecK/APPEALS-26886 eFolder Post Rails 6.0 Config Update

AKeyframe:
Resolves [APPEALS-26886](https://jira.devops.va.gov/browse/APPEALS-26886)

# Required
This story needs to be tested and merged at the same time as [APPEALS-34124](https://jira.devops.va.gov/browse/APPEALS-34124), since caseflow and eFolder share the same session cookie.

When we release to Production, end users should not notice any changes, however there is a low risk that they could experience some authentication wonkiness, forcing them to log in again.

### Description
As part of update to Rails `6.0.Z`, we set `config.action_dispatch.use_cookies_with_metadata = false` to enable seamless rollback to Rails `5.2.Z` if necessary. Now that Rails 6.0 is stable in production, we can safely migrate to the default for this setting.

Rails 5.2 introduced setting `config.action_dispatch.use_cookies_with_metadata`.  We are currently overriding the default in order to preserve the pre-existing behavior, however, since we are now changing cookie configuration for Rails 6.0, this is an opportune time to migrate to the default for this setting as well.

### Acceptance Criteria
After Rails 6.0 is stable in production, in the `config/application.rb` file:
- remove the override for `config.action_dispatch.use_authenticated_cookie_encryption`
- remove the override for `config.action_dispatch.use_cookies_with_metadata`

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -92,15 +92,6 @@ module CaseflowEfolder
     # This can be changed to the defualt and removed if we no longer support IE5-8 (old browsers)
     Rails.application.config.action_view.default_enforce_utf8 = true
 
-    # Embed purpose and expiry metadata inside signed and encrypted
-    # cookies for increased security.
-    #
-    # This option is not backwards compatible with earlier Rails versions.
-    # It's best enabled when your entire app is migrated and stable on 6.0.
-    # Default change to true as of 6.0
-    # Remove after stable 6.0
-    Rails.application.config.action_dispatch.use_cookies_with_metadata = false
-
     # Enable the same cache key to be reused when the object being cached of type
     # `ActiveRecord::Relation` changes by moving the volatile information (max updated at and count)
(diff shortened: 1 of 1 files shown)


jcroteau (approved):
⭐ LGTM!

**Note to other reviewers:** The following failing specs are failing on other PR builds as well; they are unrelated to changes in this PR:
- `spec/features/backend_error_flows_spec.rb:69`
- `spec/features/backend_error_flows_spec.rb:91`
- `spec/features/react_download_spec.rb:65`
- `spec/requests/api/v2/records_spec.rb:128`

jcroteau:
It turns out that this will break authentication after all.

While testing these changes in PreProd, we found that, while initial authentication was successful, attempting to login again after logging out would fail.

So we will omit these changes from the Rails 6.1 upgrade.

For more details, see https://jira.devops.va.gov/browse/APPEALS-54897

PR #1664, opened 2024-08-13
feature/APPEALS-26886-46558 - Rails 6.1 Upgrade

jcroteau:
Combines changes from
- https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1543
- https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1636


jcroteau:
It turns out that https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1543 will break authentication after all.

While testing these changes in PreProd, we found that, while initial authentication was successful, attempting to login again after logging out would fail.

So we will omit these changes from the Rails 6.1 upgrade.

For more details, see https://jira.devops.va.gov/browse/APPEALS-54897

PR #1675, opened 2024-09-03
Fix Missing Manifest User Causing Sensitivity Check Failures

tradesmanhelix:
Resolves [eFolder>>View Result link on Recent Downloads throwing error page](https://jira.devops.va.gov/browse/APPEALS-57091).

### Description
For the `refresh` action in `app/controllers/api/v2/manifests_controller.rb`, the app was using `Manifest.find` to load the manifest. However, this would result in the `manifest` not having a `user` set and this was causing the new user sensitivity checks to fail.

This PR updates the `refresh` action in `app/controllers/api/v2/manifests_controller.rb` to use the `Manifest.find_or_create_by_user` method which will set `manifest.user` as part of its logic.

### Acceptance Criteria
- [ ] All specs passing

### Testing Plan
NOTE: UAT only

Follow repro steps in ticket and verify the issue is fixed.

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -13,6 +13,15 @@ class Api::V1::ApplicationController < BaseController
     }, status: 500
   end
 
+  rescue_from BGS::SensitivityLevelCheckFailure do |e|
+    render json: {
+      status: e.message,
+      featureToggles: {
+        checkUserSensitivity: FeatureToggle.enabled?(:check_user_sensitivity)
+      }
+    }, status: :forbidden
+  end
+
   rescue_from BGS::PublicError do |error|
     forbidden(error.public_message)
   end
--- a/app/controllers/api/v2/manifests_controller.rb
+++ b/app/controllers/api/v2/manifests_controller.rb
@@ -2,18 +2,18 @@
 
 class Api::V2::ManifestsController < Api::V2::ApplicationController
   # Need this as a before action since it gates access to these controller methods
-  before_action :file_number, only: [:start, :refresh]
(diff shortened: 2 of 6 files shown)


app/controllers/api/v2/manifests_controller.rb
```diff
@@ -1,24 +1,25 @@
...
     manifest.start!
     render json: json_manifests(manifest)
-  rescue BGS::SensitivityLevelCheckFailure
-    forbidden("This user does not have permission to access this information")
   end
 
   def refresh
-    manifest = Manifest.find(params[:id])
-    return record_not_found unless manifest
+    manifest = Manifest.includes(:sources, :records).find_or_create_by_user(user: current_user, file_number: veteran_file_number)
```
tradesmanhelix, line 16: Root cause of bug. `find` does not set `manifest.user` since it's a transient property and not stored in the database.

Using `find_or_create_by_user` sets `manifest.user` and should allow sensitivity checks to succeed.

app/controllers/api/v1/application_controller.rb
```diff
@@ -13,6 +13,15 @@ class Api::V1::ApplicationController < BaseController
...
   end
 
+  rescue_from BGS::SensitivityLevelCheckFailure do |e|
+    render json: {
+      status: e.message,
+      featureToggles: {
+        checkUserSensitivity: FeatureToggle.enabled?(:check_user_sensitivity)
+      }
+    }, status: :forbidden
+  end
```
tradesmanhelix, line 23: Refactored here to alleviate the need for individual controller methods to handle this.

cacevesva (approved):

PR #1700, opened 2024-09-24
Zeitwerk Implementation AlecK/APPEALS-44512

AKeyframe:
Resolves #{efolder issue number} / department-of-veterans-affairs/caseflow#{caseflow issue number} / CASEFLOW-{JIRA number}

### Description
Please explain the changes you made here.

### Acceptance Criteria
- [ ] Code compiles correctly

### Testing Plan
1. Go to ...

- [ ] For higher-risk changes: [Deploy the custom branch to UAT to test](https://github.com/department-of-veterans-affairs/appeals-deployment/wiki/Applications---Deploy-Custom-Branch-to-UAT)

### DIFF
--- a/lib/scripts/current_user_load_test.rb
+++ b/lib/scripts/current_user_load_test.rb
@@ -1,4 +1,4 @@
-class CurrentUserLoadTests
+class CurrentUserLoadTest
   def initialize(iterations = 100)
     @iterations = iterations
   end
--- a/Gemfile
+++ b/Gemfile
@@ -10,7 +10,7 @@ gem "aws-sdk-s3"
 gem "aws-sdk-sqs"
 gem "bgs", git: "https://github.com/department-of-veterans-affairs/ruby-bgs.git", ref: "a2e055b5a52bd1e2bb8c2b3b8d5820b1a404cd3d"
 gem "bootsnap", require: false
-gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", ref: "9bd3635fbd8094d25160669f38d8699e2f1d7a98"
+gem "caseflow", git: "https://github.com/department-of-veterans-affairs/caseflow-commons", branch: "AlecK/Zeitwerk" # ref: "9bd3635fbd8094d25160669f38d8699e2f1d7a98"
 gem "coffee-rails", "> 4.1.0"
 gem "connect_vbms", git: "https://github.com/department-of-veterans-affairs/connect_vbms.git", branch: "master"
 gem "connect_vva", git: "https://github.com/department-of-veterans-affairs/connect_vva.git", ref: "dfd1aeb2605c1f237f520bcdc41b059202e8944d"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -20,8 +20,8 @@ GIT
 
 GIT
   remote: https://github.com/department-of-veterans-affairs/caseflow-commons
-  revision: 9bd3635fbd8094d25160669f38d8699e2f1d7a98
-  ref: 9bd3635fbd8094d25160669f38d8699e2f1d7a98
+  revision: 298c0a10c67920ad98251c0a9f3bd37abf408c35
+  branch: AlecK/Zeitwerk
   specs:
     caseflow (0.4.8)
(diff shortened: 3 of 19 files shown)


lib/scripts/current_user_load_test.rb
```diff
@@ -1,4 +1,4 @@
-class CurrentUserLoadTests
+class CurrentUserLoadTest
```
jcroteau, line 1: Renaming this constant is probably fine -- I don't see it being used anywhere anyways.

However, for cases like these, it's probably less risky to leave the constant name alone and just rename the file to match (and updating references to the filename, if any).

I'll leave it to your discretion which approach to take here.

spec/config/zeitwerk_spec.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+require "rails_helper"
```
jcroteau, line 2: I think this `require` is not necessary. This test runs fine without it.

spec/config/zeitwerk_spec.rb
```diff
@@ -0,0 +1,9 @@
...
+
+# See config/initializers/zeitwerk.rb if this fails
```
jcroteau, line 4: This comment is going to be misleading after removing the `zeitwerk` initializer. I think it's better to just let any errors speak for themselves should this test fail.

spec/config/zeitwerk_spec.rb
```diff
@@ -0,0 +1,9 @@
...
+RSpec.describe "Zeitwerk Compliance Check" do
+  it "Eager loads all files without errors" do
+    expect { Rails.application.eager_load! }.not_to raise_error
+  end
+end
```
jcroteau, line 9: The linter might complain about a missing newline here.

jcroteau (changes requested):
Looking good so far! Just a few minor changes...

jcroteau:
Also, we have a couple `file * is ignored` warnings in the Zeitwerk logs left to address:

```
Zeitwerk@rails.main: file /Users/jeremycroteau/dev/appeals/caseflow-efolder/lib/fakes/test_auth_strategy.rb is ignored because Fakes::TestAuthStrategy is already defined
```
```
Zeitwerk@rails.main: file /Users/jeremycroteau/dev/appeals/caseflow-efolder/lib/tasks/support/shell_command.rb is ignored because ShellCommand is already defined
```

jcroteau (changes requested):
In addition to the mostly cosmetic changes below, we'll need to modify a couple more initializers (I had forgotten to check these in my prior reviews):

- `config/initializers/deprecation_warning_subscriber.rb`
  - Let's wrap this in an `after_initialize` block, since it references reloadable constant `SlackService`.
  - A `to_prepare` block would **not** be appropriate here, since each application reload would then add another subscriber of the same kind, leading to multiple redundant log / Sentry entries for a single deprecation warning.

- `config/initializers/errors.rb`
  - Let's rename the file to `standard_error.rb`
  - Wrap this in a `before_initialize` block, so that this monkey-patch is applied before any other code is loaded.

jcroteau (dismissed):

jcroteau (approved):

.simplecov
```diff
@@ -1,17 +1,18 @@
 if ENV["RAILS_ENV"] == "test"
   SimpleCov.start do
+    add_filter "app/controllers/errors_controller.rb"
```
amybids, line 3: Is arranging these to be in alphabetical order necessary as a part of this work?

amybids (dismissed):

.simplecov
jcroteau: This was my doing, not Alec's. Is it necessary? No. But it was a good opportunity to leave something better than I found it, and it's a low-risk change. It was also done as a separate ["refactor" commit](https://github.com/department-of-veterans-affairs/caseflow-efolder/pull/1700/commits/b27bdd7f2291731daaac7f69467e2f974c02a24f) before adding the new filter(s). 

In general, I think it's a good practice to order things lexically to make them easier to parse (especially if it's a very long list). It also helps to avoid adding duplicate entries for things.

